PR #3, opened 2019-08-20
First shot at intention and outline for simplest smart contract tutorial

### LINKED ISSUE #1: Tutorial on how to create a simple one-player game and how to run it in Cosmic-Swingset
A couple of questions i cannot answer yet:
- is `{try(attempt){return attempt === 37 ? 'you win' : 'you lose'}` a "valid" contract?
    - currently? eventually?
- is Swingset a requirement to run contracts?
- Do contracts need to be written in Jessie?
    - currently? eventually?

DavidBruant:
Fixes #1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,62 @@
 # Documentation
-User documentation
+
+This repository contains documentation of [Agoric](https://agoric.com/) products
+
+Agoric develop technologies to enable safe collaboration via the creation of [smart contracts](http://www.erights.org/talks/pisa/paper/#smart-contracts) running on blockchains
+
+
+## Current intention
+
+The current work aims at creating a progressive learning experience into writing Smart Contracts as proposed by Agoric. An attention is given to provide a very simple first experience and increase the complexity gradually
+
+Identified sources of complexity:
+- How contracts can be expressed
+    - Jessie
+- What can be expressed in contracts
+    - multi-party contracts with invites
+    - turn-based contracts
+    - ERTP framework
+        - ContractHost
+        - Escrow
+    - CoveredCall
+    - hierarchical rights
+    - non-determinism (time, randomness)?
+- Infrastructure/Tooling
+    - @agoric npm packages
+    - User interface
+    - Vats
+    - Swingset
+    - Cosmic-Swingset
+
+
+### Tutorial on creating and running the simplest Smart Contract
+
+This first contract will be a one-player game where the player tries to guess a number. If they guess right, they win, otherwise, they lose. The game can be played any number of times and there is no actual stake. For the purpose of keeping things simple, the number is hardcoded to be `37`
+
+#### Set up the environment for the contract
+
+- clone some repo + cd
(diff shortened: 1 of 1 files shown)


DavidBruant:
Hey @Chris-Hibbert, i feel i have a reviewable version of the first smart contract tutorial

I made an effort to keep code as small as possible, and have as few steps as possible to get something meaningful running. In following tutorials, things will get gradually more complex

I haven't been able to make the tutorial simpler than this. I would be primarily interested in feedback on how to make things simpler if you can find ways

DavidBruant:
> rather than using values from history, which will change from session to session, I tend to assign to names in home:

oh! i didn't know that was an option!
I've simplified the tutorial so the redeem part is now done server-side, but this advice will be useful somewhere else!
Thanks

README.md
```diff
@@ -1,2 +1,171 @@
...
+
+# Create a directory for the contract
+mkdir contracts
+# Create the contract file
+touch contracts/guess37.js
+```
+
+- copy/paste the following code in the file:
+```js
+function guess37Contract(terms, inviteMaker){
```
Chris-Hibbert: space after the close paren

README.md
```diff
@@ -1,2 +1,171 @@
...
+
+    return harden({
+        playerInvite: inviteMaker.make('player', seat),
+    });
+}
+
+export default guess37Contract.toString()
+```
+
+The contract is materialized by the source code of the `start` function. JavaScript enables to retrieve the source code of a function by calling [toString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString) on the function
```
Chris-Hibbert: 'enables' needs a direct object, like "enables you to ...". Perhaps "JavaScript makes it possible to  retrieve ..." if you want it remain in the impersonal form.

README.md
```diff
@@ -1,2 +1,171 @@
...
+    - Explain `terms` and `inviteMaker` arguments
+- Explain the [provided top-level variables](https://github.com/Agoric/ERTP/blob/8f42af53581821bb16a9f4e3d62603833354ef8b/core/contractHost.js#L53-L63)
+- Explain Jessie
+</small>
+
+Change the content of `lib/ag-solo/vats/vat-demo.js` to be the following:
+
+```js
+import harden from '@agoric/harden';
+import guess37ContractSource from '../../contracts/guess37.js'
```
Chris-Hibbert: most of these statements need a final semicolon. Would you run this through eslint, so we can be sure that they would pass our formatting checks? It's also going to insist on a comma on line 115, and other formatting, I'd guess.

README.md
DavidBruant: Addressed in follow-up commit

README.md
DavidBruant: Addressed in follow-up commit

README.md
DavidBruant: Addressed in follow-up commit

PR #11, opened 2019-08-23
First shot at contributing rules

DavidBruant:
This is a first draft
All feedback is welcome

My intention is to have something short quickly approved and that we can iterate from here over time

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -7,6 +7,7 @@ It is meant to work like a wiki
 Here are a couple of rules to follow:
 - The content must be written in **English**
 - The content must be written in **markdown** (`.md` file extension)
+- Code samples should adhere to Agoric formatting rules. They can be found in the form of a [prettier config](https://github.com/Agoric/new-repo/blob/master/.prettierrc.json) and [eslint config](https://github.com/Agoric/new-repo/blob/master/.eslintrc.js)
 
 If it's your first contribution, it will go through a [pull request](https://help.github.com/en/articles/about-pull-requests#about-pull-requests) process, that is, someone will need to review and accept your changes first
 
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -10,7 +10,7 @@ Here are a couple of rules to follow:
 
 If it's your first contribution, it will go through a [pull request](https://help.github.com/en/articles/about-pull-requests#about-pull-requests) process, that is, someone will need to review and accept your changes first
 
-When you're trusted enough, you will granted the right to modify the content directly
+When you're trusted enough, you will be granted the right to modify the content directly
 
 If you do not feel confident about the changes you want to make, you can always create a pull request and request a review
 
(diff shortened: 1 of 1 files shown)


.github/contributing.md
```diff
@@ -0,0 +1,19 @@
+# Contributing
+
+This repository contains the Agoric's technical documentation
+
+It is meant to work like a wiki
+
+Here are a couple of rules to follow:
+- The content must be written in **English**
+- The content must be written in **markdown** (`.md` file extension)
```
erights: I notice that some of our own documentation is *.rst rather than *.md . Should we suggest that contributions can/should be *.rst ? How do we decide when to use which?

Attn @katelynsills @Chris-Hibbert @tatyana0219 @btulloh

DavidBruant: for reference, i don't have a strong intention on markdown, it's only a first proposition

> I notice that some of our own documentation is *.rst

i hadn't noticed that, thanks for the info

The points i find positive about markdown are:
- Easy to get started when you already know how to write with Microsoft Word (!), because text is valid markdown and paragraph are naturally understood
    - simple learning curve (heading levels, bold, underline, links)
- Fairly popular (github/~Trello/~Slack), so lots of people are probably already familiar + there are lots of markdown-related resources on the web
- Web-friendliness: there is a natural transformation from markdown to HTML (and it takes only [minutes to transform a github repo with markdown files into a website](https://pages.github.com/))
- complex layouts and styling are possible because HTML can be written inside markdown + it's easy to add CSS/JavaScript to a resource in markdown

From what i've understood, one downside of markdown is that it can be harder to work with when trying to write a book, but i don't have this experience myself, so i cannot comment

.github/contributing.md
```diff
@@ -0,0 +1,19 @@
...
+
+If it's your first contribution, it will go through a [pull request](https://help.github.com/en/articles/about-pull-requests#about-pull-requests) process, that is, someone will need to review and accept your changes first
+
+When you're trusted enough, you will granted the right to modify the content directly
```
Chris-Hibbert: "you will *be* granted"

DavidBruant: addressed in follow-up commit

.github/contributing.md
Chris-Hibbert: MarkDown is certainly a reasonable place to start. Let's agree that that's what we're using for most purposes. 

If we need to use something else for some purposes, or eventually change to a different standard, we'll deal with that then.

DavidBruant: I created a follow-up issue on the *.rst files: https://github.com/Agoric/Documentation/issues/19

DavidBruant: I'll wait for @erights to come back from TC39 and if there is no urgent reason not to merge this PR, i'll merge it
And everything can be modified afterwards; a github repo is de facto a wiki

katelynsills:
Closing as outdated

PR #16, opened 2019-09-20
Pixel code architecture 

DavidBruant:
Diagram related to https://github.com/Agoric/ERTP/issues/121


katelynsills:
This is very cool! @DavidBruant how do create your svg such that they have clickable links?

katelynsills:
It seems like "running code" and "stores the code of" are mostly the same, and maybe shouldn't be divided up. Was there something that pushed you in that direction?

DavidBruant:
> how do create your svg such that they have clickable links?

I use Google Drawings (see source: https://docs.google.com/drawings/d/1VCFIqPSJvv_iVbwc4Jng6rN0_4dIBrU-2wkrxqPJlUc/edit )
It's nowhere near being a feature-complete drawing software, but i find it very useful to make diagrams as it helps with:
- making arrows point to sides of boxes
- help with box alignments and same sizes
- links in the diagram

I create links from the Drawings UI and then i exported as SVG. [SVG has links via the `a` element like HTML](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/a). I was happy to discover that Google Drawing exports to SVG maintained the links :-) (it was the first time i tried to do that)

> It seems like "running code" and "stores the code of" are mostly the same

indeed

> and maybe shouldn't be divided up. Was there something that pushed you in that direction?

I'm not sure yet. I'm still exploring (i should have made this a draft pull request, but clicked too quickly)
On the blockchain side, there is no difference. The only notable difference is that the user interface code of the current Pixel Demo Gallery is stored in the solo node and then served to be run on the client-side

In the next drawing (i'll update this PR) the solo node will run less code and the ui code will be stored in the blockchain

DavidBruant:
Right now, i consider everything in this PR is very drafty (wording, visual) but in a state where it's ready for discussion

katelynsills:
Closing as outdated

PR #17, opened 2019-10-01
Initial setup for ERTP documentation

barbaraliau:
- Docs written in Markdown and built with VuePress
- Contains the initial configuration for ERTP docs
- Scripts and `deploy.sh` updated to reflect change in base folder name (typically called `/docs`)
  - `npm run docs:dev` --> `npm run docs-ertp:dev`
  - `npm run docs:build` --> `npm run docs-ertp:build`

Most of the content and formatting of the methods is changing. Much of this is set up just to show navigation and content format.

A live version is currently up at https://barbaraliau.github.io/ERTP/

PR #22, opened 2019-10-20
New first contract

DavidBruant:
This is a different take of the first contract where less of the ERTP choregraphy happens in the code and more happens on the client-side

I kept things minimalist as far as explaining ERTP itself and delegate it to the [ERTP dedicated documentation](#17)


smart-contracts-tutorial/first-contract.md
```diff
@@ -0,0 +1,153 @@
...
+
+```md
+TODO a documentation of the contract format
+- Explain the contract host
+    - Explain `terms` and `inviteMaker` arguments
+- Explain the [provided top-level variables](https://github.com/Agoric/ERTP/blob/8f42af53581821bb16a9f4e3d62603833354ef8b/core/contractHost.js#L53-L63)
+- Explain Jessie
+```
+
+Replace the content of `lib/ag-solo/vats/vat-pixel.js` with the following:
```
Chris-Hibbert: Add a note that we're doing this to save a bunch of configuration work that isn't relevant to this exercise. Promise to document the proper way to install a new contract later.

DavidBruant: https://github.com/Agoric/Documentation/issues/24

DavidBruant:
I added warnings about tutorial steps being hard sometimes
I'm re-requesting a review because the wording i chose sounds a bit like a commitment

If that's find with you as is, you can merge the PR

I'm creating an issue label called [Contract developer experience](https://github.com/Agoric/Documentation/labels/Contract%20Developer%20Experience). It will be used to track anything that is lived as annoying by contract devs. Issues with this label probably will link to follow-up issues in other repos

Chris-Hibbert:
I don't mind committing to continuing to work to make things better, since that's what we plan to do.

I don't like the phrasing "the steps in this tutorial may seem unnecessarily complicated". The work-around I pointed at is actually simpler than the right way to do things, so it's overly simplified. I think this would be better covered by an annotation at the point that we tell them to overwrite a file, saying that this is a hack that saves them work, and that better solutions are being developed. (As you've noted, they may even be in place quite soon, but they're certainly not ready now.)

barbaraliau:
Closed in favor of #23

PR #23, opened 2019-10-21
Guess37 mutiple participants

DavidBruant:
This PR builds on top of #22 

Only `guess37-multiple-participants.md` is relevant to review

I left questions at the end because i'm puzzled about how to prove an object is a genuine invite


smart-contracts-tutorial/guess37-multiple-participants.md
```diff
@@ -0,0 +1,114 @@
...
+```
+home
+handoff = home~.handoff
+board = handoff~.grabBoard("guess37 game")
+playerInvite = board~.lookup("invite")
+```
+
+Now we have the invite to participate to the contract. Let's play the game:
+```
+seat = host~.redeem(playerInvite)
```
DavidBruant: `host` is undefined

smart-contracts-tutorial/first-contract.md
```diff
@@ -109,30 +101,53 @@ make scenario3-run-client
...
 TODO : provide a proper UI
-</small>
+```
 
 In the REPL, try the following commands in the same order:
 ```
 home
-home!game!guess(14)
-home!game!guess(37)
+host = home~.contractHost
```
Chris-Hibbert: Accessing fields on home doesn't need '~.'   just '.' is sufficient

smart-contracts-tutorial/first-contract.md
```diff
@@ -109,30 +101,53 @@ make scenario3-run-client
...
+
+seat~.guess(14)
+seat~.guess(37)
 ```
 
-The first `home` command shows what was defined in `lib/ag-solo/vats/vat-demo.js` as `chainBundle`
+### Commands breakdown 
+
+The first `home` command shows what was defined in `lib/ag-solo/vats/vat-pixel.js` as return value of `createPixelBundle`
+It is a presence for an object containing the properties `contractHost`, `handoffService` and `guess37ContractSource`
```
Chris-Hibbert: Home is an object with fields in the local vat. The objects in the fields are all promises or presences, but home is local.

smart-contracts-tutorial/first-contract.md
```diff
@@ -109,30 +101,53 @@ make scenario3-run-client
...
-`home!game!guess(37)` is a winning guess
+`host = home.contractHost` retrieves the contractHost presence and stores it in the `host` variable for convenience
+
+`installation = host~.install({start: home.guess37ContractSource})` installs the guess37 contract in the contract host
+
+`invites = installation~.spawn()` runs the contract (here, the `guess37Contract` function) and store the invites bundle returned by the contract
+
+`playerInvite = invites~.playerInvite` extracts the `playerInvite` from the invites bundle
+
+`seat = host~.redeem(playerInvite)` redeems the invitation. This allows to participate to the contract. What is returned is stored in a `seat` variable as if the contract/game was a table and you were retrieving a seat at this table
```
Chris-Hibbert: "This allows you to participate in the contract." 

"Redeem returns an object that represents a right to participate in the contract (a seat at the table)."

A seat at the table is a common enough metaphor that it doesn't deserve an explanation here.

smart-contracts-tutorial/guess37-multiple-participants.md
```diff
@@ -0,0 +1,114 @@
+# Guess37 multiple participants
+
+In the [previous step](./first-contract), we set up a contract and run it and played the guess37 game
```
Chris-Hibbert: s/run/ran/

smart-contracts-tutorial/guess37-multiple-participants.md
Chris-Hibbert: did this show up as undefined on line 33?

smart-contracts-tutorial/guess37-multiple-participants.md
```diff
@@ -0,0 +1,114 @@
...
+
+`playerInvite` is a [payment object](https://github.com/Agoric/ERTP/blob/91aab6abe0f0d9db61f0ac3b858ba6c310410aa5/core/mint.chainmail#L136-L154)
+
+
+`playerInvite~.getBalance()` provides informations about the invite 
+
+```
+TODO
+Figure out how to:
+- guarantee this is a genuine invite produced by the contract host (via comparison with the invite assay?)
```
Chris-Hibbert: The fact that the contract host redeemed the invite is what ensures that the seat is a real one. Before doing that, you could call contractHost~.getInviteAssay()~.claimAll(invite) to prove that it was issued by the assay. If you wanted to verify the provenance before passing it to someone else, this would make sense. Otherwise redeem() is usually enough.

smart-contracts-tutorial/guess37-multiple-participants.md
```diff
@@ -0,0 +1,114 @@
...
+- retrieve the source code and read it
```
Chris-Hibbert: Not sure what you had in mind with this comment, but I wondered whether it would be useful to have the reader print out the contract text, either from host~.getInstallationSourceCode(installation), or from home.guess37ContractSource.

smart-contracts-tutorial/guess37-multiple-participants.md
DavidBruant: no but line 33 is the contract installer role and line 64 is the game player role. They are 2 different solo nodes, accessed through 2 different tabs (different ports), so are 2 different contexts
I would imagine REPL variables are independent (i haven't verified). At least, i will document as if they were :see_no_evil:

smart-contracts-tutorial/guess37-multiple-participants.md
DavidBruant: > The fact that the contract host redeemed the invite is what ensures that the seat is a real one

Yes, but that also "consumes" the invite. If you've been handed the object, you may want to invest time studying the terms and contract source code. However, there exists a risk that the object is not a genuine invite and you're wasting time studying a fake invite
Hence the need to be able to verify the object you're holding is a genuine invite before redeeming

> Before doing that, you could call contractHost~.getInviteAssay()~.claimAll(invite) to prove that it was issued by the assay. If you wanted to verify the provenance before passing it to someone else, this would make sense

oh cool, i think that was the answer i was looking for, thanks! I'm puzzled by the name `claimAll`. It also seem to be doing a lot more than i expect (return `true` only if the argument is a genuine invite) but i'll take it for now

My first guess was to do something like `contractHost~.getInviteAssay() === invite~.getArray()`, but it returned `false` in the REPL (`contractHost~.getInviteAssay() === contractHost~.getInviteAssay()` also returns `false` so i stopped that track). This guess assumes only the contract host can mint invites and i didn't verify this assumption

smart-contracts-tutorial/guess37-multiple-participants.md
DavidBruant: Yes exactly, i was thinking `host~.getInstallationSourceCode(installation)`

`home.guess37ContractSource` is not part of ERTP and is a temporary hack for code upload, so i'm leaving it out of the explanation
Also, for the purpose of studying the invite and the underlying contract, i think it's more reliable to retrieve the source code from the (trusted) contract host since even the contract installer shouldn't be able to corrupt it

smart-contracts-tutorial/guess37-multiple-participants.md
DavidBruant: I understand a bit better ERTP mint and i realize `contractHost~.getInviteAssay() === invite~.getAssay()` wouldn't work

barbaraliau:
@Chris-Hibbert @DavidBruant is this complete? If so, I will merge and place the files into the appropriate place for the doc site.

Chris-Hibbert:
It's ready.

DavidBruant:
Thank you for merging this PR

PR #26, opened 2019-10-28
Move zoe docs over from ERTP

katelynsills:
Feel free to let me know if we need any changes or if these aren't in the right place. Some of these are still in progress but wanted to get them over here pronto.


barbaraliau:
@katelynsills I know Zoe is technically under ERTP, but would it make sense
to have it under its own namespace e.g. /zoe/guide/?

katelynsills:
That sounds good to me! I'll move the files

katelynsills:
@barbaraliau Moved the files, let me know if this looks right!

barbaraliau:
Perfect. I'm going to merge this in so I can include it in the next update that refactors the navigation.

PR #29, opened 2019-10-29
AssetDesc to units, plus some clarification

katelynsills:
This PR renames `assetDesc` to `units`, and `descOps` or `assetDescOps` to `unitOps`. 

Also, whenever I saw something that we've changed or I thought could be clarified, I went ahead and changed it.  Anytime that I saw `label.description` used, I changed it to `label.allegedName` and included warnings about not relying on it, but we could probably use another pass to make sure that is covered everywhere.


barbaraliau:
LGTM. Fixed a couple places where 'description' was still there and changed to 'allegedName'.

PR #31, opened 2019-10-31
starting point for documenting testnet development environment

Chris-Hibbert:
What else is needed  in order to connect this in to navigation.

This is a work in progress, and some of the instructions cover facilities that are still being built, so it's better to get it merged so others can see it and fix it.


barbaraliau:
@Chris-Hibbert I'll get the nav changes needed and push them up

PR #38, opened 2019-10-31
Update mint.mint examples for clarity, add missing method

### LINKED ISSUE #36: mint method of Mint is not documented
I also proposed to rename this API, see https://github.com/Agoric/ERTP/issues/206

barbaraliau:
Updates examples so we don't have `mint.mint()` everywhere.

Add missing `mint.mint()` method in API. Added an explanation as to why the `.mint()` method creates a purse, which is a bit unintuitive at first.

Closes #36

### DIFF
--- a/main/ertp/api/assay.md
+++ b/main/ertp/api/assay.md
@@ -85,8 +85,8 @@ Make an empty purse associated with this kind of right.
 ```js
 import { makeMint } from './core/mint';
 
-const mint = makeMint('fungible');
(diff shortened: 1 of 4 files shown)


barbaraliau:
@katelynsills all good. you got this fast actually. you approved updating the methods (the issue was in the ERTP repo. I wanted you to glance over the explanation though, to make sure it made sense.

katelynsills:
yup, that all looked good to me!

PR #40, opened 2019-10-31
Add ERTP glossary

barbaraliau:


main/ertp/glossary/README.md
```diff
@@ -31,10 +54,13 @@ An [AssetHolder](#assetholder). Purses hold verified units of certain rights iss
 An [AssetHolder](#assetholder). Payments hold verified units of certain rights issued by Mints, specifically units that are in _transit_. Units from payments can be deposited in [purses](#purse), but otherwise, the entire unit is available when the payment is transferred. Payments can be converted to Purses. [Payment API](/ertp/api/mint.html#payment)
 
 ## Unit
-Label (allegedName + assay) + Extent
+Units are the canonical description of tradeable goods. They are manipulated by mints, and represent the goods and currency carried by purses and payments. They can be used to represent things like currency, stock, and the abstract right to participate in a particular exchange. Units are wrappers on extents that have been validated by an UnitOps, and can be verified as having been issued by the UnitOps.
```
katelynsills: This part is outdated. We can manually create Units by putting together labels and extents, and we've taken out the `vouch` method on UnitOps, so we shouldn't say anything about validation or verification. I think this description is the latest https://github.com/Agoric/Documentation/blob/master/main/ertp/api/unit-ops.md#unitops

main/ertp/glossary/README.md
```diff
@@ -1,25 +1,48 @@
 # ERTP Glossary
 
+## Alleged `<item>`
+`allegedName`, `allegedExtent`, `allegedUnits`
+
+Because there is no way to actually prove the item you are calling or passing through is what it says it is, we prepend `alleged` to these terms to get that sentiment across.
```
Chris-Hibbert: I would phrase this in terms of someone who receives a payment from someone they don't %100 trust. 

"Because there is no assurance when you receive an object from someone that it is the thing you were hoping to get, we refer to them as alleged until we confirm the provenance."  Maybe provenance isn't the word you watn to use.

main/ertp/glossary/README.md
barbaraliau: I see. Will remove.

main/ertp/glossary/README.md
barbaraliau: What about:

"Because there is no assurance when you receive an object from someone that it is the thing you were hoping to get, or vice versa, we initially refer to them as alleged."

main/ertp/glossary/README.md
katelynsills: Awesome, thanks!

main/ertp/glossary/README.md
katelynsills: I like that!

Issue #32, opened 2019-10-31
Todos before defi hackathon


barbaraliau:
Putting this in a more public place than my notebook. This is a living issue so I don't make `n` number of issues for every single one.

"Claim" an item by putting your name in front of it in bold and the status.

**Homepage**
- [x]  Update so CTA goes to Getting Started

**ERTP**
- [x] Complete ERTP glossary
  - path: `/ertp/glossary/README.md`

- [x] Move `ExtentOps` API into own file and place underneath UnitOps in the sidebar
- [x] Pull Assay, Payments, Purse sections into own files
  - [x] Assay
  - [x] Payments
  - [x] Purse
- [x] Pull out definitions from `UnitOps` and place into `/ertp/guide/`. Makes much more sense to put it there. API should only contain the methods.
- [x] Pull out 'Gotchas' from ERTP guide and put into own sections
- [x] Complete any missing examples in ERTP
- [x] Add the following methods to Payments API - ```A payment's balance can only fall, through the action of `depositExactly()`, `claimExactly()` or `burnExactly()`. Payments can be converted to Purses by getting a verified assay and calling `assay.makeEmptyPurse().depositAll(payment)`;```
- [ ] **BAR: starting** Finish `default-configuration` file
- [ ] **DOM: PR pending. Almost complete. Have questions/comments (see red text).** ContractHost API [contractHost.chainmal)(https://github.com/Agoric/ERTP/blob/master/core/contractHost.chainmail)

- [ ] Finish `unitOps.make(allegedExtent)` params definition
- [ ] line 106 `/ertp/api/contract-hosts.md` - write and verify example - Chris
- [ ] line 128 `/ertp/api/contract-hosts.md` - verify description -Chris

**Zoe**
- [ ] **KATE: STARTING** Complete Zoe glossary
  - path: `/zoe/glossary/README.md`
- [x] **DOM: in progress** zoe.chainmail -> Zoe API
- [ ]  Check this: `Zoe and ZoeContractFacet have some similar methods so those examples are the same. Those should be double-checked to make sure they are accurate in both locations. Methods: getInviteAssay, getEscrowReceiptAssay`
- [ ] verify `zcf.getSeatAssay()` example - Couldn't find an example in the code, this needs checking.
- [ ] verify `zoe.getInviteAssay()`

**Add Other Projects**
[...]

PR #41, opened 2019-10-31
Completes some documentation tasks

### LINKED ISSUE #47: Undeclared module in assay.makeUnits doc
https://github.com/Agoric/Documentation/blob/master/main/ertp/api/assay.md#assaymakeunitsextent

```js
import { setup } from '../setupBasicMints';

const { assays: originalAssays, mints, unitOps } = setup();
```

It's not clear to me where the '../setupBasicMints' module comes from

dominiquejane:
Completes some tasks in #32 
- Update so CTA goes to Getting Started
- Pull out 'Gotchas' from ERTP guide and put into own sections
- Clarify descriptions in Payments and Purse API
- Adds ContractHost Guide
- Adds ContractHost API -> incomplete: still have questions/comments (see red text)

Closes #47


main/ertp/guide/gotchas.md
```diff
@@ -0,0 +1,23 @@
+# Gotchas
```
barbaraliau: Let's change this title to "Other Concepts"

main/ertp/guide/gotchas.md
```diff
@@ -0,0 +1,23 @@
...
+When looking at the code in our [tests](https://github.com/Agoric/ERTP/tree/master/test), you might see some new
+concepts:
+
+* __Vats__: All user code runs in what we call a `vat`. Within a vat,
+  code is run synchronously. Communication with another vat happens
+  asynchronously. The [SwingSet
+  infrastructure](https://github.com/Agoric/SwingSet) creates the vats
+  and makes communication between vats possible.
+
+* __E() and infix bang (!)__: Instead of `obj.foo()`, we can write
```
barbaraliau: I believe this is old syntax and is now wavy dot ` ~.`

@katelynsills can you confirm that?

main/.vuepress/config.js
```diff
@@ -65,7 +65,8 @@ module.exports = {
             '/ertp/guide/label',
             '/ertp/guide/unit-ops',
             '/ertp/guide/default-configuration',
-            '/ertp/guide/contract-hosts'
+            '/ertp/guide/contract-hosts',
+            '/ertp/guide/gotchas'
```
barbaraliau: update to `/ertp/guide/other-concepts`

main/ertp/api/contract-hosts.md
```diff
@@ -0,0 +1,158 @@
...
+const invitePayments = E(someInstallation).spawn(terms);
+```
+
+## installation.checkUnits(installation, inviteUnits, terms)
+- `installation` `{Installation}`
+- `inviteUnits` `{Units}`
+- `terms` `{Terms}`
+- Returns:
+
+### <span style="color:red">What does this method return? It's not stated in the chainmail file and it's not very clear in the code usage.</span>
```
barbaraliau: @katelynsills dominique used this to help identify questions with some of these. I'm not very sure myself as I get most of the information from the chainmail file and source code. Can you clarify all the questions identified with the `<span style="color:red">` font? It's easy to see if you preview the site. thanks!

main/ertp/guide/gotchas.md
katelynsills: Yes, you're right. I had updated this text on the ERTP github but missed it here https://github.com/Agoric/ERTP/blob/master/README.md

main/ertp/api/contract-hosts.md
katelynsills: Sure, thanks for marking it up this way. I can try to clarify, but @Chris-Hibbert has been the main person working with contract host. Zoe replaces some of the features of the contract host, so it might be a lower priority to get this part specifically right.

main/ertp/api/contract-hosts.md
```diff
@@ -0,0 +1,158 @@
...
+
+### <span style="color:red">Are these params correct? In the code they follow this pattern instead: `{inviteUnits, terms, anOptionalString}`. Example from `ERTP/test/swingsetTests/contractHost/vat-alice.js`:</span>
```
katelynsills: I think these arguments are good. In this case, I think { left, right} are the terms so this matches the pattern

main/ertp/guide/other-concepts.md
```diff
@@ -0,0 +1,23 @@
...
+When looking at the code in our [tests](https://github.com/Agoric/ERTP/tree/master/test), you might see some new
+concepts:
+
+* __Vats__: All user code runs in what we call a `vat`. Within a vat,
+  code is run synchronously. Communication with another vat happens
+  asynchronously. The [SwingSet
+  infrastructure](https://github.com/Agoric/SwingSet) creates the vats
+  and makes communication between vats possible.
+
+* __E() and infix bang (!)__: Instead of `obj.foo()`, we can write
```
katelynsills: We can update this to ~.

katelynsills: I made some changes to this text on the ERTP github https://github.com/Agoric/ERTP/blob/master/README.md

main/ertp/api/contract-hosts.md
Chris-Hibbert: checkUnits either returns undefined, or throws. It is called to verify that the particular contract being invoked matches the terms the user was expecting. If there's no match, then we must not proceed.

main/ertp/api/contract-hosts.md
dominiquejane: I meant that the chainmail file suggests that the params should be 
`(installation, inviteUnits, terms)`, 
whereas in the code I see a different order/usage of params that look more like
`(inviteUnits, terms, anOptionalString)`.
So it's like the `installation` param got removed and an option String param got added to the  end. So that's what I'm confused about.

katelynsills: Ah gotcha. I'm not sure about which is right. Do you know @Chris-Hibbert

main/ertp/api/contract-hosts.md
dominiquejane: Ok, that makes more sense. Thanks!

main/ertp/api/contract-hosts.md
```diff
@@ -0,0 +1,153 @@
...
+  ```
+
+## contract.start(terms, inviteMaker)
+- `terms` `{Terms}`
+- `inviteMaker` `{InviteMaker}`
+- Returns: `{Object}`
+
+Start execution of the contract. May return invites for seats.
+
+### <span style="color:red">Need help writing/checking example.</span>
```
dominiquejane: @katelynsills Can you check this example?

main/ertp/api/contract-hosts.md
```diff
@@ -0,0 +1,153 @@
...
+// Also, see `installation.spawn()` above.
+```
+
+## inviteMaker.make(seatDesc, seat, name)
+- `seatDesc` `{String}` - Must be unique for each contract instantiation.
+- `seat` `{Object}` - An arbitrary object defined by the contract.
+- `name` `{String}` - Labels the invite payment for this seat. Optional.
+- Returns: `{Payment}`
+
+### <span style="color:red">No description in chainmail file. Please double check that this description makes sense.</span>
```
dominiquejane: @katelynsills Can you check this description?

main/ertp/guide/contract-hosts.md
```diff
@@ -11,3 +13,60 @@ Some seats provide the methods `getWinnings()` and `getRefunds()`, which return
 promises for payments for the outputs of the interaction. We provide the
 `collect()` method (described later) to simplify collection of winnings and
 refunds into appropriate purses.
+
+## Installations
+
+An installation of a Contract can spawn multiple copies each with the same or different terms. Each spawned instance has distinct invites and distinct seats representing a specific group of agents interacting according to the same prescribed roles.
+
+### <span style="color:red">Possible typo on the next sentence?</span>
```
dominiquejane: @katelynsills Do you know what the sentence below is trying to say? I think there's a typo.

"The Installation can also have functions with names starting 'check', as defined by the contract."

main/ertp/api/contract-hosts.md
Chris-Hibbert: Yeah, this is kind of weird, and is another place like the contract seat, where the developer can decide on different parameters for different contract, and different method names. Some contracts will have more than one check...() method that are called differently. [Escrow has checkUnits() and checkPartialUnits()].

The convention I wanted to establish was that the methods would usually be written with their first two parameters being the units and the terms. Other parameters can be added. Escrow adds the seat as an extra parameter, but that's not necessary when there's no distinction between possible roles. (All clients of an auction get the same kind of seat.)

The check function as written in the contract file, has installation, units, and terms. The user who has access to a seat from an invite and wants to validate it, calls it with all but the first parameter, which is curried in by the contract host when it evaluates the check...() function. So the version to be called is different from the version to be written, and I wasn't clear about that.

main/ertp/guide/contract-hosts.md
Chris-Hibbert: The installation can also provide helper functions, useful for validating taht the parameters to the contract match the expectations of the person trying to make use of it. These functions must have names starting 'check' so examples might be checkSeat() or checkAuctionInstance(). The contract gets to define them, and the parameters are specific to the particular contract.

dominiquejane: Thanks for clarifying. I updated the wording:

The Installation can also have functions with names starting with 'check'. These are helper functions that are specific to the particular contract (E.g. Escrow has checkUnits() and checkPartialUnits()).

main/ertp/api/contract-hosts.md
dominiquejane: Thanks for the clarification! I updated some of the wording and added emphasis on parameter order and the difference between calling the functions on the dev side vs user side.

main/ertp/api/contract-hosts.md
barbaraliau: adding to list for later

main/ertp/api/contract-hosts.md
barbaraliau: moved to issue to do later.

PR #42, opened 2019-11-01
Remove ERTP toc

barbaraliau:


barbaraliau:
Merging to do quick check with relative links vs absolute in live site vs local

PR #67, opened 2019-12-06
fix some minor wording issues.

Chris-Hibbert:


DavidBruant:
I'm fine with the wording fixes, but i don't understand the pdf files
Did you mean to link to them? Are they here in preparation of another work?

Chris-Hibbert:
Thanks for catching that. Those were certainly not intended to be included.

PR #68, opened 2019-12-09
Non-fungible asset guide

DavidBruant:


main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
+# Creating a specific eright with ERTP
```
katelynsills: Let's call it "non-fungible" rather than "specific". The cryptocurrency world has really embraced the term "non-fungible" - NFTs (non-fungible tokens, aka "nifties") are what people are calling them.

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Definition of "specific"
+
+In the context of ERTP and assets, "specific" means "non-fungible".
+
+Assets of the same assay are not fungible if each different item of the assay
+cannot be replaced with another.
+House ownership rights are not fungible, because each ownership right refers to a
```
katelynsills: Specificity is a complicated idea to express. I think "house ownership rights" is an uncommon phrase in English even though it makes sense, so maybe we should use a different example. How about something like:

Assets are fungible if each asset is just as valuable as other assets of the same kind. For instance, dollar bills are fungible because it doesn't matter which bill you use. Each dollar bill is interchangeable with any other. Paintings, on the other hand, are not fungible. Paintings have drastically different values depending on the content and painter. A deal involving a specific painting would not be satisfied by another painting.

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+## Definition of "specific"
+
+In the context of ERTP and assets, "specific" means "non-fungible".
+
+Assets of the same assay are not fungible if each different item of the assay
+cannot be replaced with another.
+House ownership rights are not fungible, because each ownership right refers to a 
+specific house and one house is not equal to another.
+
+## Creating a fungible asset with ERTP
```
katelynsills: fungible -> non-fungible

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Creating and sending the asset
+
+In ERTP, digital assets are created by a [`mint`](./mint.html). Having access to the mint
+gives you the power to create more digital assets of the same type at
+will.
+
+Let's say we own an Opera and want to sell tickets to seats for ballet shows. Tickets are
```
katelynsills: I love this example!

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+
+import { insist } from '@agoric/ertp/util/insist';
+import { mustBeComparable } from '@agoric/ertp/util/sameStructure';
+
+const insistOptDescription = optDescription => {
+  insist(!!optDescription)`optDescription must be truthy ${optDescription}`;
+  mustBeComparable(optDescription);
+};
+
+function makeInviteConfig() {
```
katelynsills: Let's change this name to "ticketConfig" or something similar

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+  return harden({
+    ...noCustomization,
+    makeMintKeeper: makeCoreMintKeeper,
+    extentOpsName: 'uniExtentOps',
```
katelynsills: I would have thought this would use the ListExtentOps, so that you can hold a purse with multiple tickets in it. Was there a reason not to? I think the ListExtentOps makes more intuitive sense to beginners, whereas the uniExtents aren't meant to be combined, so they're a bit strange.

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+    extentOpsArgs: [insistOptDescription],
+  });
+}
```
katelynsills: Your suggestion to have the defaults pre-made in `mint.js` is a really good one, given how much has to be imported here that isn't important to the tutorial. Thanks for suggesting that. I'll add it as an issue.

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+```js
+const startDateString = (new Date(2019, 11, 9, 20, 30)).toISOString()
+
+const ticketExtents = Array(1114).fill().map((_, i) => ({
+  seat: i+1,
+  show: 'The Sofa',
+  start: startDateString,
+}))
+```
+
```
katelynsills: Do we need to differentiate between extents and units here? Could we instead just focus on units? How about this:

```js
const startDateString = (new Date(2019, 11, 9, 20, 30)).toISOString()
const label = balletTicketMint.getAssay().getLabel();

const ticketUnits = Array(1114).fill().map((_, i) => ({
  label, 
  extent: {
    seat: i+1,
    show: 'The Sofa',
    start: startDateString,
  },
}))
```

Then we could just mention that extents are the non-labeled portion.

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+Now these ticket descriptions need to be turned into **units**. While extents can be thought of as data-only objects, [units](https://agoric.com/documentation/ertp/guide/units.html) are objects that represent goods that can be minted and exclusively shared.
+
```
katelynsills: > While extents can be thought of as data-only objects, [units](https://agoric.com/documentation/ertp/guide/units.html) are objects that represent goods that can be minted and exclusively shared.

I think this is not quite correct. Both extents and units are measurements of digital assets, so they aren't different on that dimension. Also, although units aren't easily turned into JSON because they include assays, neither are extents (extents can have complex objects in them that would be turned into presences in our system), so they don't differ there either.  Really the only difference between them is that units are labeled and extents are the unlabeled version.

main/ertp/guide/create-specific-erights.md
```diff
@@ -0,0 +1,87 @@
...
+const ticketUnits = ticketExtents.map(ticketExtent => balletTicketUnitOps.make(ticketExtent))
+```
+
+And now, let's mint tickets!
+
+```js
+const balletTicketPurses = ticketUnits.map(ticketUnit => balletTicketMint.mint(ticketUnit))
+```
+
+For each ticket unit, we've created a [purse](https://agoric.com/documentation/ertp/api/purse.html) which contains the corresponding unit. These purses can later be transformed into [payments](https://agoric.com/documentation/ertp/api/payment.html) via a call to `balletTicketPurse.withdrawAll()`. This payment can later be traded against other payments or on conditions decided by the payment holder.
```
katelynsills: > traded against other payments

This wording isn't clear as to me as it could be. Could we say the payments can be bought and sold and used in contracts?

katelynsills: Minting payments will be so helpful here.

DavidBruant:
> (We shouldn't wait for those changes before merging this in, to be clear.)

Agreed, i feel like the doc will be easy enough to change when those changes happen
For me this opens the question of whether the ERTP doc source should be part of the monorepo
This way, changes to code, tests and documentation could happen within the same pull request

main/ertp/guide/create-specific-erights.md
DavidBruant: ok, thanks for the clarification. I'm still new to the cryptocurrency world and this is something i hadn't caught on yet

main/ertp/guide/create-specific-erights.md
DavidBruant: ahah
i wasn't sure this would be accepted, i'm glad this is your reaction ^^

main/ertp/guide/create-specific-erights.md
DavidBruant: :+1:

main/ertp/guide/create-specific-erights.md
DavidBruant: I copied the code from the invite mint from contractHost and didn't question it enough
I agree with you, ListExtentOps makes more sense

DavidBruant: that said, `listExtentOps` is not exposed in 
https://github.com/Agoric/agoric-sdk/blob/e4b7c868858329928c7fb25f4cac881d81458a99/packages/ERTP/core/config/extentOpsLib.js

And in the current ERTP code, a UnitOps can only ever be created from things exposed in extentOpsLib.js

https://github.com/Agoric/agoric-sdk/blob/e4b7c868858329928c7fb25f4cac881d81458a99/packages/ERTP/core/unitOps.js#L4
+
https://github.com/Agoric/agoric-sdk/blob/e4b7c868858329928c7fb25f4cac881d81458a99/packages/ERTP/core/unitOps.js#L24-L28

But listExtentOps cannot really be exposed as is, since it requires to provide 3 methods to be instanciated

https://github.com/Agoric/agoric-sdk/blob/e4b7c868858329928c7fb25f4cac881d81458a99/packages/ERTP/core/config/extentOps/listExtentOps.js#L9-L11

I'm making an issue on this topic

DavidBruant: Issue: https://github.com/Agoric/agoric-sdk/issues/330

while writing the issue, i realized that it's not possible to use ListExtentOps without changing first how the mint API works

I'm keeping `uniExtentOps` for now, pending a discussion on ListExtentOps

main/ertp/guide/create-specific-erights.md
DavidBruant: > Could we instead just focus on units?

Now that i better understand units and why they exist, i think i agree
But i would remove the notion of extent altogether from this guide then. My belief is that the less notions introduced per article the better

In the same vein, I don't think showing the `{label, extent}` object provides any value to the reader
I would do:
```js
const ticketUnits = Array(1114).fill()
  .map((_, i) => balletTicketMint.getAssay().makeUnits({
    seat: i+1,
    show: 'The Sofa',
    start: startDateString,
  })
)
```

My belief on this is that people modeling ballet tickets want to model `{seat, show, start}` objects because it matches the mental model they have of what they're selling
And then they want to understand how to express this in the ERTP world

For this, they need to know about units because it's what mints take as an argument, and they need to know how to make units from the JavaScript objects they created, but they don't need to know the implementation details of units
So i'm leaving out the `{label, extent}` object from the code sample. Also, i think `makeUnits` is more explicit and more directly conveys the intention

main/ertp/guide/create-specific-erights.md
DavidBruant: Ok, i understand this better now

I'm cutting down content on this

main/ertp/guide/create-specific-erights.md
DavidBruant: Writing this is indeed what triggered [my supporting comment on the corresponding issue](https://github.com/Agoric/agoric-sdk/issues/84#issuecomment-563240907)

The first thing i want to do after having minting is not storing, it's sharing/exchanging what was just minted with someone else
Even if i wanted to burn what i minted because it was by mistake, i would need a payment to do that

DavidBruant:
i addressed comments in the latest commit, but i'm not fully satisfied yet
I'll work a bit more on the unit/extent part

main/ertp/guide/create-specific-erights.md
katelynsills: Agreed

main/ertp/guide/create-specific-erights.md
katelynsills: Ok, that sounds fair. I think `units` are mysterious to people until we break it down into a label + extent (hopefully in the future, an assay + extent), so if we can do that as well, that'd be great.

katelynsills:
Looks great so far; let me know when you want a review

main/ertp/guide/create-specific-erights.md
DavidBruant: @katelynsills github hides this discussion for me, so i'm not sure you've seen it

katelynsills: Ok thanks, I saw it. Sticking with uniExtentOps makes sense to me for those reasons. I'm ok publishing this before we fix ListExtentOps, but that should be a next step

main/ertp/guide/create-specific-erights.md
DavidBruant: I simplified this part by only talking about units. It also links to [units documentation](https://agoric.com/documentation/ertp/guide/units.html#labeled-extents) which explains units in terms for label+extent with a code example
This way, the guide can remain focused on the shortest path from "what the person wants to model" (`{seat, show, start}` objects) to "ERTP goods that can be bought/sold", but people who want to know have a simple way of knowing more about what units are and how they work

main/ertp/guide/create-non-fungible-tokens.md
```diff
@@ -0,0 +1,79 @@
...
+Assets are fungible if each asset is just as valuable as other assets of the same kind. 
+For instance, dollar bills are fungible because it doesn't matter which bill you use. 
+Each dollar bill is interchangeable with any other. Paintings, on the other hand, are not fungible. 
+Paintings have drastically different values depending on the content and painter. 
+A deal involving a specific painting would not be satisfied by another painting.
+
+
+## Creating a non-fungible asset with ERTP
+
+### Moldeing and creating the asset
```
katelynsills: Let's change this to "Creating the asset." I think English-speaking readers will be unfamiliar with Moldeing as a word.

DavidBruant: oh... I think every reader from every language will have difficulty :-p
I meant "modeling" in the broad sense of https://en.wikipedia.org/wiki/Conceptual_model_(computer_science)

Thanks for catching the typo!

katelynsills:
Awesome! Feel free to merge this in.

PR #71, opened 2020-01-28
Agoric CLI updates

michaelfig:
Modify the getting started instructions to use only the `agoric-sdk` version of `agoric-cli`.  We hope to make `npx agoric` also work in the future, but this has fewer dependencies and is more efficient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,9 @@
   "homepage": "https://github.com/Agoric/documentation#readme",
   "dependencies": {},
   "devDependencies": {
+    "vue": "^2.6.11",
+    "vue-loader": "^15.8.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -28,6 +28,9 @@
   "homepage": "https://github.com/Agoric/documentation#readme",
   "dependencies": {},
   "devDependencies": {
+    "vue": "^2.6.11",
```
katelynsills: Is there a reason these are added to devDependencies?

michaelfig: Actually, the only change I needed was to the `package-lock.json`, so I reverted the `package.json` change.

Thanks!

PR #90, opened 2020-03-20
Documentation Update for Zoe v0.3.0

katelynsills:


main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -32,21 +32,13 @@ exit safety. In this case, Alice's exit rule is `onDemand`, meaning
...
-    kind: 'wantAtLeast',
-    units: simoleanAssay.makeUnits(7),
-  },
-  exitRule: {
-    kind: 'onDemand',
-  },
-]);
-const alicePayments = [aliceMoolaPayment, undefined];
+const aliceOfferRules = harden({
+  offer: { Asset: moolaAssay.make(3) },
```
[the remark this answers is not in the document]
katelynsills: Hmm, just realized that the Zoe docs never got updated for the new ERTP. Anything that says `assay` needs to change to `issuer` and this should actually change to 
```suggestion
  give: { Asset: moola(3) },
```
or equivalently:
```
give: { Asset: moolaAmounthMath.make(3) }

File: main/zoe/guide/contracts/atomic-swap.md [hunk 2]
```diff
@@ -89,21 +81,10 @@ make an offer in the same way as Alice, but his `offerRules` match Alice's:
...
-      units: bobAssays[1].makeUnits(7),
-    },
-  ],
-  exitRule: {
-    kind: 'onDemand',
-  }
-);
-const bobPayments = [undefined, bobSimoleanPayment];
+  want: { Asset: bobAssays[0].make(3) },
+  offer: { Price: bobAssays[1].make(7) },
```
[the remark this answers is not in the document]
Chris-Hibbert: The parallelism would be clearer if Bob also used `simoleans()` and `moola()`.

main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -123,8 +104,8 @@ moolaPayment is empty, and the simoleanPayment has a balance of 7.
 The same is true for Bob, but for his specific payout.
```
Chris-Hibbert: >> Now that Bob has made his offer, Alice's `payout` resolves to an array	

I'd say 

"Now that Bob has made his offer, the contract executes and Alice's `payout` resolves to an array	" or something similar so people don't get the idea that Bob's offer normally gets funnelled straight through to Alice. It's just atomic-swap.

katelynsills: Yeah good point. We need to replace the use of "array" in that sentence too. It's a record with keyword keys now.

dominiquejane: @katelynsills 
Further into the sentence it provides an array. Would we update that to:
`{ Asset: moola.make(0), Price: simoleans.make(7) }` or
`{ Asset: 0, Price: 7 }` or something else?

katelynsills: Let's update it to `{ Asset: moola(0), Price: simoleans(7) }`. We've been using the names as a shorthand, assuming that somewhere earlier in the code we have done:

```
const moola = moolaAmountMath.make;
const simoleans = simoleanAmountMath.make;
``` 

In fact, let's add the above snippet to the example if we can to make it very clear what we are doing.

main/zoe/api/records.md
```diff
@@ -0,0 +1,151 @@
...
+```js
+{ 
+  handle, // opaque identifier for the offer, used as the table key
+  instanceHandle, // opaque identifier for the instance
+  proposal, // the offer proposal (including want, give, exit)
+  amounts, // the amountKeywordRecord that will be turned into payouts
+}
+```
+
+## Issuer Record
```
erights: This must never be exposed outside Zoe because it holds the actual purses. I just checked, and it isn't. Therefore it should not appear in user-facing documentation. It is purely part of Zoe's private implementation.

main/zoe/api/records.md
```diff
@@ -0,0 +1,151 @@
...
+
+These records (JavaScript objects with data properties and no
+functions) are used throughout Zoe.
+
+## SeatAndPayout
+
+This is returned by a call to `redeem` on Zoe. A seat is an arbitrary object whose methods allow the user to take certain actions in a contract. The payout is a promise that resolves to an object which has keywords as keys and promises for payments as values. Note that while the payout promise resolves when an offer is completed, the promise for each payment resolves after the remote issuer successfully withdraws the payment.
+
+```js
+someSeatAndPayout: {
```
erights: Using property-colon notation both on the outside, where the name is only descriptive, and on the inside, where usually is actually a literal name you might use, is confusing. It is too confusing where you use `someExitRule:` below, because the only property name that would actually appear there under normal use, and under the only use you explain, is `exit:`

main/zoe/api/records.md
```diff
@@ -0,0 +1,151 @@
+# Records
+
+<Zoe-Version/>
+
+These records (JavaScript objects with data properties and no
+functions) are used throughout Zoe.
+
+## SeatAndPayout
+
+This is returned by a call to `redeem` on Zoe. A seat is an arbitrary object whose methods allow the user to take certain actions in a contract. The payout is a promise that resolves to an object which has keywords as keys and promises for payments as values. Note that while the payout promise resolves when an offer is completed, the promise for each payment resolves after the remote issuer successfully withdraws the payment.
```
erights: Need to explain keywords, and the syntactic requirements on keywords. If this is already explained elsewhere, link to it from here.

main/zoe/api/records.md
katelynsills: Good point! Thanks!

tyg: On the other hand, maybe we should have a private internal documentation repo for this kind of thing which'd be useful for future hires or people moving onto Zoe, but that shouldn't be public-visible? I don't want to see possibly useful info lost.

PR #93, opened 2020-03-27
Before using Agoric software

tyg:


Before-Using-Agoric.md
```diff
@@ -0,0 +1,76 @@
...
+  <th><b>Action</b></th>
+  <th><b>Explanation</b></th>
+  <tr>
+    <td><center>1</center></td>
+    <td><a href="https://nodejs.org/">Install Node.js</a>, version 10.16.0 or higher</td>
+    <td></td>
+  </tr>
+  <tr>
+    <td><center>2</center></td>
+    <td><a href="https://classic.yarnpkg.com/en/docs/install#mac-stable">Install Yarn 1</a></td>
```
Chris-Hibbert: "install yarn" isn't enough instruction. Tell them how to do it with some particular package manager, or what to download and install. Yeah, doing this in a platform-neutral way isn't easy. Maybe that means we need a different version for Mac, Windows, Linux, or an intro paragraph giving directions so everyone can start in the same place.

AFAIC, pick a random installer, and anyone who has a different favorite is sophisticated enough to figure out the differences.

tyg: "Install yarn" isn't sufficient, but the link is. It goes to what amounts to yarn install central, where you can select one of around 8 operating systems to get instructions and downloads needed to install for that OS. I'll clarify in the explanation column about the link.

Before-Using-Agoric.md
```diff
@@ -0,0 +1,79 @@
...
+    <td>Yarn is a package manager for your code and lets developers
+  share code with others. Code is shared via a <i>package</i> (also
+  called a <i>module</i>) that contains all shared code and a
```
katelynsills: Hmm, I think this is not quite right. Packages and modules are two different things - packages are made of modules.

Before-Using-Agoric.md
```diff
@@ -0,0 +1,79 @@
...
+  <tr>
+    <td><center>6</center></td>
+    <td>Run <code>yarn install</code></td>
+    <td>Install NPM dependencies.</td>
+  </tr>
+  <tr>
+    <td><center>7</center></td>
+    <td>Run <code>yarn build</td>
+    <td>Build sources that need compiling. <b>Note:</b>
+  <code>build</code> is not a standard <code>yarn</code> parameter,
```
katelynsills: I thought this note was very helpful

Before-Using-Agoric.md
```diff
@@ -0,0 +1,79 @@
...
+  but one installed with the Agoric SDK.</td>
+  </tr>
+  <tr>
+    <td><center>8</center></td>
+    <td>Install the Agoric CLI by running: <code>yarn link-cli 
+      &lt;<i>agoric-cli's install directory</i>&gt;</td>
+    <td>Select a directory to install the Agoric CLI in. For example,
+      <code>yarn link-cli
```
katelynsills: Do we expect this step and the next step to change?

michaelfig: @tyg The parameter is not `agoric-cli's install directory`, it is `filename of new cli script to create`.  I unfortunately do not have a really good way of explaining this.

Remind me to rename (or at least create an alias of) `link-cli` to `create-sdk-cli` and update these docs at the same time.

michaelfig: Issue https://github.com/Agoric/agoric-sdk/issues/931 created on agoric-sdk.

PR #112, opened 2020-04-08
Intro to Zoe

tyg:
Shorter Getting Started introduction to how Zoe works and how to use it.


main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+## How do Zoe contracts work?
+
+
+To start using a contract, you need to *install* it on Zoe. This is putting the contract’s code into Zoe. Multiple contracts can be installed on Zoe, and each contract can have multiple simultaneous active offers, each of which is handled by a separate contract instance. For clarity, in this document a *contract* is the installed code that defines how the contract works. A *contract instance* is an active execution of the contract code.
+
+
+Once installed on Zoe, a contract waits for an *offer* to be made. That starts a lifecycle for the offer and new contract instance of:
+1. Sending out *invites* to request counter-offers.
+2. Partie *redeeming* those invites and making a counter-offer.
```
erights, line 24: I have never seen the word "Partie" before.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+*Zoe* is itself a long lived and trusted contract that serves as a framework which runs other contracts.
+
+
+## How do Zoe contracts work?
+
+
+To start using a contract, you need to *install* it on Zoe. This is putting the contract’s code into Zoe. Multiple contracts can be installed on Zoe, and each contract can have multiple simultaneous active offers, each of which is handled by a separate contract instance. For clarity, in this document a *contract* is the installed code that defines how the contract works. A *contract instance* is an active execution of the contract code.
+
+
+Once installed on Zoe, a contract waits for an *offer* to be made. That starts a lifecycle for the offer and new contract instance of:
```
erights, line 22: Actually, the contract must first be instantiated using `makeInstance`. It is only a *contract instance* that waits for an offer to be made.

There is an idea for a future design that would fold together initial offer and instantiation. But that's not what we have now.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+1. Sending out *invites* to request counter-offers.
```
erights, line 23: An offer is only ever made using an invite. Instantiation (`makeInstance`) returns the first invite, which is an invite to make an offer to that instance.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+2. Partie *redeeming* those invites and making a counter-offer.
+3. Determining when the offer *matches* a counter-offer.
+4. Either succeeding or failing to make the transaction. 
+
+
+An offer’s completion only affects that offer and the assets involved in it; essentially its contract instance. The contract it was made on keeps running, waiting for and processing new offers. 
+
+
+*Offer safety* is automatically enforced for parties in a contract; if an offer succeeds and completes, all parties get what they wanted. If it fails to complete successfully for any reason, parties keep what they offered.
```
erights, line 32: Better to say "parties are refunded what they offered". Saying "keep" implies that they never lost possession of it.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+All parties in a Zoe contract receive a *seat*, which lets them take action in that contract. They also receive a *promise* for a future payout. The contract itself does not have direct access to any parties’ assets. The payments for an offer are all run through Zoe and its *escrow* capabilities.  
+
+
+Zoe has two distinct APIs. One is client/user-facing, with `zoe.` prefixes on its commands. The other is only exposed to contracts built on Zoe (its *contract facet*), with `zcf.` prefixes on its commands. .
```
erights, line 38: The `zoe.` "prefix" is only usable by tests in the same vat, which is not possible for actual usage. Real usage is from different vats sending remote messages, which will be "prefixed" by `E(zoe).` or `zoe~.`. Also, "prefix" is an odd term for this. Should instead emphasize the oo perspective, where this designates the receiver of the message/request/command. I prefer "request" to "command", since it is always up to the object receiving the message how to react.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Running contracts on Zoe
+
+
+First, use `zoe.makeInstance()` to create an instance of the installed contract you want to use. The method takes as arguments:
+- the `installationHandle` of the specific contract to use.
+- a *keyword record* of name : value pairs.
+  - The names are words used in the contract, such as “Asset” or “Bid” that, in the contract, take an amount as its value. 
+  - The values are references to an *issuer* (see Introduction to ERTP **tyg todo: add link) of the type you want the associated name to accept as its value in the contract instance. For example, if your contract uses the name “Asset” for what the offer in the contract is (e.g. 3) and the resource unit is quatloos, and uses “Price” for the counter-offer, which is in moola, the keyword record value would be ((“Asset” : quatlooIssuer) (“Price” : moolaIssuer)).
```
erights, line 57: (Missing close "**" causes loss of boldface.)

erights, line 57: I don't understand the parens in the example code. Should be
```js
{ Asset: quatlooIssuer, Price: moolaIssuer }
```

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+- *terms*: Another set of key : value pairs.
+  -   These customize contract aspects expressed in the contract without specific values. For example, an auction contract’s may not define what the minimum bid has to be or the minimum amount one has to raise from the current high bid, just that there are such values which the contract uses. These values are different for a collection of used Harry Potter paperbacks and a signed first British hardback edition of *Harry Potter and the Philosopher’s Stone*. If you write code in the contract to access that instance’s `terms`’ name:value pairs of `('MinBid' : 25, 'Currency : 'USDollar', 'MinRaise' : 5)` or  `('MinBid' : 100000, 'Currency : 'USDollar', 'MinRaise' : 5000)`, you’ve customized the generic contract for that specific instance.
```
erights, line 59: parens should be curlies again. Don't need to quote keywords since they are (and must be) identifiers.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+- *terms*: Another set of key : value pairs.
+  -   These customize contract aspects expressed in the contract without specific values. For example, an auction contract’s may not define what the minimum bid has to be or the minimum amount one has to raise from the current high bid, just that there are such values which the contract uses. These values are different for a collection of used Harry Potter paperbacks and a signed first British hardback edition of *Harry Potter and the Philosopher’s Stone*. If you write code in the contract to access that instance’s `terms`’ name:value pairs of `('MinBid' : 25, 'Currency : 'USDollar', 'MinRaise' : 5)` or  `('MinBid' : 100000, 'Currency : 'USDollar', 'MinRaise' : 5000)`, you’ve customized the generic contract for that specific instance. 
+
+
+`zoe.makeInstance()` returns an *invite* to be a party in that contract instance. Thus, the party who made the contract instance has an invite to it. The `instanceHandle`, the unique identifier for the contract instance, can be retrieved from the invite using `GetInstanceHandle()`.
```
erights, line 62: Should show arguments of `zoe~.makeInstance(...)` (or `E(zoe).makeInstance(...)`). Show the arguments where I said "`...`" above.

erights, line 62: > `GetInstanceHandle()`

Do we have an operation with that strange capitalization?

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+## What is Zoe?
+
+
+*Zoe* is itself a long lived and trusted contract that serves as a framework which runs other contracts.
+
+
+## How do Zoe contracts work?
+
+
+To start using a contract, you need to *install* it on Zoe. This is putting the contract’s code into Zoe. Multiple contracts can be installed on Zoe, and each contract can have multiple simultaneous active offers, each of which is handled by a separate contract instance. For clarity, in this document a *contract* is the installed code that defines how the contract works. A *contract instance* is an active execution of the contract code.
```
erights, line 19: > each contract can have multiple simultaneous active offers, each of which is handled by a separate contract instance.

Could easily be misunderstood to mean that there is an instance per active offer.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+Once installed on Zoe, a contract waits for an *offer* to be made. That starts a lifecycle for the offer and new contract instance of:
+1. Sending out *invites* to request counter-offers.
+2. Partie *redeeming* those invites and making a counter-offer.
+3. Determining when the offer *matches* a counter-offer.
+4. Either succeeding or failing to make the transaction. 
+
+
+An offer’s completion only affects that offer and the assets involved in it; essentially its contract instance. The contract it was made on keeps running, waiting for and processing new offers.
```
erights, line 29: There is *not* a contract instance per offer. There is a contract instance per `makeInstance`. Within a contract instance, multiple offers are matched. Only contract instances keep running. A contract (installation) is just code.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
+# Introduction to Zoe
+
+
+**Note:** Before reading this, you should be familiar with ERTP basics. (**tyg todo**: Add link)
+
+
+This is an introduction to Zoe, Agoric's smart contract framework. The Zoe Guide (**tyg todo**: Add link) has more complete and detailed information, including concepts not covered here. In particular, here we simplify Zoe examples such that there are always two parties in a basic trading/swap contract. We do not discuss more complicated contracts with many participants, such as auctions, or simpler contracts, such as our AutomaticRefund one party demo contract.
```
erights, line 7: I'm not sure what to do about this, but the whole explanation in terms of "parties" (or participants) leads to a very identity-oriented understanding of what's going on. Instead, everything attributed to "each party" here is really about each offer. Any one participant can make any number of offers. It may be that referring to parties or participants is a good expository simplification. Or it might amplify the identity-oriented misunderstandings that users of other systems will already be drawn to. Where in our expository sequence to we try to cure them of an identity-oriented pov on participating?

tyg: Hmm. I think it's a bit of a combination of I have a more identity-oriented take than would seem to be correct and that I do think things need to be simplified for the intro; when going over a draft with Kate, she mentioned things like auctions with multiple parties, offers, non-winners getting their offer stakes back even when things go right and the contract completes, etc.  But that's getting away from the goal, which is to give them a basic understanding of what Zoe is/does and our contract execution workflow before they start looking at contracts, which'll start with the most basic ones. 

Curing of identity-orientation would happen in the Zoe Guide, which needs to be redone for many reasons, and/or in comments/exposition on our demo contracts that get away from two parties making an offer. And I need to get myself more in an offer mindset, so I'll try to do at least some rewriting on this to make it moreso.

main/zoe/api/Zoe-Intro.md
tyg: Yeah, I see this. Will rephrase, although I could really do with a good word for what's going on from the time an instance is instantiated and "activated" by getting an offer through when the contract instance finishes and everyone gets whatever they get out of Zoe escrow. "Contract (or contract instance) execution" doesn't seem to be it. Any suggestions?

main/zoe/api/Zoe-Intro.md
tyg: Thanks. Will rewrite.

main/zoe/api/Zoe-Intro.md
tyg: Typo for "parties".

main/zoe/api/Zoe-Intro.md
tyg: I think I did note that in the preceding line, but it's phrased poorly and needs to be clarified. Thanks for the catch.

main/zoe/api/Zoe-Intro.md
tyg: I'm confused here. The zoe. commands are documented in our public documentation, and include things like zoe.makeInstance, which would seem to be used by, well, users rather than tests.

"Prefix" was used since both sets seem to be sent to Zoe, just one of them is sent to a more limited command facet of it. Is there a preferred way of phrasing where the zcf: requests are being sent?

main/zoe/api/Zoe-Intro.md
tyg: Oops. Will fix.

tyg: Was doing a more generic way of writing key:value pairs, will correct to our actual usage.

main/zoe/api/Zoe-Intro.md
tyg: Was doing an abstract example, but you're right that showing what would actually be there is better. Will correct.

main/zoe/api/Zoe-Intro.md
tyg: I'll think about this; in at least some cases for an intro, I just want them to know what's happening and what the request to do it is, putting off the more detailed arguments a bit. 

Quick question; are the calls in fact "zoe~." instead of "zoe."? I don't recall seeing the ~ version, certainly not in the existing docs.

tyg: Typo, will fix, thanks.

main/zoe/api/Zoe-Intro.md
tyg: OK, this is a tricky bit. This was my attempt to address issues that Kate brought up about an earlier draft about how to phrase/describe when things stop/end. So we've got an installed contract, which is the code. That never ends, at least as long as Zoe and/or the server is running. Then we have a contract instance. If I understood Kate correctly, it's not right say that ends, as it's really the first offer within the instance that gets the process going.  So is it "first offer" finishes?  

What I'm trying to ask here is that we have a lifecycle kicked off by a combo of instantiating a contract and creating a contract instance. But that's still just sitting there, not doing anything, until someone takes the invite generated by the instantiation and uses it to make an initial offer, then makes other invites and sends them out, etc.  At some point, things stop because the contract code is satisfied that the transaction is finished and (assume a success) everyone has what they should have at the end.    But I don't know what the appropriate way is to describe that things are done for this execution (in a legal sense) of the contract?

main/zoe/api/Zoe-Intro.md
tyg: Will rephrase, thanks.

katelynsills: I think this will give the reader a wrong assumption that will be very hard to change later. An offer can be a perfectly valid offer, and the contract can still choose to give a refund for that offer, not what the user wanted. 

> if an offer succeeds and completes, all parties get what they wanted

This is generally not a true if/then. It really depends on the contract whether or not a user gets what they wanted or whether the user gets a refund for a valid offer. For an invalid offer (one that is not what the contract expects, but which was at least successfully escrowed with Zoe), the user will usually get a refund because the contract will complete the offer immediately.

> If it fails to complete successfully for any reason, parties keep what they offered.

There's a weird connection here between whether a single offer is completed by the contract with a refund allocation, and whether all parties get a refund. There's no such connection generally. If we want to talk about the cases in which any offer being rejected by a contract results in everyone getting a refund, we should use a specific contract example in this guide.

main/zoe/api/Zoe-Intro.md
katelynsills, line 38: I think talking about prefixes here is wrong - there will never be a case where a user is in the same context and choosing among different objects (zoe or zcf). Instead, there are two different facets, the Zoe Service (what the user of Zoe sees) and the Zoe Contract Facet (what the contract developer can use). In our API documentation, we call the Zoe Service `zoe` and the Zoe Contract Facet `zcf`, but Mark is correct that any calls on the testnet of the Zoe Service API will need to use `E`: `E(zoe).makeInstance()`. The contract facet can be called synchronously within the contracts, like `zcf.getCurrentAllocation`, without using `E`. But, again, I think we don't need to mention what the objects are called here (the prefixes), but rather just use the names Zoe Service and Zoe Contract Facet.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+All parties in a Zoe contract receive a *seat*, which lets them take action in that contract. They also receive a *promise* for a future payout. The contract itself does not have direct access to any parties’ assets. The payments for an offer are all run through Zoe and its *escrow* capabilities.  
+
+
+Zoe has two distinct APIs. One is client/user-facing, with `zoe.` prefixes on its commands. The other is only exposed to contracts built on Zoe (its *contract facet*), with `zcf.` prefixes on its commands. . 
+
+
+## Starting a contract on Zoe
+
+
+While there can be more than one Zoe instance installed and running at any time, for now we’ll focus on the single instance running on Agoric’s testnet. Of course, before you can run a smart contract on Zoe, you have to install it there. For Zoe purposes, *installation* means installing the code your contract is written in (JavaScript and of course the ERTP and Zoe API commands we’ve defined) on a Zoe instance.
```
katelynsills, line 44: I think we don't need to explain that there could be more than one Zoe instance running. That won't be the case for most basic uses and is confusing.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Starting a contract on Zoe
+
+
+While there can be more than one Zoe instance installed and running at any time, for now we’ll focus on the single instance running on Agoric’s testnet. Of course, before you can run a smart contract on Zoe, you have to install it there. For Zoe purposes, *installation* means installing the code your contract is written in (JavaScript and of course the ERTP and Zoe API commands we’ve defined) on a Zoe instance. 
+
+
+Contract installation returns an `installationHandle`, a unique identifier for the installed contract used to create *contract instances* from the installed code. A contract installation is a way to reuse the same code multiple times and ensure that widely used contract code is recognizable under the same identity. For example, if I negotiate with someone and we agree to use the “Atomicswap” contract **tyg todo link**, if I make an instance of that contract, my counterparty can use its `installationHandle` with the `getInstallation()` API method to check the code is the `Atomicswap` code they expected.
```
katelynsills, line 47: We can call this the "Atomic Swap" contract

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Running contracts on Zoe
+
+
+First, use `zoe.makeInstance()` to create an instance of the installed contract you want to use. The method takes as arguments:
+- the `installationHandle` of the specific contract to use.
+- a *keyword record* of name : value pairs.
```
katelynsills, line 55: Let's call this an "issuer keyword record"

main/zoe/api/Zoe-Intro.md
katelynsills: Great explanation here!

main/zoe/api/Zoe-Intro.md
katelynsills: > These customize contract aspects expressed in the contract without specific values

I think this is too fuzzy. maybe: "Terms allow the creator of a contract Instance to further customize the contract operations, as enabled by the contract code."

> For example, an auction contract’s may not define what the minimum bid has to be or the minimum amount one has to raise from the current high bid, just that there are such values which the contract uses.

Maybe instead: "For example, the code of an auction contract might allow the instantiator to decide the minimum bid, or the minimum amount one has to raise from the current high bid, or the number of bids before the auction closes."

I think the used Harry Potter paperback and signed first edition is a great example of why you might want a different minimum bid. The start of the sentence is confusing to me - maybe instead of "these values are different", it should be "Why do we need to be able to parameterize a contract with terms? Well, our minimum bid would be quite different if the auction is for a used Harry Potter paperback vs a signed first edition..."

main/zoe/api/Zoe-Intro.md
katelynsills: We shouldn't use ~ currently in the docs, since we are not recommending that our users use it until they can use it with VSCode and not get linting errors.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+Using their seat at the contract instance, any user can create new invites to the offer and its contract instance. Once a contract party holds an invite to an offer, they can send it to anyone they'd like to also possibly be a party to the contract. In other words, usually someone who has the asset you want to get. 
+
+
+Finding these other parties to invite is out of Zoe or the contract’s scope in most cases (you could possibly write something into the contract that seeks out potential partners, but that’s way beyond this introduction’s purpose). You have to either already know of potential other parties, for example, from past experience trading with them, or communicate somewhere outside the contract to find them.
+
+
+*Invite*s are ERTP *payments*, so the same principles apply to them as to what you usually think of ERTP payments. You can send the same payment or invite to multiple parties...but the first party to *redeem* it removes the value from the shared payment. As part of redeeming it, they then create their own new payment object with the same amount. Remember, Zoe destroys invites on their redemption.
```
katelynsills, line 75: >  As part of redeeming it, they then create their own new payment object with the same amount.

I think this is confusing ERTP's `claim' method, which is what (under the hood) creates a new payment for a user containing the amount  yanked from the previous (possibly shared) payment. Redeeming `burns` the payment, which means that the value - the amount minted that is stored in the payment - is deleted.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+When you receive an *invite* to an *offer*, you should first ensure it is a valid invitation to the type of contract that you want to participate in.  The second thing to check is the `installationHandle` in the invite: this indicates what particular contract code will be used. You can use the `installationHandle` to make sure it’s a contract you’ve used before, and you can also use `zoe.getInstallation(installationHandle)` to have Zoe send you the contract code itself.
```
katelynsills, line 78: > When you receive an *invite* to an *offer*, you should first ensure it is a valid invitation to the type of contract that you want to participate in

This is the `E(inviteIssuer).claim(invite)` step. We should actually write out this call.

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+If you want to be a party to the contract, *redeem* the invite by providing a *proposal* and a *payment* for the proposal, which Zoe escrows. 
+
+
+*Proposals* consist of:
+- What the user wants (*want key*).
+- What the user offers for what they want (*give key*).
+- How the user can cancel the contract (*exit key*). 
+
+
+The names of the *want* and *give* *keys* are the keywords passed in the keyword record when you created the contract instance. Their values are ERTP amounts of how many of the asset you either want or are willing to give in exchange.
```
katelynsills, line 90: > The names of the *want* and *give* *keys* are the keywords passed in the keyword record when you created the contract instance. 

I think this should be "the keys of the *want* and *give* values are the keywords passed in the issuer keyword record when you created the contract instance". That may be too complicated of a sentence, though, so maybe let's break it up: "The values of `want` and `give` are JavaScript records, objects with keys and values. The keys of these records are the keywords passed in..."

katelynsills: > Their values are ERTP amounts of how many of the asset you either want or are willing to give in exchange. 

"The values of the records are ERTP amounts..."

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+*Proposals* consist of:
+- What the user wants (*want key*).
+- What the user offers for what they want (*give key*).
+- How the user can cancel the contract (*exit key*). 
+
+
+The names of the *want* and *give* *keys* are the keywords passed in the keyword record when you created the contract instance. Their values are ERTP amounts of how many of the asset you either want or are willing to give in exchange. 
+Redeeming an offer returns a *seat* at that contract instance and a *payout*. The *seat* object's methods let the user call API commands on the contract instance. (**Note**: we expect this last to change very soon). The *payout* is what the user gets when their offer is completed; either a refund of what they offered or what they wanted to get from the offer.
```
katelynsills, line 91: > Redeeming an offer returns a *seat* at that contract instance and a *payout*

"and a  promise for the *payout*"

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+Redeeming an offer returns a *seat* at that contract instance and a *payout*. The *seat* object's methods let the user call API commands on the contract instance. (**Note**: we expect this last to change very soon). The *payout* is what the user gets when their offer is completed; either a refund of what they offered or what they wanted to get from the offer. 
+
+
+Once two proposals for a contract match, as determined by the contract instance, the contract instance tells Zoe to *reallocate* the current allocations (allocations = I have 3 brick, you have 2 wool) for those offers, and then completes the offers with Zoe, giving the users their payout. By match we mean that, for example, one party wants three wood and offers two brick for it. The other party wants two brick and offers three wood for it. What they want and what they offer for it sync up.
+
+
+If the offer completes successfully, Zoe creates two new two payments from its escrow, which is stored in Zoe purses. and assigns them to the successful parties. If it doesn't execute successfully, Zoe creates two payments from its escrow and returns them to their original owners.
```
katelynsills, line 97: > If the offer completes successfully, Zoe creates two new two payments from its escrow, which is stored in Zoe purses. and assigns them to the successful parties. If it doesn't execute successfully, Zoe creates two payments from its escrow and returns them to their original owners.

I think this part should be rewritten, because both cases are the same thing in fact. It should be something like:

"When a contract instance completes an offer, Zoe withdraws the current allocation amounts for that offer from its purses (Zoe was holding the value in escrow), and resolves the user's payout promise with these payment. (Note: actually, it is a promise for a payment, but we explain more here, yada yada). It doesn't matter whether the user is getting a refund or getting what they want - the payout mechanism is the same"

main/zoe/api/Zoe-Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+For example, Alice has escrowed two brick, wanting three wood for it, and Bob has escrowed three wood, wanting two brick for it. On the offer’s success, Alice receives the three wood from escrow, while Bob receives the two brick from escrow. On its failure, for whatever reason (buggy code, malicious party, etc.), Alice gets her two brick back and Bob gets his three wood back.
```
katelynsills, line 100: > On the offer’s success

There's no such step. It should be something like "When the contract receives Bob's matching offer, Alice receives the three wood"

> On its failure, for whatever reason (buggy code, malicious party, etc.), Alice gets her two brick back and Bob gets his three wood back.

This should be something like: "if the offer is rejected by the contract, Alice gets her two brick back as a refund." Alice's offer failing has nothing to do with whether Bob gets three wood back. Offers are independent unless the contract specifically connects their outcomes somehow.

main/zoe/api/Zoe-Intro.md
erights: By "minimum bid" do we mean the reserve price?

Normally, an auction would use the seller's `want:` as the reserve price.

main/zoe/api/Zoe-Intro.md
erights: If we're not using `~.`, then it should be
```js
E(zoe).makeInstance(...)
```
It should never be
```js
zoe.makeInstance(...)
```
If it is in our other docs, those need to be corrected to. There are no realistic circumstances in which the user can write the second and have it work.

main/zoe/api/Zoe-Intro.md
erights: @katelynsills by `makeContract` above, did you mean `makeInstance`.

Also, rather than `zoe.make...` it should be `E(zoe).make...`.

katelynsills: Yup, I did mean `zoe.makeInstance`.

DavidBruant:
I think that @erights and @katelynsills reviews are already thorough enough, so i'm removing myself as a reviewer

PR #115, opened 2020-04-10
First draft of contract-installation-and-instance.md

DavidBruant:
This is a first draft to address the last point of https://github.com/Agoric/agoric-sdk/issues/496

I went a bit further than the what the issue mentionned, but i felt it was a good place to talk about the registrar (which i think is mentionned nowhere else in the docs right now)

For now, i'm more interested in feedback on the content:
- is the rationale for 2-step contract creation well-captured?
- is using the registrar the current best practice for sharing installationHandles?
- am i using the registrar properly?

### DIFF
--- a/main/zoe/guide/contract-installation-and-instance.md
+++ b/main/zoe/guide/contract-installation-and-instance.md
@@ -1,0 +1,98 @@
+# Contract installation and instance
+
+Zoe has two methods to "create a contract": 
+- `zoe.install(code, moduleFormat)` which returns an `installationHandle`
+- `zoe.makeInstance(installationHandle, issuerKeywordRecord, terms)` which returns an `invite` to participate in the contract
+
+This two-step process to create may feel unnecessarily convoluted at first.
+
+The reason Zoe is designed this way is that it is assumed that the JavaScript source `code` of the contract may be large (several kilobytes, maybe several megabytes) and for the purpose of transparency, this source code will be stored in blockchains when Zoe runs on a blockchain.
+
+```js
+const mySuperContractSourceCode = ` ... 1 Megabyte string ... `;
+const moduleFormat = 'getExport';
+
+const superInstallHandle1 = zoe.install(mySuperContractSourceCode, moduleFormat); // stores 1Mb in the blockchain
+const invite1 = zoe.makeInstance(superInstallHandle1, issuerKeywordRecord, terms); 
+
+const superInstallHandle2 = zoe.install(mySuperContractSourceCode, moduleFormat); // stores another 1Mb in the blockchain 
+// now 2Mb for the same contract
+const invite2 = zoe.makeInstance(superInstallHandle2, issuerKeywordRecord, terms); 
+
+const superInstallHandle3 = zoe.install(mySuperContractSourceCode, moduleFormat); // stores another 1Mb in the blockchain 
+// now 3Mb for the same contract
+const invite3 = zoe.makeInstance(superInstallHandle3, issuerKeywordRecord, terms); 
+
+// etc...
+```
+
+`installationHandle`s are a solution to prevent a situation where people who want to use the same contract would store a copy of the code several times.
+
+```js
+const mySuperContractSourceCode = ` ... 1 Megabyte string ... `;
+const moduleFormat = 'getExport';
+
+const superInstallHandle = zoe.install(mySuperContractSourceCode, moduleFormat); // stores 1Mb in the blockchain
+const invite1 = zoe.makeInstance(superInstallHandle, issuerKeywordRecord, terms); 
+
+// second contract instance; relies on the source code already stored
+const invite2 = zoe.makeInstance(superInstallHandle, issuerKeywordRecord, terms); 
+
+// still 1Mb in the blochain, but third contract instance
+const invite3 = zoe.makeInstance(superInstallHandle, issuerKeywordRecord, terms); 
+
+// etc...
+```
+
+
+## Using installationHandles
+
+### How can i make sure an installationHandle refers to the correct source code?
+
+Zoe has a `zoe.getInstallation(installationHandle)` handle which returns the source code
+
+```js
+const mySuperContractSourceCode = ` ... 1 Megabyte string ... `;
+const moduleFormat = 'getExport';
+
+const superInstallHandle = zoe.install(mySuperContractSourceCode, moduleFormat); // stores 1Mb in the blockchain
+
+const thisInstallSourceCode = zoe.getInstallation(superInstallHandle)
+
+console.log(thisInstallSourceCode);
+// ` ... 1 Megabyte string ... `
+
+console.log(mySuperContractSourceCode === thisInstallSourceCode);
+// true
+```
+
+
+### Sharing installationHandles with other people
+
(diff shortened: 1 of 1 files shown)


main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
+# Contract installation and instance
+
+Zoe has two methods to "create a contract":
```
katelynsills, line 3:
```suggestion
Zoe has two steps to "create a contract": 
```
Two methods is correct, but it sounds like the user has a choice between them, rather than needing to call both at some point in the process.

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
+# Contract installation and instance
+
+Zoe has two methods to "create a contract": 
+- `zoe.install(code, moduleFormat)` which returns an `installationHandle`
```
katelynsills, line 4: Mark suggests always using `E`, as in `E(zoe).install`, which is probably the right way to go, but we will need to have an aside to an explanation of what `E` is.

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+- `zoe.makeInstance(installationHandle, issuerKeywordRecord, terms)` which returns an `invite` to participate in the contract
+
+This two-step process to create may feel unnecessarily convoluted at first.
+
+The reason Zoe is designed this way is that it is assumed that the JavaScript source `code` of the contract may be large (several kilobytes, maybe several megabytes) and for the purpose of transparency, this source code will be stored in blockchains when Zoe runs on a blockchain.
```
katelynsills, line 9:
```suggestion
Zoe uses a two-step installation process because the JavaScript source `code` of the contract may be large (several kilobytes, maybe several megabytes). Because Zoe is intended to be run on a blockchain, storage is scarce and we want to avoid duplicate copies.
```

^ I think this is missing the second (potentially more important point), that we want people to be able to compare smart contracts easily. Comparing whether an installation handle matches another installation handle (object identity check) is as easy as it gets. The alternative would be asking someone to go line by line or do a git diff or something else to try to ensure that the contract someone is trying to get them to use is one they have used before and trust. 

Imagine that Alice uses the Atomic Swap contract regularly and finds it very useful. Then, Bob wants to do an Atomic Swap with Alice and sends her an invite. How does Alice know that the contract code that Bob uses is the code that she is used to? How does Alice prevent the attack in which most everything looks the same, but an essential condition is changed and she agrees to participate in something that she didn't intend? Having reusable `installationHandle`s solves this problem.

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+const mySuperContractSourceCode = ` ... 1 Megabyte string ... `;
+const moduleFormat = 'getExport';
```
katelynsills, line 13: We should probably explain how to use `bundleSource` here, so that users don't think they have to do this themselves, and so that `moduleFormat` makes a little more sense.

```
import bundleSource from '@agoric/bundle-source';
const { source, moduleFormat } = await bundleSource(contractPath);
```

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+
+const superInstallHandle1 = zoe.install(mySuperContractSourceCode, moduleFormat); // stores 1Mb in the blockchain
+const invite1 = zoe.makeInstance(superInstallHandle1, issuerKeywordRecord, terms); 
+
+const superInstallHandle2 = zoe.install(mySuperContractSourceCode, moduleFormat); // stores another 1Mb in the blockchain 
+// now 2Mb for the same contract
+const invite2 = zoe.makeInstance(superInstallHandle2, issuerKeywordRecord, terms); 
+
+const superInstallHandle3 = zoe.install(mySuperContractSourceCode, moduleFormat); // stores another 1Mb in the blockchain
```
katelynsills, line 22: I think having this much code is maybe over-laboring the point, especially since I think the ability to easily check you are using the contract you expect is the more important reason to have `installationHandle`s. Still, this is very persuasive.

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+// still 1Mb in the blochain, but third contract instance
+const invite3 = zoe.makeInstance(superInstallHandle, issuerKeywordRecord, terms); 
+
+// etc...
+```
+
+
+## Using installationHandles
+
+### How can i make sure an installationHandle refers to the correct source code?
```
katelynsills, line 50:
```suggestion
### How can I make sure an installationHandle refers to the correct source code?
```

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+
+Zoe has a `zoe.getInstallation(installationHandle)` handle which returns the source code
```
katelynsills, line 52:
```suggestion
Zoe has a `E(zoe).getInstallation(installationHandle)` handle which returns the source code
```

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Sharing installationHandles with other people
+
+If you write a smart contract and want people to know about it, you can store the installationHandle in a place where people can find it. In the current Agoric stack/testnet, the proper way would be to create a registrar entry
+
+```js
+const mySuperContractSourceCode = ` ... 1 Megabyte string ... `;
+const moduleFormat = 'getExport';
+
+const superInstallHandle = zoe.install(mySuperContractSourceCode, moduleFormat); // stores 1Mb in the blockchain
```
katelynsills, line 78:
```suggestion
const superInstallHandle = await E(zoe).install(mySuperContractSourceCode, moduleFormat); // stores 1Mb in the blockchain
```

main/zoe/guide/contract-installation-and-instance.md
```diff
@@ -0,0 +1,98 @@
...
+
+const key = registrar.register('the-super-contract', superInstallHandle);
```
katelynsills, line 80:
```suggestion
const key = await E(registrar).register('the-super-contract', superInstallHandle);
```

katelynsills: FYI The registrar will be changing names soon https://github.com/Agoric/agoric-sdk/issues/827

Chris-Hibbert:
I'm going to wait to comment until Kate's suggestions are incorporated.

There was an attempt to document many of the facilities in home and the wallet in 
https://github.com/Agoric/documentation/blob/master/main/manual-setup/pixel-demo.md

They were in the context of the pixel demo, but there are (short) descriptions there of timerService, registrar, sharingService and uploads there, which probably ought to be transitioned to the current environment.

katelynsills:
@DavidBruant would you be able to push this PR forward with @tyg? I think this is good stuff to include

katelynsills:
Closing as outdated because the signature of zoe.install, and zoe.startInstance have changed. Registry is no longer used.

PR #142, opened 2020-04-20
Update before-using-agoric.md

tyg:


main/getting-started/before-using-agoric.md
```diff
@@ -16,7 +16,7 @@ software, you need to install the following.
   </tr>
   <tr>
     <td><center>2</center></td>
-    <td><a href="https://classic.yarnpkg.com/en/docs/install#mac-stable">Install Yarn 1</a></td>
+    <td><a href="https://classic.yarnpkg.com/en/docs/install#mac-stable">Install Yarn</a></td>
```
katelynsills, line 19: Installing Yarn 1 rather than Yarn 2 is important. We shouldn't get rid of this.

PR #155, opened 2020-04-20
First draft of contract requirements

### LINKED ISSUE #99: documentation on what's required for a zoe contract
We should write up a description of the format and expectations for a zoe contract. We currently have a description of what contracts can expect from zoe, but not what zoe expects from contracts, and what choices contracts can make in their implementation

The starting point is that a zoe contract is a javascript file that can import other javascript code, including  `harden`, `Nat`, `zoeHelpers` and anything else we make available. We should describe whether there's any limitation on what library code they can use, and how they can split their code across multiple files.

The file needs to export a makeContract function that takes zoe as a parameter, and return a record that contains at least `invite:`. We should describe the conventional use of the `publicAPI:` keyword and say whether there's any particular support for it. What else needs to be written up about makeContract's return value?

We should describe what you do to create an invite, and what's required/expected of the embedded seat. I think it makes sense to have sections describing the case where instantiating the contract hands out a particular seat that provides access to other seats versus contracts that expect all the seats to be handed out from the publicAPI, or something similar.

There's probably more. It might be worthwhile to provide a skeleton template if enough of this is always the same.

katelynsills:
Closes #99

### DIFF
--- a/main/zoe/guide/contract-requirements.md
+++ b/main/zoe/guide/contract-requirements.md
@@ -41,15 +41,19 @@ export const makeContract = harden(zcf => {
 });
 ```
 
-`zcf` is the <router-link
-to="/zoe/api/zoe-contract-facet.html">Zoe Contract Facet</router-link>. The `makeContract` function
-must return a record with an `invite` property. This should be a Zoe
+* `zcf` is the <router-link
+  to="/zoe/api/zoe-contract-facet.html">Zoe Contract
+  Facet</router-link>.
+* `invite` must be a Zoe
 invite that will be provided to the user who instantiates the
-contract. `publicAPI` is an optional object whose methods will be
+contract. 
+* `publicAPI` is an optional object whose methods will be
 available to anyone who knows the `instanceHandle` of the contract
 instance. `publicAPI` is a good place to put public queries (i.e.
 `getCurrentPrice` of Autoswap) and other
 requests that shouldn't require making an offer or having an invite.
+
+## Creating an invite
 
 To create an invite in the contract, you can use the Zoe Contract
 Facet method directly (<router-link
--- a/main/zoe/guide/contract-requirements.md
+++ b/main/zoe/guide/contract-requirements.md
@@ -1,0 +1,59 @@
+# Contract Requirements
+
+<Zoe-Version/>
+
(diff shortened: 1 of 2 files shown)


main/zoe/guide/contract-requirements.md
```diff
@@ -0,0 +1,59 @@
...
+* [@agoric/zoe](https://www.npmjs.com/package/@agoric/zoe): Zoe has
+  helpers that contracts can use by importing
+  `@agoric/zoe/src/contractSupport/zoeHelpers`
+
+A Zoe contract will be bundled up, so you should feel free to divide
+your contract into multiple files (perhaps putting helper functions in a
+separate file, for example) and import them.
+
+A Zoe contract needs to be able to run under [Agoric's SES](https://github.com/Agoric/ses-shim#secure-ecmascript-shim-ses-shim). Some legacy
+JavaScript code is incompatible with SES, because SES freezes the
```
tyg, line 28: I'm not clear on what I should do with the info about the SES incompatibility. Is there a workaround or patch? The advice here seems to be "don't mutate JavaScript primordial objects" but what should I do instead?

main/zoe/guide/contract-requirements.md
```diff
@@ -0,0 +1,59 @@
...
+`zcf` is the <router-link
+to="/zoe/api/zoe-contract-facet.html">Zoe Contract Facet</router-link>. The `makeContract` function
+must return a record with an `invite` property. This should be a Zoe
+invite that will be provided to the user who instantiates the
+contract. `publicAPI` is an optional object whose methods will be
+available to anyone who knows the `instanceHandle` of the contract
+instance. `publicAPI` is a good place to put public queries (i.e.
+`getCurrentPrice` of Autoswap) and other
+requests that shouldn't require making an offer or having an invite.
+
```
tyg, line 53: I suggest breaking up the above into a bulleted list; you're slamming together a lot of separate things/definitions in one paragraph.

main/zoe/guide/contract-requirements.md
katelynsills: That's pretty much our answer: don't mutate them because it's bad programming practice. There's not a workaround or patch or anything like that, purposefully.

PR #158, opened 2020-04-21
Update before-using-agoric.md

tyg:


main/getting-started/before-using-agoric.md
```diff
@@ -39,6 +39,12 @@ software, you need to install the following.
   </tr>
   <tr>
     <td><center>4</center></td>
+    <td>If you already have a <code>~/agoric-sdk</code> directory, remove or rename it.</td>
```
katelynsills, line 42: Should this be what we recommend?  If they already have an agoric-sdk directory, they shouldn't do a git clone, but should pull down the latest from the hackathon branch.

tyg: Adding Michael and Chip to review since 1) Chip was the one who brought up already having an agoric-sdk and 2) Michael was general SME for setting up a project. Y'all's thoughts?

FUDCo: We Agoric folk are probably edge cases here since we're actually working on `agoric-sdk` as our main thing separate from the hackathon, and so comingling them might be problematic.  But for most folks, I think Kate's suggestion of just pulling from the other branch makes a lot of sense.

PR #160, opened 2020-04-21
First draft about distributed programming with Agoric

### LINKED ISSUE #130: There is currently not enough doc about the programming model
Proposed solution:
- Add a "JavaScript distributed programming" section in the navigation with the following submenus:
    - `E`
    - vats
    - SES
    - harden

Each should at least link to the most relevant source of information about each subtopic

Also add entries in the glossary

DavidBruant:
The text is really not perfect
but given the timeframe, i would recommand merging if this is considered better than the current nothing
I'm asking for a review specifically to have a different look and be sure it's actually better than nothing (make sure nothing is false or misleading) 

Closes #130

### DIFF
--- a/main/distributed-programming.md
+++ b/main/distributed-programming.md
@@ -42,7 +42,7 @@ lockdown(); // turns the current realm into an immutable realm
 
 JavaScript code runs in the context of a [`Realm`](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-code-realms). A realm is composed of the set of *primordials* (objects and functions of the standard library like `Array.prototype.push`) and of a global object. In a web browser, an iframe is a realm. In Node.js, a Node process is a realm.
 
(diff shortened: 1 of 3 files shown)


main/distributed-programming.md
```diff
@@ -0,0 +1,88 @@
...
+lockdown(); // turns the current realm into an immutable realm
+// calling this function also adds a global Compartment constructor to create compartments
+```
+
+
+### Compartments and realms
+
+JavaScript code runs in the context of a [`Realm`](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-code-realms). A realm is composed of the set of *primordials* (objects and functions of the standard library like `Array.prototype.push`) and of a global object. In a web browser, an iframe is a realm. In Node.js, a Node process is a realm.
+
+For historical reasons, the ECMAScript specification requires the *primordials* to be mutable (`Array.prototype.push = yourFunction` is valid ECMAScript but not recommanded). SES provides to turn the current realm into an **immutable realm**, that is a realm within which, the primordials are deeply frozen. It also allows to create **Compartments**.
```
katelynsills, line 45:
```suggestion
For historical reasons, the ECMAScript specification requires the *primordials* to be mutable (`Array.prototype.push = yourFunction` is valid ECMAScript but not recommended). SES provides to turn the current realm into an **immutable realm**, that is a realm within which, the primordials are deeply frozen. It also allows to create **Compartments**.
```

PR #166, opened 2020-04-22
Docs: Improved description of our programming environment

Chris-Hibbert:
better descriptions of SES and programming with E.

### DIFF
--- a/main/distributed-programming.md
+++ b/main/distributed-programming.md
@@ -9,7 +9,10 @@ Some specificities are **concepts**, other are at the **syntax level**, others a
 
 ## Vats
 
-A vat is a *unit of synchrony*. This means that within a JavaScript vat, objects and functions can communicate with one another synchronously.
+A vat is a *unit of synchrony*. This means that within a JavaScript vat, objects
+and functions can communicate with one another synchronously. It also means that
+communication with objects outside the vat can be done, but it requires
+[managing asynchronous messages and responses](#Communicating-with-remote-objects-via-E).
 
 A vat runs a single *event loop*.
 
@@ -19,14 +22,24 @@ Different vats can communicate by sending asynchronous messages to other vats.
 
 A vat is the moral equivalent of a Unix Process.
 
-The internal state of a vat can be stored in a persistent memory so that the vat can be turned off and later turned back on (on the same or a different physical machine) by loading the stored state.
+The internal state of a vat can be stored in persistent memory so that the vat
+can be turned off and later turned back on (on the same or a different physical
+machine) by loading the stored state.
 
-In cosmic-swingset, several vats are started. Each vat hosts a service (registrar, zoe, etc.). Currently (April 2020), all contracts run in the Zoe vat, but eventually, each contract will run in its own dedicated vat.
+In cosmic-swingset, several vats are started. Each vat hosts a service (registrar, zoe, etc.). Currently (April, 2020), all contracts run in the Zoe vat, but eventually, each contract will run in its own dedicated vat.
 
 
 ## Secure EcmaScript (SES)
 
-SES is a variant of [EcmaScript](https://www.ecma-international.org/publications/standards/Ecma-262.htm) (the standard behind JavaScript - common to web browsers JavaScript and Node.js) that is secure in the sense of object capability security. This means that codes that runs in an immutable realm has no access to ambiant authority by default (no access to disk or network, etc.) and new authority can only be granted by the entity that loaded this code.
+SES is a standards-track extension of
+[EcmaScript](https://www.ecma-international.org/publications/standards/Ecma-262.htm),
+the JavaScript standard - common to web browsers, Node.js, and embedded
+devices. SES provides a secure platform for executing programs, which means that
+it's possible to run code you don't completely trust, without being vulnerable
+to its bugs or bad intentions. The hosted code runs in an immutable realm has no
+access to ambient authority by default. The hosting code can provide access to
+disk or network or other services, and any authority not granted is inaccessible
+to the hosted code.
 
 SES can be enabled via the [`ses` package](https://www.npmjs.com/package/@agoric/harden).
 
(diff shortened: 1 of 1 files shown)


main/distributed-programming.md
```diff
@@ -42,22 +55,35 @@ lockdown(); // turns the current realm into an immutable realm
...
-In a nutshell, `harden` creates objects that can be trusted. In practice, `harden` performs the deep version of [`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) (which only freezes the object's own properties).
-
-Most methods of the Agoric platform require their arguments to be `harden`ed. This is the only way they can provide the security properties they claim to provide.
+In a nutshell, `harden` creates objects that can't be changed. In practice,
+`harden` performs the deep version of
+[`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze). Freeze
+only freezes the object's own properties, so harden is necessary in order to
+ensure that no further changes can be made. It is useful in distributed systems
+to distinguish objects with mutable state from objects that are
+unchanging. Objects that can change over time are extremely useful, but in order
```
katelynsills, line 79: The rest of this paragraph was a little confusing to me because of the vagueness.

main/distributed-programming.md
```diff
@@ -42,22 +55,35 @@ lockdown(); // turns the current realm into an immutable realm
...
+to interact with them, you have to send the message to where they are. If an
+object is guaranteed not to change, then copies of it can be widely shared, and
+interacted with directly.
+
+Many services provided by Agoric require their parameters to be hardened so they
```
katelynsills, line 84: I think we're conflating hardening with something being a record. Even a presence can be hardened, but that doesn't turn it into a record that is passed by copy. The distinction is having functions or not, as I understand it.

Chris-Hibbert: I don't think a user would call `harden()` on a presence. 

Perhaps you mean that presences are also hardened? I don't know whether that's true, but it doesn't seem related to what kinds of objects a user would call harden on and why.

What I'm trying to explain here is which kinds of objects a user would want to call harden on, and why.  Are there other reasons to call harden()?

main/distributed-programming.md
Chris-Hibbert: In a lot more explicit detail, what I was trying to say was this:

Why is a deep freeze operation useful?

* Objects with mutable state don't migrate. 
* If you want to interact with such an object, you have to send data to it. 
* The data you're sending can't itself consist of mutable objects.
* You can send primitive values directly.
* If we want mobile data to have more structure, we can make records, but those must be transitively frozen in order to be passed around.

Does that make more sense? Is it worth saying?

main/distributed-programming.md
katelynsills: We do call harden when making objects that will be represented as presences all the time though. For instance, the Zoe service is hardened. I think we're misleading readers by associating harden at all with passing a record. 

It's a good question why we want to do that. @erights why do we do that?

erights: `harden` is not about immutability. It is about tamper-proofing an API surface. Our counter-example:

```js
function makeCounter() {
  let count = 0;
  return harden({
    incr() { return ++count; },
    decr() { return --count; },
  });
}
```

Returns a hardened counter object. Its whole purpose is to be mutated. Hardening the object is only about ensuring that the behavior of the object is according to the code of the object. If Alice and Bob both share the hardened counter object, they can communicate by incrementing, decrementing, and observing the count, but nothing else.

Without the harden, Alice could replace the counter's incr method with something else, causing Bob to see behavior that is not according to the code of the contract. This way madness lies.

Our general style is not to represent mutable state by modifying properties. Rather, as above, by mutating captured lexical variables. `harden` only locks down properties, nothing else.

## When it is about immutability

For objects that are deeply pass-by-copy according to marshal: they contain no functions other than inherited pure primordial methods. Without functions, they have no where else to store mutable state. Locking down all their properties, as `harden` does, makes them immutable. This case is important, but from `harden`'s pov, it's a special case.

Chris-Hibbert: I can see the point of thinking about `harden()` and describing it in those terms when discussing objects that are shared within a vat, but I get confused when thinking about it in the context of writing a contract. Contract writers share objects as presences, so the remote users of those objects don't have any ability to modify them other than by their API. 

When we're trying to explain things to developers who are writing contracts in the Zoe framework, what role does this aspect of `harden()` have in what they need to understand?  I can see that it's a useful concept as far as their defensive programming if they define multiple objects over multiple source files, and want to convince themselves and reviewers of their code that they're following best practices, but it doesn't seem to add anything to the story about the contracts they're publishing.

erights: I think the high order role of `harden` is simply that only hardened objects are passable. Anything used as receiver, argument, or return value of a remote message must be hardened.

Chris-Hibbert: I simplified the description of harden. Please take another look.

main/distributed-programming.md
Chris-Hibbert: I dropped most of this.

main/distributed-programming.md
```diff
@@ -68,22 +68,17 @@ Agoric "deploy scripts" and smart contract code run in an immutable realm with c
...
-service without waiting for further round trips to collect information.
+`harden` is a function provided by the
+[`@agoric/harden` package](https://www.npmjs.com/package/@agoric/harden).  It ensures
+that external callers of hardened objects can only interact with them through
+functions present in their API.  `harden` does a transitive version of
+[`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze),
+which only locks up an object's own properties.
+
+CapTP, our comms layer for passing references to distributed objects, requires that
+all objects that will be transferred to other trust domains must have their API
```
erights, line 79: "trust domains" is not accurate. There's a trust boundary between the zoe contract facet and the contract in the same vat, but there is no captp there, and nothing gets marshaled across that boundary. Nonetheless, because it is a security boundary, each side *should* harden the objects they expose to the other.

As for the captp/marshall-imposed requirement, that's only at vat boundaries.

main/distributed-programming.md
```diff
@@ -36,7 +36,7 @@ SES is a standards-track extension of
 the JavaScript standard - common to web browsers, Node.js, and embedded
 devices. SES provides a secure platform for executing programs, which means that
 it's possible to run code you don't completely trust, without being vulnerable
-to its bugs or bad intentions. The hosted code runs in an immutable realm has no
+to its bugs or bad intentions. The hosted code runs in an immutable realm that has no
 access to ambient authority by default. The hosting code can provide access to
 disk or network or other services, and any authority not granted is inaccessible
 to the hosted code.
```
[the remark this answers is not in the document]
katelynsills: How about "third-party code" instead of hosted code?

main/distributed-programming.md
```diff
@@ -73,15 +94,51 @@ harden(o);
 o.a  = 37; // throws a TypeError because o is now hardened
 ```
 
-### Communicating with remote objects via `E`
+### Communicating with remote objects using `E`
+
+
+On the Agoric platform, object may be running in distinct vats, or on a remote
```
katelynsills, line 100:
```suggestion
On the Agoric platform, objects may be running in distinct vats, or on a remote
```

main/distributed-programming.md
```diff
@@ -73,15 +94,51 @@ harden(o);
...
-However, the deploy script is given access to a `zoe` *presence*. In order to call methods on the actual Zoe object, the deploy code can do:
 ```js
 const installationHandle = await E(zoe).install(source, moduleFormat);
 ```
 
-The `E` function is a local "bridge" that allows to talk to remote object and call their methods. The local version of a remote object is called a **presence**. `E` takes a presence as an argument all created an object with the corresponding methods. It performs the communication asynchronously under the hood. Method calls can take arguments that are objects in the current vat or presences to the zoe vat or presences of objects from other vats.
-
+The `E` function is a local "bridge" that makes is possible to invoke methods on
+remote object. The local version of a remote object is called a
+**presence**. `E` takes a presence as an argument all creates an object with the
```
katelynsills, line 140:
```suggestion
**presence**. `E` takes a presence as an argument and creates an object with the
```

main/distributed-programming.md
Chris-Hibbert: I changed it to say 
```
All objects that will be transferred across a trust boundary must have their API surface frozen
```
and mention at the end that CapTP enforces the rule across vat boundaries.

main/distributed-programming.md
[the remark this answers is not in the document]
Chris-Hibbert: done

PR #187, opened 2020-05-02
Zoe point 6 patch

tyg:


main/zoe/api/zoe-contract-facet.md
```diff
@@ -71,6 +77,8 @@ invoked in the contract with a handle for the offer. The result of the
 `offerHook` will be returned as the "outcome" of making the offer via 
 the invitation.
 
+The `inviteDesc` is a string **tyg todo** need to find definition.
+
```
Chris-Hibbert, line 81: The only restriction is that it must be a string. I think the relevant thing to say about it is that it will be visible in the Invite's amount, which can be useful for distinguishing invites and keeping track of them.

tyg: Thanks.

PR #198, opened 2020-05-16
Wallet api doc added and put under Guides in menubar

tyg:


main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
+# Wallet API
+
+This document lists and describes the API commands available to the wallet.
+
+## Petnames
```
dtribble, line 5: Move this after the overview section.

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+Grandpa, Kate S., etc.
+
+## Overview
+
+Wallet API commands work with the following object types:
+- `purse`: Stores assets until you withdraw them into a payment for use 
+- `issuer`: Creates empty purses and payments and maps minted assets
+  to them when assets are added or removed. Issuers verify and move
+  digital assets.
+- `offer`: Consists of what amount of what brand you're willing to
```
dtribble, line 23: uses the term "brand" here, so need to define that before it. Lift or point at ERTP definitions?  or glossary?

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+  give, and what amount of what brand you want, as well as the
+  conditions under which the offer holder can cancel it.
+
+## Purse API Methods
+
+### `deposit()`
+- **Function:**
```
dtribble, line 30: this is different in format from the rest of our API docs. Why?

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+  - Deposits a payment into the specified purse in the wallet. Returns the deposited amount.
+- **Required Arguments:**
+  - `pursePetname`
+    - **String**. The petname of the purse in which to deposit the specified payment.
+  - `payment`
+    - **Object**. A payment object, which is deposited into the specified purse.
+- **Optional Arguments:**
```
dtribble, line 37: Delete these sections. Just use "optional" in the description for such arguments e.g., [burn](https://agoric.com/documentation/ertp/api/issuer.html#issuer-burn-payment-optamount).

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+  - None.
+
+### `getPurses()`
+- **Function:**
+  - Returns all purse objects and their petnames in the wallet. They are returned as an array of [petname, purseObject] arrays which can be turned into a `Map` (`new Map(getPurses())`)
```
dtribble, line 42: it's actually remove references for purses. I don't know that we need to say that here. Check out other examples?

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+- `issuer`: Creates empty purses and payments and maps minted assets
+  to them when assets are added or removed. Issuers verify and move
+  digital assets.
+- `offer`: Consists of what amount of what brand you're willing to
+  give, and what amount of what brand you want, as well as the
+  conditions under which the offer holder can cancel it.
+
+## Purse API Methods
+
+### `deposit()`
```
dtribble, line 29: This should list the arguments (see example ERTP methods)

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+- **Function:**
+  - Returns all purse objects and their petnames in the wallet. They are returned as an array of [petname, purseObject] arrays which can be turned into a `Map` (`new Map(getPurses())`)
+- **Required Arguments:**
+  - None.
+- **Optional Arguments:**
+  - None.
+
+### `getPurse(pursePetname)`
+- **Function:**
+  - Returns the `purse` object with the given petname
```
dtribble, line 50: Rather than "with", it's really that one assigned to that pet name.  It's more that the petname has a purse, rather than the purse having a petname. I'm a little unsure about how we want this usage, so if wallet comments speak to it, use that slant.

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+    - **String**. The petname assigned to the new purse.	
+- **Optional Arguments:**
+ - None.
+- **Errors:**
+  - There is already a purse in this wallet with the name of `pursePetname`'s value.   
+
+
+## Issuer API Methods
+
+### `addIssuer()`
```
dtribble, line 86: missing arguments in template

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+  - `issuer`
+    - An issuer object
+- **Optional Arguments:**
+  - None.
+
+## Offer API Methods
+
+### `acceptOffer()`
+- **Function:**
+  - Changes the specified offer's status to "accepted".
```
dtribble, line 119: This actually makes the offer to the target contract. It approves the "proposal" by submitting the real offer on behaf of the user.

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+- **Required Arguments:**
+  - `id`
+    - **String**: An offer id.
+- **Optional Arguments:**
+  - None.
+- **Errors:**
+  - If the offer has already been resolved or rejected. 
+
+### `addOffer()`
```
dtribble, line 128: This really just adds a proposal that the user may approve to actually make an offer to the contract invitation.

main/wallet-api.md
```diff
@@ -0,0 +1,206 @@
...
+
+`hooks: {
+     publicAPI: {
+        getInvite: ['makeInvite'], // E(publicAPI).makeInvite()
+     },
+}`
+  
+### `cancelOffer()`
+- **Function:**
+  - Changes the status of the specified offer in the wallet inbox to
```
dtribble, line 153: The "changes the status" is not the important thing. It's "transition the state" by e.g., submitting or canceling the offer in the contract.

main/wallet-api.md
katelynsills: I don't understand what you're saying here, dean. remove references?

main/wallet-api.md
katelynsills: I think this usage makes sense, since it will always be a bidirectional one-to-one mapping.

main/wallet-api.md
tyg: Because I didn't like the other format with respect to it not clearly denoting which parameters were optional, no slot for errors, etc. When there's time, I want to write up a really good API template and convert everything to it.

main/wallet-api.md
tyg: I think it's clearer for optional and required arguments to be separate.

dtribble: That's not typical for JS apis; there's too many variants (e.g., required objects with optional subfields, arguments that default, etc.)

main/wallet-api.md
dtribble: bah. no idea how I typed those words.  "it actually *returns promises* for purses rather than direct references...".  I was thinking about the purse deletion and my brain swapped words around.

PR #216, opened 2020-06-16
MathHelpers fix

tyg:
Revised ERTP Guide section on MathHelpers and intro to ERTP API MathHelpers doc to explain in both places about the three types of MathHelpers, what and why to use them, and how to specify in code which type to use.

### DIFF
--- a/main/ertp/api/math-helpers.md
+++ b/main/ertp/api/math-helpers.md
@@ -1,14 +1,31 @@
-# Math Helpers
-All of the differences in how digital asset amount are manipulated can
-be reduced to the behavior of the math on extents. We extract this
-custom logic into mathHelpers. MathHelpers are about extent
-arithmetic, whereas AmountMath is about amounts, which are the extents
-labeled with a brand. AmountMath use mathHelpers to do their extent
-arithmetic, and then brand the results, making a new amount.
+# MathHelpers
+AmountMath is about doing math operations on `amount`s, which are extents labeled with a brand.
+AmountMath uses MathHelpers to do its extent arithmatic operations. It then brands the results,
+creating a new `amount`. 
+
+There are three different types of MathHelpers, each of which implements all the methods shown 
+on this page. You only have to specify the MathHelper type when creating an `issuer`; it then knows
+which type's operations to use on itself. 
+
+There are three types of MathHelpers, each of which implements all of the same 
+set of API methods (i.e. MathHelpers are polymorphic):
+- `nat`: Used with fungible assests.
+- `strSet`: Used with non-fungible assets.
+- `set`: Used with sets of objects, primarily non-fungible assets.
+
+Use `produceIssuer(allegedName, mathHelpersName)` to specify which type of MathHelpers
+your contract uses. The second parameter, `mathHelpersName` is optional and defaults 
+to `nat` if not given. For example
+```js
+produceIssuer('quatloos`); // Defaults to 'nat'
+produceIssuer('quatloos', 'strSet');
+produceIssuer('quatloos, 'set');
+```
+For more details on the MathHelper types, see the [ERTP Guide's MathHelpers section](https://agoric.com/documentation/ertp/api/math-helpers.html)
 
 ::: warning MathHelpers versus AmountMath
-These MathHelpers should not be used on their own. Instead, a local
-version of AmountMath should be made (see below).
+MathHelper operations should not be used on their own. Instead, you 
+should first make a local version of AmountMath as shown below.
 :::
 
 
--- a/main/ertp/guide/math-helpers.md
+++ b/main/ertp/guide/math-helpers.md
@@ -20,17 +20,18 @@ See the [MathHelpers API documentation](https://agoric.com/documentation/ertp/ap
 
 ## MathHelpers Choices
 
-### 'nat'
-The default fungible asset case. For example, an amount you'd use `nat` MathHelpers with might look like:
+### nat
+The default fungible asset case, which works with natural number extents. 
+For example, an amount you'd use `nat` MathHelpers with might look like:
 
 ```js
 { brand: myBrand, extent: 40 }
 ```
 
-### 'strSet'
-High performance operations on sets of string IDs for a basic nonfungible asset
-case. An amount you'd use `strSet` with might look like the following. Note that
-the `extent` is not a numeric value, but an array specifying a specific `myBrand` object.
+### strSet
+Used for MathHelper operations on sets of string IDs for a basic nonfungible asset
+case. Note that the `extent` is not a numeric value, but an array specifying a 
+specific `myBrand` appropriate object. An amount you'd use `strSet` with might look like the following.
 
 ```js
 { brand: myBrand, extent: ['39283', 'bas029s'] }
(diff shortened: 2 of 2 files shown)


main/ertp/guide/math-helpers.md
```diff
@@ -1,19 +1,36 @@
 # MathHelpers
-MathHelpers are the "arithmetic" operations on extents, used for actions like
-addition, subtraction, and greater than or equal to. There are currently three choices for mathHelpers.
+AmountMath uses MathHelpers to do their extent arithmetic. The results are branded, making
+a new amount (which consists of an extent and a brand). MathHelper operations include addition,
+subtraction, equality checking, greater than testing, and working with empty extents.
```
Chris-Hibbert, line 4: I'd say `comparison` rather than `greater than`, simply because the operation is actually greater than or equal, and you don't want objections from pedantics like me.

main/ertp/guide/math-helpers.md
```diff
@@ -31,6 +48,5 @@ information. An amount might look like:
 { brand: myBrand, extent: [{ seat: '16F', flight: '39DFK', date: '2020-06-08'}] }
 ```
 
-While the `set` mathHelpers are slower than the `strSet`, they are
-more trustworthy because all of the necessary information is in the
-extent.
+While the `set` mathHelpers are slower than the `strSet` equivalents, they are
```
Chris-Hibbert, line 51: I don't understand the purpose of this statement. Why is it helpful to say that there's a performance difference? What do we mean by saying "all the necessary information is in the extent" of one but not the other?

main/ertp/api/math-helpers.md
```diff
@@ -1,14 +1,31 @@
...
+produceIssuer('quatloos', 'strSet');
+produceIssuer('quatloos, 'set');
+```
+For more details on the MathHelper types, see the [ERTP Guide's MathHelpers section](https://agoric.com/documentation/ertp/api/math-helpers.html)
 
 ::: warning MathHelpers versus AmountMath
-These MathHelpers should not be used on their own. Instead, a local
-version of AmountMath should be made (see below).
+MathHelper operations should not be used on their own. Instead, you 
+should first make a local version of AmountMath as shown below.
```
Chris-Hibbert, line 28: I think this is conflating two different points.

 * it's better to use AmountMath to group the operations together because it gives you more type coherence and is easier to manage, and
 * Issuers will happily give you an AmountMath object to use, but it will normally be remote. You get better performance by making a local version.

main/ertp/api/math-helpers.md
```diff
@@ -1,14 +1,31 @@
...
+
+There are three different types of MathHelpers, each of which implements all the methods shown 
+on this page. You only have to specify the MathHelper type when creating an `issuer`; it then knows
+which type's operations to use on itself. 
+
+There are three types of MathHelpers, each of which implements all of the same 
+set of API methods (i.e. MathHelpers are polymorphic):
+- `nat`: Used with fungible assests.
+- `strSet`: Used with non-fungible assets.
+- `set`: Used with sets of objects, primarily non-fungible assets.
```
katelynsills, line 14: I think this isn't explaining the difference between `strSet` and `set` correctly. Both are for nonfungible assets and implement set logic (or to be pedantic, logic similar to set logic). The only difference is that `strSet` has elements that are only strings, but in `set` the elements are records with keys and values. Maybe 
```suggestion
- `strSet`: Used with non-fungible assets, operates on an array of string identifiers
- `set`: Used with non-fungible assets, operates on an array of records (objects) with keys and values
```

main/ertp/api/math-helpers.md
```diff
@@ -1,14 +1,31 @@
...
+
+Use `produceIssuer(allegedName, mathHelpersName)` to specify which type of MathHelpers
+your contract uses. The second parameter, `mathHelpersName` is optional and defaults 
+to `nat` if not given. For example
+```js
+produceIssuer('quatloos`); // Defaults to 'nat'
```
katelynsills, line 20:
```suggestion
produceIssuer('quatloos'); // Defaults to 'nat'
```

main/ertp/api/math-helpers.md
katelynsills: Yeah, I think we should always recommend making a local AmountMath, though, and should never suggest using the remote AmountMath. I like the explanation as it is, I think it's short and more focused on what you *should* do than an academic explanation

main/ertp/guide/math-helpers.md
katelynsills: That sounds good to me.

main/ertp/guide/math-helpers.md
```diff
@@ -1,19 +1,36 @@
...
-addition, subtraction, and greater than or equal to. There are currently three choices for mathHelpers.
+AmountMath uses MathHelpers to do their extent arithmetic. The results are branded, making
+a new amount (which consists of an extent and a brand). MathHelper operations include addition,
+subtraction, equality checking, greater than testing, and working with empty extents.
+
+The MathHelper operations are polymorphic. There are three types of MathHelpers, each of which 
+implements all of the same set of API methods:
+- `nat`: Used with fungible assests.
+- `strSet`: Used with non-fungible assets.
+- `set`: Used with sets of objects, primarily non-fungible assets **tyg todo: Is the non-fungible part correct?**
```
katelynsills, line 10: Can we use the descriptions I put in another comment in api/math-helpers.md?

main/ertp/guide/math-helpers.md
```diff
@@ -1,19 +1,36 @@
...
+
+Use `produceIssuer(allegedName, mathHelpersName)` to specify which type of MathHelpers your contract uses. The second
+parameter, `mathHelpersName` is optional and defaults to `nat` if not given. For example
+```
+produceIssuer('quatloos`); // Defaults to 'nat'
```
katelynsills, line 15:
```suggestion
produceIssuer('quatloos'); // Defaults to 'nat'
```

main/ertp/guide/math-helpers.md
```diff
@@ -1,19 +1,36 @@
...
+produceIssuer('quatloos', 'strSet');
+produceIssuer('quatloos, 'set');
+```
+See the [MathHelpers API documentation](https://agoric.com/documentation/ertp/api/math-helpers.html) for the set of operations all the above types implement.
 
-## Choices of MathHelpers
+## MathHelpers Types
```
katelynsills, line 21: I think I prefer to call these Choices rather than Types.

main/ertp/guide/math-helpers.md
katelynsills: We can probably back off of claims about performance, because I implemented a performance enhancement for `set` so it may not actually be much worse than `strSet` performance. 

> What do we mean by saying "all the necessary information is in the extent" of one but not the other?

By this, I meant that in `strSet` extents, often you only get the ID (the VIN for cars, to use an analogy), and other important information is stored elsewhere (the car make, model, color, etc.) In `set`, you often put all the information that is important for a third party to know in the extent. Information that is in an immutable extent is more credible than information that is retrieved from an external source that could potentially manipulate the information without anyone realized.

I think we want to keep these kinds of considerations in this document because we want to be moving towards documentation that is less factual and more helpful.

Chris-Hibbert: What is it about `str` vs. `strSet` that would lead people to use them in these distinct ways? I could easily imagine other developers using them in ways that don't follow this pattern. 

How about "amounts are more trustworthy when all the necessary information is in the extent"

PR #222, opened 2020-07-02
Zoe07 contracts

tyg:


main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -2,130 +2,80 @@
 
 <Zoe-Version/>
```
katelynsills, line 3: Can we update the Zoe version? It should be updated for 0.7.0. I can show you where it gets updated

main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -2,130 +2,80 @@
...
-behave opportunistically.
-
-In the `atomicSwap` contract, anyone can securely swap with a counterparty by escrowing digital assets with Zoe and sending an invite to the swap to the counterparty.
-
-Let's say that Alice wants to swap with Bob as the counterparty. She
-knows that the code for the swap has already been installed, so she
-can create a swap instance from the swap installation (`handle` is the
-unique, unforgeable identifier):
-
+##### [View all example contracts on Github](https://github.com/Agoric/agoric-sdk/tree/master/packages/zoe/src/contracts)
```
katelynsills, line 5: Thanks for changing this

main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -2,130 +2,80 @@
...
+
+To solve this problem, swap contracts let users securely trade one kind of 
+digital asset for another kind. In the `atomicSwap` contract, anyone can securely 
+swap with a counterparty by escrowing digital assets with Zoe and sending an invite 
+to the swap to the counterparty.
+
+Alice's initial offer is `{ give: { Asset: A }, want: { Price: B } }`.
+Its outcome is an invitation for the second party, Bob, who should offer
+`{ give: { Price: B }, want: { Asset: A } }`, with a `want`
+amount no greater than Alice's `give`, and a `give` amount at least as
```
katelynsills, line 20: The "greater than" language was hard to parse. Maybe: For the trade to succeed, Bob shouldn't want any more than what Alice gives, and Bob should give enough to cover what Alice says she wants.

main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -2,130 +2,80 @@
...
+large as Alice's 's `want`.
+
+First, let's get the needed ZoeHelper methods, and specify and harden the
```
katelynsills, line 23: This seems like a major shift in the intention of the document. The old doc explained how to use the contract, but this new doc is explaining how to write it. I think I like the new direction, but maybe we need to preface this paragraph with something like: "Let's explain how this contract works" to let the user know what we're describing

main/zoe/guide/contracts/atomic-swap.md
```diff
@@ -2,130 +2,80 @@
...
+  const firstOfferExpected = harden({
+    give: { Asset: null },
+    want: { Price: null },
+  });
 ```
-
-Bob decides to be the counter-party. He also escrows his payments and uses his invite to
-make an offer in the same way as Alice, but his `Proposal` match Alice's:
-
+And finally, we return our invitation to send to Bob. **tyg todo: Not clear on why there are two makeInvitation calls?**
```
katelynsills, line 74: There are two makeInvitation calls because Alice needs an invite and Bob needs a different invite.

main/zoe/guide/contracts/atomic-swap.md
katelynsills: Actually, let's check in about this - I'm not sure the rest of the doc is going where it should in terms of what the user needs to know.

main/zoe/guide/contracts/barter-exchange.md
```diff
@@ -1,17 +1,144 @@
 <Zoe-Version/>
 
-##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/958a2c0a3dec38bdba2234934119ea2c28958262/packages/zoe/src/contracts/barter-exchange.js) (Last updated: 6/30/2020)
+##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/master/packages/zoe/src/contracts/barterExchange.js (Last updated: 6/30/2020)
```
katelynsills, line 3: So in the atomicSwap doc, it was a link to all contracts. I like linking to the individual contract, but it seems like we need to be consistent about that

main/zoe/guide/contracts/barter-exchange.md
```diff
@@ -1,17 +1,144 @@
...
 
-Barter Exchange takes offers for trading arbitrary goods for one another.
+Barter Exchange takes offers for trading arbitrary goods for other things
+It only accepts offers that look like:
+- `{ give: { In: amount }, want: { Out: amount}` }
 
-This Barter Exchange accepts offers to trade arbitrary goods for other
-things. It doesn't require registration of Issuers. If two offers satisfy
-each other, it exchanges the specified amounts in each side's want clause.
+You don't have to register Issuers. If two offers satisfy each other, it exchanges the  specified amounts in each side's want clause. The want amount is matched, while the give amount is a maximum. Each successful trader gets their `want` and may trade with counter-parties who specify any amount up to their specified `give`.
```
katelynsills, line 10: I realize this part is coming from the barterExchange.js comments, but you do have to register issuers, right?

main/zoe/guide/contracts/simple-exchange.md
```diff
@@ -2,19 +2,208 @@
...
+
+The `Asset` is an exact amount to be exchanged, while the
+`Price` is a limit that may be improved on. SimpleExchange does
+not partially fill orders.
+
+The invitation returned on contract installation is the same as what
+is returned by calling `publicAPI.makeInvite()`.
+
+## Setting things up
+
```
Chris-Hibbert, line 30: The above is a good description of the external interface. I would make the description of the internals much shorter. The high points are (1) it keeps two lists of offerHandles for buys and sells, (2) each new offer is compared to all the existing offers looking for a match with `satisfies()` and `satisfiedBy()`, (3) if a match is found `swap(offerHandle, iHandle)` is called, (4) otherwise the new offer is added to the appropriate list, and (5) the contract uses `zcf.initPublicAPI()` to provide three functions that anyone can call to get the status of an offer, to get a notifier to see the book, or to make a new invitation.

PR #238, opened 2020-07-22
Fixed error in coerce() description/argument (removed 'value)

tyg:


main/ertp/api/amount-math.md
```diff
@@ -53,11 +53,11 @@ const { amountMath } = makeIssuerKit('bucks');
...
 ```
 
-## amountMath.coerce(allegedAmountOrValue)
-- `allegedAmountOrValue` `{Amount}`
+## amountMath.coerce(allegedAmount)
+- `allegedAmount` `{Amount}`
 - Returns: `{Amount}`
 
-Make sure this amount (or value) is valid and return it as an amount if so.
+Make sure this amount is valid and return it as an amount if so.
```
Chris-Hibbert, line 61: Throw an exception if not.

katelynsills: Yup, good point

PR #240, opened 2020-07-24
Patched non-specific handle parameter names.

tyg:


main/zoe/api/records.md
```diff
@@ -144,7 +144,7 @@ The possible records are:
 ## Installation Record
 ```js
 {
-  handle, // opaque identifier, used as the table key
+  installationHandle, // opaque identifier, used as the table key
   installation, // contract code
 }
```
Chris-Hibbert, line 149: The second field is now `bundle`.  We should say  something about it containing the source code and other info. The source code is accessible as `bundle.source`.

PR #249, opened 2020-08-24
ERTP API Alpha Release update and clean up. 

tyg:
This PR covers both updating the ERTP API docs to Alpha Candidate and in general cleaning them up and enhancing them. It has all files under the ../ertp/api.  Related PRs are for updating the ERTP Introduction in Getting Started and updating the Zoe API doc files.


main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,279 @@
...
 - Returns: `{boolean}`
 
-Returns true if the leftAmount is greater than or equal to the rightAmount. For non-scalars, "greater than or equal to" depends on the kind of amount, as defined by the MathHelpers. For example, whether rectangle A is greater than rectangle B depends on whether rectangle A includes rectangle B as defined by the logic in MathHelpers.
+Returns `true` if the `value` of `leftAmount` is greater than or equal to
+the `value` of `rightAmount`. Both `amount` arguments must have the same
+`brand`.
+
+For non-fungible `values`, what "greater than or equal to" is depends on the 
+kind of `amountMath`. For example, whether rectangle A is greater than rectangle B 
+depends on whether rectangle A includes rectangle B as defined by the logic in `amountMath`
```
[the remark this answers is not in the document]
tyg: How's this?

For example, { 'seat 1', 'seat 2' } is considered
greater than { 'seat 2' } because the former 
both contains all of the latter's  contents and has more elements.

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,279 @@
...
+An `amountMath` has a one-to-one association with a `brand`, established when
+both are created by `makeIssuerKit()`. The association cannot be broken or changed;
+a particular `amountMath` will always and only be used on `amounts` with its
+initially associated `brand`. 
 
 ```js
-const { issuer } = makeIssuerKit('bucks');
+const { issuer } = makeIssuerKit('quatloos');
+//Get the issuer's associated amountMath.
 const exampleAmountMath = issuer.getAmountMath();
```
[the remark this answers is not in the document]
tyg: Deleted issuer.getAmountMath. Added section near top on local Amount Maths, and documented makeLocalAmountMath(issuer) just under it.

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
+
+A `value` must be `Comparable`.
+
+## LocalAmountMath
+
+We encourage you to make and use `LocalAmountMath`, a local and thus synchronous version of `AmountMath`. 
+Their local or remote status is the only different between the two types; each has the same methods, the
+same kinds (`MathKind.NAT`, etc.) and the same one-to-one relationship with a `mint`, `issuer`, and `brand`.
+If necessary, both a `LocalAmountMath` and an `AmountMath` can exist and be associated with the same
+`mint`, `issuer`, and `brand` **tyg todo This last is a guess on my part. Correct?**.
```
Chris-Hibbert, line 48: This is mostly a description for TYG, and not intended as text for this doc.

local AmountMaths aren't a different type from remote ones, they merely reside in the same vat as the user. This allows them to be accessed without remote calls. Both use the same code, and hold the same state. The Brand is a unique object that has a unique instance somewhere in the distributed object system. 

The way we share objects and references across the networks distinguishes things with identity and copyable values. by-copy values include numbers and strings. Every instance of 3 is indistinguishable, so we make copies when we need to, and one can't tell whether any particular '3' is the same as another. OTOH, objects are unique and live in only one place. If an object in one vat has a reference to another object from another vat, we represent that with a local object called a "presence" that can forward messages to the original. *We ensure that there is only one presence for any particular object in each vat* that knows about it. This means if there are two different objects in vat A that both have references to some object in vat B, they will be the same object.

This unique object is what we use for the Brand held by an AmountMath. The rest of the state of the AmountMath is copyable state. When you have a local AmountMath, it's holding onto the presence for the remote Brand, and so can recognize any Amount that is from the same issuer. A remote AmountMath has the same ability to distinguish between Amounts that came from that Issuer and others because as the Amounts are shared around, their values are copied, but the Brands they contain are represented as presences, and so can be uniquely identified in every vat in which they appear. 

The only reason one doesn't want to use a remote AmountMath is that every call takes a network roundtrip. We designed the representation of AmountMath so a local copy is every bit as good as a remote one, and faster to access.

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
 
-Value must be Comparable.
+## makeLocalAmountMath(issuer)
+- `issuer`: `{issuer}`
+Returns: `{ LocalMathAmount }`
```
Chris-Hibbert, line 52: LocalAmountMath isn't a type.

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
+
+Creates and returns a local (synchronous) Amount Math object. The new `localAmountMath` has a one-to-one
+relationship with the `issuer`, and thus also to the `brand` and `mint` in one-to-one
+```js
+import { makeLocalAmountMath } from '@agoric/ertp';
+const quatloosLocalAmountMath = await makeLocalAmountMath(quatloosIssuer);
```
Chris-Hibbert, line 58: users wouldn't name it with "local". We've designed the API so the only kind of AmountMath you can get is a local one. Perhaps we should have left the "local" out of the function's name, but we wanted to call attention to the fact that it has to do some asynchronous operations to get the Brand, and that's why you have to include the "await".

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
 ```
 
 ## amountMath.make(allegedValue)
 
 - `allegedValue` `{Value}`
 - Returns: `{Amount}`
 
-Make an amount from a value by adding the brand.
+Make an `amount` from a `value` by adding the `brand` associated with
+the `amountMath`..
```
Chris-Hibbert, line 95: double period

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
 
 ```js
-const { amountMath } = makeIssuerKit('bucks');
-const amount837 = amountMath.make(837);
+const { amountMath: quatloosAmountMath } = makeIssuerKit('quatloos');
+//amount837 = { value: 837 brand: quatloos }
```
Chris-Hibbert, line 99: I may have led you astray here. There needs to be a comma after '837'.
```suggestion
//amount837 = { value: 837, brand: quatloos }
```

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
-Make sure this amount is valid and return it as an amount if so.
+Make sure this `amount` is valid and if so, return it.
 If not valid, throws an exception. 
 
-```js
-const { amountMath } = makeIssuerKit('bucks');
-const bucks50 = amountMath.make(50);
 
-amountMath.coerce(bucks50); // equal to bucks50
+Make sure this amount is valid and if so, return it.
```
Chris-Hibbert, line 111: duplicate text

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
+
+The example would show
```
Chris-Hibbert, line 113: editorial comment?

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
+Returns `true` if the `value` of `leftAmount` is equal to
+the `value` of `rightAmount`. Both `amount` arguments must have the same
+`brand`.
+
+For non-fungible `values`, "equal to" depends on the kind of `amountMath`. 
+For example, { 'seat 1', 'seat 2' } is considered
+unequal to { 'seat 2' } because the number of items in the former is
+different from that of the latter. Similarly { `seat 1` } and { `seat2` } 
+are considered unequal because the conthe former both contains all of the latter's 
+contents and has more elements.
```
Chris-Hibbert, line 223:
```suggestion
different from that of the latter. Similarly { `seat 1`,  `seat2`  } and { `seat2` } 
are considered unequal because the former has elements that are not contained in the latter.
```

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
 ```js
-const { amountMath } = makeIssuerKit('fungible');
-const empty = amountMath.getEmpty();
-const fungible1 = amountMath.make(1);
-const anotherFungible1 = amountMath.make(1);
+const { amountMath: quatloosAmountMath } = makeIssuerKit('quatloos');
+const empty = quatloosAmountMath.getEmpty();
+const quatloos10 = quatloosAmountMath.make(10);
+const quatloos5 = quatloosAmountMath.make(5);
+const quatloos5-2 = quatloosAmountMath.make(5);
```
Chris-Hibbert, line 230: javascript variable names can't have dashes in them. perhaps `quatloos5b`?

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,304 @@
...
 ```js
-const { amountMath } = makeIssuerKit('myItems', 'strSet');
-const listAmountA = amountMath.make(harden['1','2','4']);
-const listAmountB = amountMath.make(harden['3']);
+import { MathKind, makeIssuerKit } from '@agoric/ertp';
+const { amountMath: itemsAmountMath } = makeIssuerKit('myItems', MathKind.STRING_SET');
+const listAmountA = itemsAmountMath.make(harden['1','2','4']);
+const listAmountB = itemsAmountMath.make(harden['3']);
 
 // Returns ['1', '2', '4', '3']
```
Chris-Hibbert, line 263:
```suggestion
// Returns an amount whose value is ['1', '2', '4', '3']
```

main/ertp/api/mint.md
```diff
@@ -1,31 +1,48 @@
...
-[`Brand`](./brand.md).
+Only a `mint` can issue new digital assets, so only the holder of a `mint` can
+do so. A `mint` has a one-to-one relationship with both an `issuer` and a `brand`,
+and can only mint new assets of that `brand` and is the only `'mint` that can mint
+new `brand` type assets for its associated `brand`.
+
+## makeIssuerKit(allegedName, amountMathKind)
+- `allegedName` `{String}`
+- `amountMathKind` `{MathKind}`
+- Returns: `{Purse}`
```
Chris-Hibbert, line 11:
```suggestion
- Returns: `{IssuerKit}`
```

main/ertp/api/mint.md
```diff
@@ -1,31 +1,48 @@
...
 
 ```js
-const { issuer, mint } = makeIssuerKit('bucks');
-const mintIssuer = mint.getIssuer();
+const { issuer: quatloosIssuer, mint: quatloosMint } = makeIssuerKit('quatloos');
+const quatloosMintIssuer = quatloosMint.getIssuer();
 
 // returns true
-Object.is(issuer, mintIssuer);
+Object.is(issuer, quatloosMintIssuer);
```
Chris-Hibbert, line 28:
```suggestion
issuer === quatloosMintIssuer
```

triple-equals is more common and more useful. I had to look up `Object.is()` and learned that it's an alternative check that handles fewer cases correctly.  I'll never use it.

main/ertp/api/payment.md
```diff
@@ -1,24 +1,67 @@
...
 
 ```js
-const { issuer, mint, brand, amountMath } = makeIssuerKit('bucks');
-const payment = mint.mintPayment(amountMath.make(10));
-
+const { issuer: quatloosIssuer, mint: quatloosMint, 
+        brand: quatloosBrand, amountMath: quatloosAmountMath } = makeIssuerKit('quatloos');
+const payment = quatloosMint.mintPayment(quatloosAmountMath.make(10));
+//Should return 'quatloos'
 const officialBrand = payment.getAllegedBrand();
```
Chris-Hibbert, line 36:
```suggestion
const allegedBrand = payment.getAllegedBrand();
```

main/ertp/api/payment.md
```diff
@@ -1,24 +1,67 @@
...
+
+## Related Methods
+
+The following methods on other ERTP components also either operate
+on or return a `brand`. While a brief description is given for each, 
+you should click through to a method's main documentation entry for 
+full details on what it does and how to use it.
+
+- [`issuer.getAmountOf(payment)`](./issuer.html#issuer-getamountof-payment)
+  - Get a description of a `payment` balance as an amount`.
```
Chris-Hibbert, line 47: stray quote at EOL

main/ertp/api/payment.md
```diff
@@ -1,24 +1,67 @@
...
+- [`issuer.isLive(payment)`](./issuer.html#issuer-islive-payment)
+  - Returns `true` if `payment` has value. 
+- [`mint.mintPayment(newAmount)`](./mint.html#mint-mintpayment-newamount)
+  - Returns a new `payment` containing the newly minted assets corresponding to the `newAmount` argument. 
+- [`purse.deposit(payment, optAmount`)](./purse.html#purse-deposit-payment-optamount)
+  - Deposit all of `payment` into this `purse`.
+- [`purse.makeDepositFacet()`](./purse.html#purse-makedepositfacet)
+  - Creates a deposit-only facet on the `purse` that can be given to other parties.
+- [`purse.withdraw(amount)`](./purse.html#purse-withdraw-amount)
+  - Returns a new `payment` whose balance is described by the `amount` argument.
```
Chris-Hibbert, line 67: The link anchors don't seem to work. At least my browser (Brave or Chrome) doesn't scroll to the link on the destination page.  Also the links don't render as visible links because of the code formatting, until you hover.  I don't know how to fix this other than to do something like

```
- [burn](./issuer.html#issuer-burn-payment-optamount) `issuer.burn(payment, optAmount)`
  - Burn (destroy) all of the digital assets in the `payment`.
```

Maybe the description "you should click through to ..." is sufficient.

main/ertp/api/purse.md
```diff
@@ -1,78 +1,126 @@
...
+- Other kinds of rights, such as the right to participate in a particular contract.
+
+While each `purse` can only hold assets of one `brand`, any number of `purses` can be
+created that hold that `brand`. So you could have, say, three Quatloos `purses`, your
+friend Alice could have eight Quatloos `purses`, and so on. 
+
+You change a `purse`'s balance by calling either `deposit()` (to add assets)
+or `withdraw()` (to remove assets) on it. A `purse` can be empty, meaning it
+has a balance value of 0 (which depending on the kind of asset, could also be
+that just aren't any, say, theater tickets in the `purse`).
```
Chris-Hibbert, line 20:
```suggestion
has a balance value of 0 (which, for a ticket-holding purse, means that
there just aren't any theater tickets in the `purse`).
```

main/ertp/api/purse.md
```diff
@@ -1,78 +1,126 @@
...
+const { issuer: quatloosIssuer, mint: quatloosMint, amountMath: quatloosAmountMath } = 
+      makeIssuerKit('quatloos');
+const quatloosPurse = quatloosIssuer.makeEmptyPurse();
+const payment = quatloosMint.mintPayment(quatloosAmountMath.make(123));
+const quatloos123 = quatloosAmountMath.make(123);
 
-// Deposit a payment for 123 bucks into the purse. Ensure that this is the amount you expect.
-purse.deposit(payment, bucks123);
+// Deposit a payment for 123 Quatloos into the purse. Ensure that this is the amount you expect.
+quatloosPurse.deposit(payment, quatloos123);
```
Chris-Hibbert, line 74:
```suggestion
const depositAmount = quatloosPurse.deposit(payment, quatloos123);
```

One has to look at the return value to find out if the deposit succeeded. I dunno if you want to add a line of code checking that the returned value was 123.

main/ertp/api/purse.md
```diff
@@ -1,78 +1,126 @@
...
 ```js
 const depositOnlyFacet = purse.makeDepositFacet();
-// Give depositOnlyFacet to someone else. They can pass a payment that will be deposited:
+
+// Give depositOnlyFacet to someone else. Anyone with a deposit facet reference can tell it to receive
+// a `payment`, thus depositing the `payment` assets in the `purse` associated with the deposit facet.
 depositOnlyFacet.receive(payment);
 ```
+Once you have created a `depositFacet`, there is one method you can call 
+on it. **tyg todo: is it "recieve" or "deposit"?** `deposit.receive(payment)`.
```
Chris-Hibbert, line 99: it is `receive`

main/ertp/api/purse.md
```diff
@@ -1,78 +1,126 @@
...
+// Create a purse and give it a balance of 10 Quatloos
+const { issuer, mint } = makeIssuerKit('quatloos');
 const purse = issuer.makeEmptyPurse();
 const payment = mint.mintPayment(amountMath.make(10));
-const fungible10 = amountMath.make(10);
-purse.deposit(payment, fungible10);
+const quatloos10 = amountMath.make(10);
+purse.deposit(payment, quatloos10);
 
 // Withdraw 3 amount from the purse
```
Chris-Hibbert, line 117:
```suggestion
// Withdraw an amount of 3 from the purse
```

main/ertp/api/brand.md
```diff
@@ -21,8 +34,19 @@ const isIssuer = brand.isMyIssuer(issuer);
 ## brand.getAllegedName()
```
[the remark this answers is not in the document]
tyg: No, the first is more of a definition. I'm going to leave it in for now; a next couple of weeks project is for me to go back over the API docs and add a *lot* of links, so at that point I'll link 34 to the Glossary entry and kill 20.

main/ertp/api/purse.md
```diff
@@ -1,78 +1,126 @@
...
 - Returns: `{Amount}`
 
-Get the current `amount` contained in this purse, confirmed by the
-issuer. Note that this amount will change from call to call if assets
-have been deposited or withdrawn between calls. 
+Get an `amount` describing the current digital assets balance in the `purse`.
+This `amount` is confirmed by the `issuer` for the `brand` the `purse` can hold.
+Of course, the returned `amount` `value` might be different the next time you
+call `getCurrentAmount()` on the same `purse` if assets have been deposited or
+withdrawn from it in-between calls. .
```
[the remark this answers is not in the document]
tyg: Good to know, but can the current doc text stand? I think so, but want to check.  I think this ends up being more a digression on asynchronous programming, which perhaps should be added to the "Diestributed Programming with JavaScript" catchall programming tips doc, but I don't think fits here. Could easily be wrong though; I'm not clear on whether this was meant as educational text for me or if you'd like this included in some form in the text. Thanks.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,252 @@
...
 
 ## issuer.getBrand()
-- Returns: `{Brand}` - The brand for the issuer.
+- Returns: `{Brand}` 
 
-Get the Brand for this Issuer. The Brand indicates the kind of digital asset and is shared by the mint, the issuer, and any purses and payments of this particular kind. The brand is not closely held, so this function should not be trusted to identify an issuer alone. Fake digital assets and amounts can use another issuer's brand.
+Returns the `brand` for this `issuer`. The `brand` indicates the kind of digital asset
+and is the same for the `issuer`'s associated `mint`, and any `purses` and `payments` of this particular
+kind. The `brand` is not closely held, so this function should not be trusted to identify
+an `issuer` alone. Fake digital assets and amounts can use another `issuer's `brand`.
```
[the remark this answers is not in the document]
tyg: Honestly, I think it might just make it more confusing. The API docs don't really mention or explain vats or how we're using them. I think this also might fit best in the JavaScript Distributed Programming doc, aka tips on how to program with some behind the scenes implementation details like this to explain.

main/ertp/api/amount-math.md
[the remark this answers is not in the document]
Chris-Hibbert: that's good.

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,297 @@
...
-const badList = amountMath.subtract(listAmountA, listAmountB)
+const badList = itemsAmountMath.subtract(listAmountA, listAmountB)
 ```
+## Related Methods
+
+The following methods on other ERTP components and objects also either operate
+on or return a brand. While a brief description is given for each, you should
+click through to a method's main documentation entry for full details on
+what it does and how to use it.
+- [`issuer.getAmountOf(payment)`](./issuer.html#issuer-getamountof-payment)
```
katelynsills, line 291:
```suggestion
- <router-link to="./issuer.html#issuer-getamountof-payment">issuer.getAmountOf(payment)</router-link>
```

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,297 @@
...
+  - Returns the `amount` description of the `payment`
+- [`issuer.getAmountMathKind()`](./issuer.html#issuer-getamountmathking)
+  - Returns the kind of the `issuer`'s associated `amountMath`.
+- [`zcf.getAmountMath(brand)`](https://agoric.com/documentation/zoe/api/zoe-contract-facet.html#zcf-getamountmath-brand)
+  - Returns the `amountMath` associated with the `brand`.
+
```
katelynsills, line 297: can we use router-link here too? If we use absolute links like https://agoric.com, there's a couple of downsides. 1) it's fragile if we move the files around. 2) it shows the icon for an "external link" on the page.

main/ertp/api/payment.md
```diff
@@ -47,25 +47,28 @@ on or return a `payment`. While a brief description is given for each,
 you should click through to a method's main documentation entry for 
 full details on what it does and how to use it.
 
-- [`issuer.burn(payment, optAmount)`](./issuer.html#issuer-burn-payment-optamount)
+<router-link 
+to="/distributed-programming.html#vats">vat</router-link>
```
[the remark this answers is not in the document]
tyg: Nope. Fixed.

main/ertp/api/brand.md
```diff
@@ -49,7 +49,7 @@ The following methods on other ERTP components also either operate on or
 return a `brand`. While a brief description is given for each, you should click through
 to a method's main documentation entry for full details on what it does and how
 to use it. 
-- <router-link to="/ertp/api/issuer.html#getbrand">issuer.getBrand()</router-link>: Returns
+- <router-link to="./issuer.html#getbrand">issuer.getBrand()</router-link>: Returns
 the `brand` for the `issuer`.  
-- <router-link to="/ertp/api/payment.html#getallegedbrand">payment.getAllegedBrand()</router-link>: Returns
+- <router-link to="./payment.html#getallegedbrand">payment.getAllegedBrand()</router-link>: Returns
```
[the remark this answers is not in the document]
tyg: Is this still necessary or was it part of your tabbing issue?

main/ertp/api/amount-math.md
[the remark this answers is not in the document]
tyg: Ah. I caught one of the two, missed the other. Removed.

main/ertp/api/brand.md
```diff
@@ -21,8 +34,22 @@ const isIssuer = brand.isMyIssuer(issuer);
 ## brand.getAllegedName()
 - Returns: `{String}`
 
-Return the alleged name of the brand. Should not be trusted as accurate.
+Returns the alleged name of the `brand`. Should not be trusted as accurate.
+
+An alleged name is a human-readable string name of a kind of digital asset. 
+The alleged name should not be trusted as accurate, since it is not provided
+by a trusted source.
```
katelynsills: I think this reason isn't quite right. It's that the allegedName is given by whoever called `makeIssuerKit` and there's no registry of names of mints, so there can be many mints of the same name, so the allegedName itself shouldn't be trusted to indicate identity.

main/ertp/api/brand.md
```diff
@@ -21,8 +34,22 @@ const isIssuer = brand.isMyIssuer(issuer);
...
 ```js
 const name = brand.getAllegedName();
 ```
+## Related Methods
+
+The following methods on other ERTP components also either operate on or 
+return a `brand`. While a brief description is given for each, you should click through
+to a method's main documentation entry for full details on what it does and how
+to use it. 
+- <router-link to="./issuer.html#getbrand">issuer.getBrand()</router-link>: Returns
```
katelynsills:
```suggestion
- <router-link to="./issuer.html#issuer-getbrand">issuer.getBrand()</router-link>: Returns
```

main/ertp/api/brand.md
```diff
@@ -21,8 +34,22 @@ const isIssuer = brand.isMyIssuer(issuer);
...
+the `brand` for the `issuer`.  
+- <router-link to="./payment.html#getallegedbrand">payment.getAllegedBrand()</router-link>: Returns
```
katelynsills:
```suggestion
- <router-link to="./payment.html#payment-getallegedbrand">payment.getAllegedBrand()</router-link>: Returns
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
 # Issuer
 
-The issuer cannot mint new amounts, but it can create empty purses and payments. The issuer can also transform payments (splitting payments, combining
-payments, burning payments, and claiming payments exclusively). The issuer should be obtained from a trusted source and then relied upon as the
-authority as to whether an untrusted payment is valid.
+An `issuer` represents the identity of a specific kind of digital asset.
```
katelynsills: I would say the brand is actually the identity of the kind of asset. The Issuer is the authority on what holds digital assets of this kind.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+determine if an untrusted `payment` of that `brand` is valid.
+
+## makeIssuerKit(allegedName, amountMathKind)
+- `allegedName` `{String}` 
+- `amountMathKind` `{MathKind}` - optional
+- Returns `{IssuerKit}`
+
+Makes a new `issuer` as well as its one-to-one associated ERTP objects; a `mint` and a `brand`. 
+All are in unchangeable one-to-one relationships with each other. It also makes a new
+`amountMath`, which is in an unchangeable many-to-one relationship with the new `issuer`.
```
katelynsills:
```suggestion
All are in unchangeable one-to-one relationships with each other. 
```
I have issues with the many-to-one language, and also I think it's confusing to point out amountMath in the same way as brand if we want the users to make their own amountMath locally.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+
+The `allegedName` is available from the `brand` in asset descriptions. The `allegedName` is
```
katelynsills: You can also get the allegedName from the issuer.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+- Returns `{IssuerKit}`
+
+Makes a new `issuer` as well as its one-to-one associated ERTP objects; a `mint` and a `brand`. 
+All are in unchangeable one-to-one relationships with each other. It also makes a new
+`amountMath`, which is in an unchangeable many-to-one relationship with the new `issuer`.
+
+The `allegedName` is available from the `brand` in asset descriptions. The `allegedName` is 
+useful for debugging and double-checking assumptions, but should not be trusted.
+
+The optional `amountMathKind` specifies which of three kinds the created `amountMath` is.
```
katelynsills:
```suggestion
The optional `amountMathKind` specifies the kind of math to use with the digital assets.
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
-Makes an issuer and related objects. 
-
-The `allegedName` is available from the brand in asset descriptions. The `allegedName` is useful for debugging and double-checking assumptions, but
-should not be trusted.
-
-The `mathHelpersName` will be used to import a specific mathHelpers from the mathHelpers library. For example, `natMathHelpers`, the default, supports
-natural numbers and is used for basic fungible tokens.
+```js
+import { MathKind, makeIssuerKit } from '@agoric/ertp';
+makeIssuerKit('quatloos`); // Defaults to 'MathKind.NAT'
```
katelynsills:
```suggestion
makeIssuerKit('quatloos'); // Defaults to 'MathKind.NAT'
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+makeIssuerKit('foobars', 'MathKind.STRSET');
```
katelynsills:
```suggestion
makeIssuerKit('foobars', MathKind.STRING_SET);
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+makeIssuerKit('kitties', MathKind.SET');
```
katelynsills:
```suggestion
makeIssuerKit('kitties', MathKind.SET);
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+## issuer.getAllegedName()
+- Returns: `{allegedName}`
+
+Returns the `allegedName` for this `issuer`.
 
-Get the Brand for this Issuer. The Brand indicates the kind of digital asset and is shared by the mint, the issuer, and any purses and payments of this particular kind. The brand is not closely held, so this function should not be trusted to identify an issuer alone. Fake digital assets and amounts can use another issuer's brand.
+An alleged name is a human-readable string name 
+of a kind of digital asset. The alleged name 
+should not be trusted as accurate, since it 
+is not provided by a trusted source.
```
katelynsills: I commented elsewhere about this reason not being quite right. The real reason is that there can be multiple issuers/mints/brands with the same alleged name since there is no public registry or expectation of uniqueness. That means that the name itself can't be used to correctly identify an issuer. It's the brand that does that. The name is just the human readable version that is helpful for debugging.

katelynsills: The issuer is a very trusted source, which is part of why this is wrong. It's just that I can create an issuer with the allegedName 'Bitcoin' and it doesn't make it THE bitcoin issuer.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+- `MathKind.SET` (`set`): Used with non-fungible assets. `amount` values are objects or records with multiple properties.
 
 ```js
-const { issuer } = makeIssuerKit('bucks');
-const issuerAllegedName = issuer.getAllegedName();
-// issuerAllegedName === 'bucks'
+const { issuer: quatloosIssuer } = makeIssuerKit('quatloos');
+quatloosIssuer.getAmountMathKind(); // Returns 'MathKind.NAT', the default value.
+const { issuer: moolaIssuer } = makeIssuerKit('moola', MathKind.STR);
+moolaIssuer.getAmountMathKind(); // Returns 'MathKind.STR`
```
katelynsills:
```suggestion
moolaIssuer.getAmountMathKind(); // Returns 'str', also known as MathKind.STR
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
 ```
+## issuer.getAmountOf(payment)
+- `payment` `{Payment}`
+- Returns: `{Amount}`
 
-## issuer.getAmountMath()
-- Returns: `{AmountMath}`
+Get the `payment`'s balance. Because the `payment` is not trusted, we cannot
+trust it to provide its true value, and must rely on the `issuer` to guarantee
```
katelynsills:
```suggestion
trust it to provide its true value, and must rely on the `issuer` to validate
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
-equal to `optAmount`, to prevent sending the wrong payment and other confusion.  If `payment` is a promise, the operation will proceed after the
-promise resolves.
+Burn (destroy) all of the digital assets in the `payment`
+and return an `amount` of what was burned. 
+
+`optAmount` is optional. If `optAmount` is present, 
+the code insists the `payment` balance is equal to `optAmount`, to prevent sending the wrong `payment`
+and other confusion.  
+
+If `payment` is a `promise`, the operation proceeds after the
```
katelynsills:
```suggestion
If `payment` is a `promise` for a payment, the operation proceeds after the
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+- `optAmount` `{Amount}` 
+- Returns: `{Payment}` 
+
+Transfer all digital assets from `payment` to a new `payment` and burn the
+original. This allows the owner to be sure no other references to this
+payment survive, so they are the exclusive owner. `optAmount` is optional. 
+If `optAmount` is present, the code insists that `payment`'s balance is
+equal to `optAmount`, to prevent sending the wrong `payment` and other confusion. 
+If `optAmount` does not equal the balance in the original `payment`
+then it throws an error.  If `payment` is a promise, the operation will
```
katelynsills:
```suggestion
then it throws an error.  If `payment` is a promise for a payment, the operation will 
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
 - Returns: `{Payment}`
 
-Combine multiple payments into one payment.  If any payment in `paymentsArray` is a promise, the operation will proceed after all the payments
-resolve. The payments in `paymentsArray` will be burned.
+Combines multiple `payments` into one `payment`.  If any `payment` in `paymentsArray` is 
+a `promise`, the operation proceeds after all the `payments`
+resolve. The `payments` in `paymentsArray` are burned.
+
+If the optional `optTotalAmount` is present, the total of all the `payment` `amounts` in the
+array must equal `optTotalAmount`'s `value` or it throws an error.
```
katelynsills:
```suggestion
array must equal `optTotalAmount` or it throws an error.
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
+const oldPayment = quatloosMint.mintPayment(quatloos.AmountMath.make(100));
+const goodAmounts = Array(10).fill(quatloosAmountMath.make(10));
 
-Note that the total amount in the `amountArray` must equal the amount in the original `payment`:
+const arrayOfNewPayments = quatloos.Issuer.splitMany(oldPayment, goodAmounts);
 
-```js
-const { mint, issuer, amountMath } = makeIssuerKit('fungible');
-const payment = mint.mintPayment(amountMath.make(1000));
+// The total amount in the amountArray must equal the original payment amount
```
katelynsills: This comment seems slightly misplaced. Should it go a couple of lines below? Or should we change it?

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,263 @@
...
-issuer.splitMany(payment, badAmounts);
+quatloosIssuer.splitMany(payment, badAmounts);
 ```
 
 ## issuer.isLive(payment)
 - `payment` `{Payment}`
 - Returns: `{Boolean}`
 
-Returns true if the payment continues to have value. If `payment` is a promise, the operation will proceed upon resolution.
+Returns `true` if the `payment` continues to have value. If `payment` is a promise, the operation proceeds upon resolution.
```
katelynsills:
```suggestion
Returns `true` if the `payment` was created by the issuer and has not yet been destroyed. If `payment` is a promise, the operation proceeds upon resolution.
```

main/ertp/api/purse.md
```diff
@@ -1,78 +1,130 @@
...
+- Other kinds of rights, such as the right to participate in a particular contract.
+
+While each `purse` can only hold assets of one `brand`, any number of `purses` can be
+created that hold that `brand`. So you could have, say, three Quatloos `purses`, your
+friend Alice could have eight Quatloos `purses`, and so on. 
+
+You change a `purse`'s balance by calling either `deposit()` (to add assets)
+or `withdraw()` (to remove assets) on it. A `purse` can be empty, meaning it
+has a balance value of 0 (which, for a ticket-holding `purse`, means that
+there just aren't any theater tickets in the `purse`).
```
katelynsills: I can see that we're trying to accommodate the nonfungible side here too, but can we remove the value of 0? Technically, the value could be []. So although I can see what we're trying to do, I think it is technically incorrect.
```suggestion
or `withdraw()` (to remove assets) on it. A `purse` can be empty. (A purse for a currency
has a balance value of 0, but a ticket-holding purse just doesn't have any tickets).
```

tyg: How about adding "fungible" to have "A purse for a fungible currency has a balance value of 0, but..." to make the difference clear?

katelynsills: sounds good!

main/ertp/api/brand.md
```diff
@@ -1,18 +1,31 @@
...
-Human-readable string name of a kind of digital asset. The alleged name should
-not be trusted as an accurate depiction, since it is provided by
-the maker of the issuer and could be deceptive.
+The `brand` identifies the kind of assets. A particular `brand` has one-to-one relationships within a
+contract `instance`
+with an `issuer` and a `mint`, and one-to-many relationships with `amountMaths`, `purses` and `payments`.
+Meaning for, say, the Quatloos `brand`:
+- There is only one Quatloos `issuer`.
+- There is only one Quatloos `mint`.
+- All `amounts` created via the Quatloos `amountMath` have a Quatloos `brand`.
```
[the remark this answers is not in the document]
tyg: That seems a bit tautological to me.  How about "All amounts created by a Quatloos amountMath.make() method have a Quatloos brand"?

main/ertp/api/brand.md
```diff
@@ -1,18 +1,31 @@
...
+- There can be many local copies of the Quatloos `amountMath`, but those `amountMaths` only work on Quatloos branded `amounts`.
+- There can be any number of Quatloos holding `purses`, but those `purses` can only hold Quatloos.
+- There can be any number of Quatloos holding `payments`, but those `payments` can only hold Quatloos.
+
+But recipients of a 
+purported `amount` cannot use the `brand` by itself to verify its authenticity,
+since the `brand` can be reused by a misbehaving `issuer`
+
+All of these relationships are unchangable; if, say, a `mint` is created that makes new Quatloos, it
+can never create new assets of any other `brand`.`
```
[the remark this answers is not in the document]
tyg: That's worth adding, but it's not what I was trying to get across. If a mint makes Xs, it will always make Xs. It can't start making Ys. How about:

"All of these relationships are unchangable; if, say, another `mint` is created, it
can never create Quatloos. It has a different `brand`. And once a Quatloos mint is
created, it can only make Quatloos; it can never change and mint Moola, Simoleans, or anything
other than Quatloos"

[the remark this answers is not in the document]
katelynsills: Sounds good!

main/ertp/api/issuer.md
tyg: How's this: "While this also returns an `amountMath`, we strongly urge that instead of using that one, which
is asynchronous, you use `makeLocalAmountMath(issuer)` to create a local, synchronous, version
and use that one."

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,297 @@
...
+`mint`, `issuer`, and `brand`.
+
+The advantage of making and using a local `amountMath` is that it's synchronous.
+
+## makeLocalAmountMath(issuer)
+- `issuer`: `{issuer}`
+Returns: `{ AmountMath }`
+
+Creates and returns a local (synchronous) `amountMath` object. The new local copy has the same many-to-one
+relationship with the original's `issuer`, and thus also to the original's `brand` and `mint`.
```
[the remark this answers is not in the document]
tyg: Huh. It seems odd that a local amountMath doesn't have a connection to issuer, given that those are created by makeLocalAmountMath(issuer). Could you explain why it doesn't given that, 'cause that's why I was thinking it did.

main/ertp/api/payment.md
```diff
@@ -1,24 +1,70 @@
...
+original balance, or it is used up entirely. You cannot partially use a
+`payment`. In other words, if a `payment` is 10 Quatloos, you can't
+take out, say, 3 Quatloos from it, leaving the `payment` with 7 Quatloos.
 
-Payments can be converted to Purses by getting a trusted issuer and calling `issuer.makeEmptyPurse()` to create a purse, then `purse.deposit(payment)`.
+However, you can split a `payment` into multiple `payments`, for example 
+into two new `payments` of 3 Quatloos and 7 Quatloos.
+The `split()` operation *burns* (destroys) the original 10 Quatloos `payment`.
+
+`Payments` are often received from other actors and therefore should not be
```
Chris-Hibbert, line 16: the "therefore" isn't right here. It isn't the fact that you got it from someone else that makes them untrusted. It's that they're not self-authenticating. 

Issuers are the other case. If someone gives you an alleged issuer, and says "this is the currency in which I want to be paid" or "this currency represents rights to access my thingamajig", there is no alternate source of authority. If you can get payments or purses containing amounts that match that issuer, then you can participate, otherwise not.

main/ertp/api/purse.md
```diff
@@ -1,78 +1,134 @@
...
+friend Alice could have eight Quatloos `purses`, and so on. 
+
+You change a `purse`'s balance by calling either `deposit()` (to add assets)
+or `withdraw()` (to remove assets) on it. A `purse` can be empty, which if it holds 
+a fungible currency means it has a value of 0. If it holds non-fungible theatre tickets, 
+it means it just doesn't have any tickets.
+
+meaning it
+has a balance value of 0 (which, for a ticket-holding `purse`, means that
+there just aren't any theater tickets in the `purse`).
```
Chris-Hibbert, line 24: redundant

main/ertp/api/purse.md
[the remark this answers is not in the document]
Chris-Hibbert: Mostly yes, the current text can stand. I think the fact that `deposit()` uniquely requires a non-promise (and the implications) should be called out explicitly in the entry for `deposit()`.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,275 @@
...
+- `allegedName` `{String}` 
+- `amountMathKind` `{MathKind}` - optional
+- Returns `{IssuerKit}`
+
+Makes a new `issuer` as well as its one-to-one associated ERTP objects; a `mint` and a `brand`. 
+All are in unchangeable one-to-one relationships with each other. 
+
+While this also returns an `amountMath`, we strongly urge that instead of using that one, which
+is asynchronous, you use `makeLocalAmountMath(issuer)` to create a local, synchronous, version
+and use that one.
```
Chris-Hibbert, line 25: The party that gets an `amountMath` by calling `makeIssuerKit()` *should* use that one, because it's synchronous for them. They just shouldn't share it outside the vat, because any code receiving it remotely will get an asynchronous presence. Outside the originating vat, you should make a local copy, for reasons explained several times.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,275 @@
...
+is asynchronous, you use `makeLocalAmountMath(issuer)` to create a local, synchronous, version
+and use that one. 
+
+The `allegedName` is available from the `brand` or the `issuer` in asset descriptions. The `allegedName` is 
+useful for debugging and double-checking assumptions, but should not be trusted.
+
+The optional `amountMathKind` specifies the kind of math to use with the digital assets. 
+Each implements all of the same set of API methods (i.e. `amountMath` methods are 
+polymorphic). We recommend you import the `MathKind` values from `@agoric/ERTP` 
+instead of making the strings yourself.
```
Chris-Hibbert, line 33:
```suggestion
polymorphic). We recommend you import the `MathKind` values from `@agoric/ERTP` 
instead of using strings.
```
I don't think JavaScript programmers think of using 'foo' as  making a string, though that' s a fair description of the effect.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,275 @@
...
-Get the `allegedName` for this mint/issuer.
+Get the kind of this `issuer`'s `amountMath`. It returns one of
+`MathKind.NAT` (`nat`), `MathKind.STR` (`str`), or `MathKind.STRING_SET` (`strSet`).
+
+The `amountMathKind` value specifies which of three kinds an `amountMath` is,
+and what kind of values it is used on. Each kind implements all of the same set 
+of API methods (i.e. `amountMath` methods are polymorphic). 
+- `MathKind.NAT` (`nat`): Used with fungible assets. `amount` values are natural numbers (non-negative integers). Default value.
+- `MathKind.STRING_SET` (`strSet`): Used with non-fungible assets. `amount` values are strings.
+- `MathKind.SET` (`set`): Used with non-fungible assets. `amount` values are objects or records with multiple properties.
```
Chris-Hibbert, line 92: I don't think it's helpful to explain that `MathKind.NAT` corresponds to `'nat'` everytime MathKind is used. We defined MathKind so they don't have to be aware of the strings.  I would remove the strings almost everywhere.  (At least almost everywhere on this page.)

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,275 @@
...
-const originalPayment = mint.mintPayment(amountExpectedToTransfer);
+const { mint: quatloosMint, issuer: quatloosIssuer, amountMath: quatloosAmountMath } = makeIssuerKit('quatloos');
+const amountExpectedToTransfer = quatloosAmountMath.make(2);
+const originalPayment = quatloosMint.mintPayment(amountExpectedToTransfer);
 
-const newPayment = issuer.claim(originalPayment, amountToTransfer);
+const newPayment = quatloosIssuer.claim(originalPayment, amountToTransfer);
 ```
 
 ## issuer.combine(paymentsArray)
```
Chris-Hibbert, line 185:
```suggestion
## issuer.combine(paymentsArray, optTotalAmount)
```

katelynsills:
Replying to https://github.com/Agoric/documentation/pull/249#discussion_r480257762 (I can't reply normally for some reason)

> These suggestions to remove the line defining quatloosAmountMath based on makeIssuerKit() are marked as resolved, but the lines are still in the examples. @katelynsills why do you think it's better without? (Perhaps because the user of an amountMath is only rarely the creator of the issuer?)

Yes, I want to remove these because the user of amountMath will almost never make an amountMath in this way if they are writing or using Zoe contracts.

main/ertp/api/issuer.md
tyg: The reason they're there so many times is that people usually use API references one method at a time, rather than going through from the start (random vs. sequential access : -)). So I shouldn't put something about a method just once at the top, even if I already wrote it about another method. I think the question here is remove the 'nat' etc. strings from everywhere in the API doc except maybe the top of the Amount Math section and just mention the values in the ERTP Guide, or leave them in as is. Can go either way, but wanted to be sure you knew why I had them in there before making the decision.

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,284 @@
...
+- `MathKind.NAT` (`nat`): Used with fungible assets. `amount` values are natural numbers (non-negative integers). Default value.
+- `MathKind.STRING_SET` (`strSet`): Used with non-fungible assets. `amount` values are strings.
+- `MathKind.SET` (`set`): Used with non-fungible assets. `amount` values are objects or records with multiple properties.
 
 ```js
-const { issuer } = makeIssuerKit('bucks');
-const issuerAllegedName = issuer.getAllegedName();
-// issuerAllegedName === 'bucks'
+const { issuer: quatloosIssuer } = makeIssuerKit('quatloos');
+quatloosIssuer.getAmountMathKind(); // Returns 'nat', also known as 'MathKind.NAT', the default value.
```
katelynsills, line 93:
```suggestion
quatloosIssuer.getAmountMathKind(); // Returns 'nat', also known as MathKind.NAT, the default value.
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,284 @@
...
+const { issuer: moolaIssuer } = makeIssuerKit('moola', MathKind.STR);
```
katelynsills, line 94:
```suggestion
const { issuer: moolaIssuer } = makeIssuerKit('moola', MathKind.STRING_SET);
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,284 @@
...
+moolaIssuer.getAmountMathKind(); // Returns 'str', also known as'MathKind.STR`
```
katelynsills, line 95:
```suggestion
moolaIssuer.getAmountMathKind(); // Returns 'str', also known as'MathKind.STRING_SET`
```

main/ertp/api/issuer.md
```diff
@@ -1,214 +1,284 @@
...
-## issuer.getMathHelpersName()
-- Returns: `{String}`
+## issuer.getBrand()
+- Returns: `{Brand}` 
 
-Get the name of the MathHelpers for this Issuer.
+Returns the `brand` for this `issuer`. The `brand` indicates the kind of digital asset
+and is the same for the `issuer`'s associated `mint`, and any `purses` and `payments` of this particular
+kind. The `brand` is not closely held, so this function should not be trusted to identify
+an `issuer` alone. Fake digital assets and amounts can use another `issuer's `brand`.
```
katelynsills, line 117:
```suggestion
an `issuer` alone. Fake digital assets and amounts can use another issuer's `brand`.
```

Missing `

main/ertp/api/mint.md
```diff
@@ -1,31 +1,52 @@
...
+new assets of that `brand`.
+
+## makeIssuerKit(allegedName, amountMathKind)
+- `allegedName` `{String}`
+- `amountMathKind` `{MathKind}`
+- Returns: `{IssuerKit}`
+
+While not a method called on a `mint`, clearly it's important to know how to create a new `mint`. 
+`makeIssuerKit()` returns a new `issuer`, `mint`, `amountMath`, and `brand`. 
+See <router-link to="./issuer.md#makeissuerkit-allegedname-amountmathkind">here</router-link> for details.
```
katelynsills, line 16: 404

main/ertp/api/mint.md
```diff
@@ -1,31 +1,52 @@
...
-Object.is(issuer, mintIssuer);
+issuer === quatloosMintIssuer;
 ```
 
 ## mint.mintPayment(newAmount)
 - `newAmount` `{Amount}`
 - Returns: `{Payment}`
 
-Create a new Payment containing newly minted amount.
+**Important**: `mint.mintPayment()` is the <ins>only</ins> way
```
katelynsills, line 36: Thanks for adding this!

main/ertp/api/brand.md
```diff
@@ -1,18 +1,32 @@
...
+- There can be any number of Quatloos holding `purses`, but those `purses` can only hold Quatloos.
+- There can be any number of Quatloos holding `payments`, but those `payments` can only hold Quatloos.
 
-## AllegedName
-Human-readable string name of a kind of digital asset. The alleged name should
-not be trusted as an accurate depiction, since it is provided by
-the maker of the issuer and could be deceptive.
+But recipients of a 
+purported `amount` cannot use the `brand` by itself to verify its authenticity,
+since the `brand` can be reused by a misbehaving `issuer`
```
katelynsills, line 15:
```suggestion
since the `brand` can be reused by a misbehaving `issuer`.
```

main/ertp/api/brand.md
```diff
@@ -21,8 +35,29 @@ const isIssuer = brand.isMyIssuer(issuer);
...
 const name = brand.getAllegedName();
 ```
+## Related Methods
+
+The following methods on other ERTP components also either operate on or 
+return a `brand`. While a brief description is given for each, you should click through
+to a method's main documentation entry for full details on what it does and how
+to use it. 
+- <router-link to="./issuer.html#issuer-getbrand">issuer.getBrand()</router-link>: Returns
+the `brand` for the `issuer`.
```
katelynsills, line 60: Is this newline intended?

main/ertp/api/purse.md
```diff
@@ -1,78 +1,147 @@
...
+```
 
 ## purse.getCurrentAmount()
 - Returns: `{Amount}`
 
-Get the current `amount` contained in this purse, confirmed by the
-issuer. Note that this amount will change from call to call if assets
-have been deposited or withdrawn between calls. 
+Get an `amount` describing the current digital assets balance in the `purse`.
+This `amount` is confirmed by the `issuer` for the `brand` the `purse` can hold.
```
katelynsills, line 43:
```suggestion
```

katelynsills: This doesn't make sense. I think if this came from the types files, we have an issue on agoric-sdk to remove it.

main/ertp/api/purse.md
```diff
@@ -1,78 +1,147 @@
...
 ## purse.deposit(payment, optAmount)
 - `payment` `{Payment}`
-- `optAmount` `{Amount}` - Optional. This parameter ensures you are depositing the amount you expect.
+- `optAmount` `{Amount}` - Optional. 
 - Returns: `{Amount}`
 
-Deposit all the contents of `payment` into this purse, returning the payment's amount
-of digital assets (i.e. the deposit amount). If the optional argument `optAmount` does not equal the balance of
-`payment`, or if `payment` is an unresolved promise, throw an error.
+All I came up with that is specific to deposit was this:
```
katelynsills, line 62: Is this an internal comment that accidentally got in??

main/ertp/api/purse.md
```diff
@@ -1,78 +1,147 @@
...
+1) you have to have a non-promise before you can call deposit(). When you call deposit() you'll get a response back (after waiting only for the round trip) telling you whether it succeeded. All calls that arrive later will see that the value has been transferred into the purse, and the payments are no longer valid.
```
katelynsills, line 63:
```suggestion

```
Remove? I think this was an accident

main/ertp/api/purse.md
```diff
@@ -1,78 +1,147 @@
...
 ## purse.withdraw(amount)
 - `amount` `{Amount}`
 - Returns: `{Payment}`
 
-Withdraw the `amount` from this purse into a new Payment.
+Withdraw the `amount` specified digital assets from this `purse` into a new `payment`.
+
+If the call succeeds, it immediately extracts the value into a new `payment`. 
+The caller won't get the new `payment` until a later turn, since the call is (nearly always) remote.
+But as soon as the message is processed, the value is gone from the `purse`.
```
katelynsills, line 128: Hmm, where is this coming from? if we are calling it in this way `purse.withdraw`, then we are assuming the purse is local, so the caller gets the result synchronously, in the same turn. If we are thinking the purse is remote, then we need to call it like `E(purse).withdraw` I think it's best to stick to the purse being local, which means we should remove this
```suggestion
```

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,294 @@
 # Amount Math
 
-Logic for manipulating amounts.
+Logic for manipulating `amounts`.
+
+## Obtaining an AmountMath
+
+There are three ways and circumstances you can get access to an `amountMath`:
+- You made its associated `mint` and `issuer` with `makeIssuerKit()`, so use the `amountMath` returned from the call. 
+- You are writing a Zoe contract and the `issuer` is saved in Zoe, so call `zcf.getAmountMath(brand)`..
```
katelynsills, line 9:
```suggestion
- You are writing a Zoe contract and the `issuer` is saved in Zoe, so call `zcf.getAmountMath(brand)`.
```

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,294 @@
...
+- `MathKind.NAT` (`nat`): Used with fungible assets. `amount` `values` are natural numbers (non-negative integers).
+- `MathKind.STRING_SET` (`strSet`): Used with non-fungible assets. `amount` `values` are strings.
+- `MathKind.SET` (`set`): Used with non-fungible assets. `amount` `values` are objects or records with multiple properties.
+
+Use `makeIssuerKit(allegedName, MathKind)` to specify which `amountMath` 
+kind your contract uses. The second parameter, `MathKind` is optional and 
+defaults to `MathKind.NAT` if not given. For example
+```js
+import { MathKind, makeIssuerKit } from '@agoric/ertp';
+makeIssuerKit('quatloos'); // Defaults to 'MathKind.NAT'
```
katelynsills, line 27:
```suggestion
makeIssuerKit('quatloos'); // Defaults to MathKind.NAT
```

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,294 @@
...
+that already has 3 Quatloos gives a new balance of 5 Quatloos. 
+An empty `purse` has 0 Quatloos. 
 
 ```js
 someAmount: {
-  brand,
-  value: someValue
+  brand: someBrand,
 }
+  value: someValue,
```
katelynsills, line 46: `}` should come after `someValue`

main/ertp/api/amount-math.md
```diff
@@ -1,189 +1,294 @@
...
+## makeLocalAmountMath(issuer)
+- `issuer`: `{issuer}`
+- Returns: `{ Promise<AmountMath> }`
 
-Value must be Comparable.
+Creates and returns a local `amountMath` object. The new
+local copy uses the same remote `brand` as the `issuer` does.
+
+This should be used when you need an `amountMath`, and receive or 
+learn about an `issuer`, and are not writing a Zoe contract,
```
katelynsills, line 61:
```suggestion
learn about an `issuer`, and are not writing a Zoe contract.
```

main/ertp/api/purse.md
tyg: It's coming from a Keybase conversation with Chris about one of his comments. 
Me: "You made a comment in purse.md " I think the fact that deposit() uniquely requires a non-promise (and the implications) should be called out explicitly in the entry for deposit()."   What are the implications, so I can add them to the deposit() section? Thanks."

main/ertp/api/amount-math.md
```diff
@@ -290,5 +289,5 @@ what it does and how to use it.
   - Returns the `amount` description of the `payment`
 - <router-link to="./issuer.html#issuer-getamountmathkind">`issuer.getAmountMathKind()`</router-link>
   - Returns the kind of the `issuer`'s associated `amountMath`.
-- <router-link to="./zoe/api/zoe-contract-facet.html#zcf-getamountmath-brand">`zcf.getAmountMath(brand)`</router-link>
+-  [`zcf.getAmountMath(brand)`](https://github.com/Agoric/documentation/master/main/zoe/api/zoe-contract-facet.md#zcfgetamountmathbrand)
```
katelynsills, line 292: Not the right URL

main/ertp/api/mint.md
```diff
@@ -13,7 +13,7 @@ new assets of that `brand`.
 
 While not a method called on a `mint`, clearly it's important to know how to create a new `mint`. 
 `makeIssuerKit()` returns a new `issuer`, `mint`, `amountMath`, and `brand`. 
-See [here](./issuer.md#makeissuerkit-allegedname-amountmathkind) for details.
+See <router-link to="./issuer.md#makeissuerkitallegedname-amountmathkind">here</router-link> for details.
```
katelynsills, line 16: Not the right link

main/ertp/api/purse.md
katelynsills: Ok, I'm wondering if we should be adding so much new material when we're just trying to get this out. This new material has nuances, as I mentioned above. In the rest of the doc, we describe the purse as if it is local. Chris' analysis assumes it is remote. We will need to explain why we are suddenly shifting perspective.

main/ertp/api/purse.md
Chris-Hibbert: yep, that's my note to Tom. Shouldn't be here.

main/ertp/api/purse.md
Chris-Hibbert: Why would it make sense to present the API as if most interactions are with a local purse? There are certainly aspects of the interaction for which the distinction isn't relevant but nearly all interactions are with remote purses, so giving developers an awareness of the asynchronous behavior where it's pertinent seems important.

katelynsills: It's certainly correct that most interactions will be with a remote purse. I'm just pointing out that this document, with this new  inclusion, will be remarkably inconsistent. We need to explain that this particular analysis is for a remote purse, and our instructions previously (in which we mention calling `purse.deposit` or `purse.withdraw`) only work on local purses. If the purse is remote, we must do `E(purse).deposit` and `E(purse).withdraw`. It's the inconsistency that is the problem, not the analysis of remote purses.

PR #272, opened 2020-09-10
Tightened and clarified prose, fixed link

tyg:
This was all straightforward and basic conceptual. Did have one question
about what exitRule's default value is.

### DIFF
--- a/main/zoe/guide/offer-safety.md
+++ b/main/zoe/guide/offer-safety.md
@@ -55,8 +55,8 @@ one is true. Both could also be true.
 
 If there are no rules or `give` is omitted, then you get a full refund
 of what you put in, fulfilling offer safety. It's just the case that
-you put in nothing, so you get back nothing. However, it may also be
-your lucky day; you *might* also get what you wanted, depending on whether your offer of
+you put in nothing, so you get back nothing. However, you *might* also 
+get what you wanted, depending on whether your offer of
 nothing for it is accepted.
 
 **What if there are no rules under `want`**
--- a/main/zoe/guide/offer-safety.md
+++ b/main/zoe/guide/offer-safety.md
@@ -8,8 +8,9 @@ get what they wanted or get back a full refund of what they offered.
 For Zoe to enforce offer safety, the user must
 give Zoe a `proposal`. This is a description of both what they want and what
 they are offering, and when and how the user can exit the contract. Zoe uses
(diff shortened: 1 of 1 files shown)


main/zoe/guide/offer-safety.md
```diff
@@ -2,66 +2,61 @@
...
+get what they wanted or get back a full refund of what they offered.
 
-In order for Zoe to be able to enforce offer safety, the user must
-provide Zoe a description of what they want and a description of what
-they are offering. This is called a `proposal`. A proposal also
-contains the conditions under which a user can exit a contract (this is used by Zoe to enforce payout liveness).
+For Zoe to enforce offer safety, the user must
+give Zoe a `proposal`. This is a description of both what they want and what
+they are offering, and when and how the user can exit the contract. Zoe uses
+this last to enforce payout liveness. The exit specification is optional. **tyg todo: Huh. Looking at types.js, and
```
katelynsills, line 11: `onDemand` is the default. It may not be in the types.

main/zoe/guide/offer-safety.md
```diff
@@ -2,66 +2,61 @@
...
+Yes, you can get a full refund *and* what you wanted. Offer safety guarantees at least 
+one is true. Both could also be true.
 
-### What if there are no rules under `give`?
+**What if there are no rules under `give`?**
 
-If there are no rules under `give` or `give` is omitted, then
-trivially, whatever you get will fulfill offer-safety, because you are
-always getting a full refund of what you put in, which was nothing.
+If there are no rules or `give` is omitted, then you get a full refund
```
katelynsills, line 55: This sounds like they won't get what they wanted. Can we state that they may also get what they wanted in addition to the refund in this case?

Chris-Hibbert: We've thought of several examples where a contract would expect to payout for no explicit payment. e.g. a coat check or a drop-box would commit to giving the proceeds to whomever has the invitation.

tyg: I'm not sure I follow. Say for coat check, do you mean the invitation is the equivalent of a numbered ticket indicating which slot my coat is in, because I'd think the ticket would an asset be what I gave to get my coat (i.e. the right to whatever is in slot 56)?.

Chris-Hibbert (approved):
These changes are all improvements. We can come back later and make it even better.

PR #273, opened 2020-09-10
Updated to Alpha, expanded from proposals to offers.

tyg:


main/zoe/guide/proposal.md
```diff
@@ -1,41 +1,68 @@
...
 
-## Structure of the proposal
+## Invitations
+
+`Invitations` are a special case of ERTP `payments`. They are linked to a specific contract `instance`, and
+having one gives you the right to participate in that contract `instance`, for example, by making offers in it.
+
+There are two main ways for contract users to get an `invitation`:
+- If you create the contract `instance`, you get a special creator `invitation`.
+- Someone who holds the right objects has created an `invitation` for a contract `instance` and given it to
```
Chris-Hibbert, line 19:
```suggestion
- Someone (possibly you) who holds the right objects has created an `invitation` for a contract `instance` and gives it to
```

"Someone ... gives it to you" is the subject of "two main ways ... to get an invitation"

main/zoe/guide/proposal.md
```diff
@@ -1,41 +1,68 @@
...
+  you in some way. This could've been by sending it to you, posting it on a public online location, etc. It
+  doesn't matter (nor does Zoe specify or have any requirements) how or why it got to you, only that you have it.
+
+## Proposals
 
 Proposals are objects with `give`, `want` and `exit` keys.
```
Chris-Hibbert, line 25: I'm not sure we're consistent about this, but I think we should be saying that Proposals are records. Javascript probably doesn't distinguish, but the switch from records to objects was made because objects have methods and behavior. Proposals have fields, but no methods, to they're JavaScript Records.

main/zoe/guide/proposal.md
```diff
@@ -1,41 +1,68 @@
...
+- `'onDemand'`: (Default) Whenever the user wants.
+- `'waived'`: The user cannot cancel, relying on the contract to finish the offer.
+- `'afterDeadline'`: Cancelled automatically after a deadline. This requires two
+  more properties, a `timer` object and a deadline value.
+
+## Escrowed Payments
+
+Using the same keywords as your `proposal`, you must specify a `PaymentKeywordRecord`.
+This is a record with the keywords as keys, and `payments` containing digital assets as
+values. Zoe escrows these `payments` until the offer is completed or rejected.
```
Chris-Hibbert, line 56:
```suggestion
values. Zoe escrows these `payments` on behalf of this seat until the offer is completed or rejected or the assets are reassigned to another seat. 
```

tyg: Changed, but I didn't have time to upgrade this to cover seats (the original only did proposals; offers was in-between). So could "seats" be changed to "offers" here for the moment until this can be upgraded to cover seats?

Chris-Hibbert: sure.

Chris-Hibbert (approved):
this is good.

PR #276, opened 2020-09-10
Update Zoe Guide README to Alpha

tyg:


Chris-Hibbert (approved):
This is a good cleanup.

main/zoe/guide/README.md
```diff
@@ -64,176 +63,170 @@ handle the credit cards themselves.
...
-      price: want.Price,
-      inviteDesc: 'matchOffer',
-    },
-  });
-};
+{ give: { Asset: amount1 }, want: { Price: amount2 } }
 ```
-In the `makeMatchingInvite` method we call the [`swap` helper function](../api/zoe-helpers.html#zoehelper-swap-keephandle-tryhandle-keephandleinactivemsg), which handles a lot of the logic. The code for `swap` is:
+`amount1` and `amount2` are amounts with the correct issuers for the keywords.
+`assertProposalShape()` then pulls out the elements **tyg todo: Which are the elements? The amounts?** of the
```
dtribble, line 173: Replace with:
The contract then uses `getProposal` to extract the properties of the proposal for later matching.

main/zoe/guide/README.md
```diff
@@ -64,176 +63,170 @@ handle the credit cards themselves.
...
 
-Once we're sure that we *do* have a matching offer, we can do the most
-exciting part, the reallocation.
+`makeMatchingInvitation()`, our first handler, then constructs a handler for the second offer,
+with the first offer's `want` and `give` in scope. This second
+handler, `matchingSeatOfferHandler()` does the final step.
+It uses the [`swap` helper  function](../api/zoe-helpers.html#swap-zcf-leftseat-rightseat-lefthasexitedmsg-righthasexitedmsg),
+a powerful Zoe Helper that handles a lot of the logic of doing a basic swap of assets.
 
+When the swap succeeds, it exits both seats and the contract shuts down. **tyg todo: Should we mention reallocation of assets between the parties?**
```
dtribble, line 191: If the swap succeeds, it reallocates the assets between the parties, as described above. The handler then exits both seats (causing payouts to be made available to both parties) and shuts down the contract.

main/zoe/guide/README.md
```diff
@@ -64,176 +63,170 @@ handle the credit cards themselves.
...
+```js
+    const matchingSeatOfferHandler = matchingSeat => {
+      const swapResult = swap(zcf, firstSeat, matchingSeat);
+      zcf.shutdown();
+      return swapResult;
+    };
+```
+Now let's put it together. The last step of the first handler, `makeMatchingInvitation()`
+is to create and return the second party's invitation, using
+`matchingSeatOfferHandler()` and including custom properties for the expected proposal.  **tyg todo: Maybe "offer proposal" to reinforce the connection?**
```
dtribble, line 201: no parentheses:  matchingSeatOfferHandler.

dtribble, line 201: `matchingSeatOfferHandler` and including custom properties for the proposal of the invited offer.

main/zoe/guide/README.md
```diff
@@ -64,176 +63,170 @@ handle the credit cards themselves.
...
+    makeMatchingInvitation,
+    'firstOffer',
+  );
+  return { creatorInvitation };
+};
+```
+The `creatorInvitation` is only available to the contract instance's creator.
+(see [`startInstance`](../api/zoe.html#e-zoe-startinstance-installation-issuerkeywordrecord-terms)).
+The creator can use it (by making an offer with it) or send it to some other party.
+**tyg todo: I think the comment below is missing a close comment? Or does it just not recognize the
```
dtribble, line 227: I forgot to delete the other start comment. I was just adding this generally to the comment in case we want to later revivify it.

PR #278, opened 2020-09-11
Added information to make explicit how invitations are minted.

tyg:


main/zoe/api/zoe-contract-facet.md
```diff
@@ -161,7 +162,9 @@ await zcf.saveIssuer(secondaryIssuer, keyword);
 
 Make a credible Zoe `invitation` for a smart contract. Note that `invitations` are a special case
 of an ERTP `payment`. They are in a one-to-one relationship with the `invitationIssuer`, which is used
-to validate invitations and their `amounts`.
+to validate invitations and their `amounts`. They are also in a one-to-one relationship with
+a `mint` devoted to minting new `invitations`. `zcf.makeInvitation()` serves as an interface to
```
Chris-Hibbert, line 166: I wouldn't do this as two separate sentences.
"They are in a one-to-one relationship with the `invitationIssuer` and its mint, which are used
to mint invitations and validate their `amounts`."

Chris-Hibbert (approved):
This is good after condensing the second description.

main/zoe/api/zoe-contract-facet.md
```diff
@@ -127,7 +127,8 @@ mySynchronousMint.mintGains({ MyKeyword: amount }, seat);
 Zoe has a single `invitationIssuer` for the entirety of its
 lifetime. This method returns the Zoe `InvitationIssuer`, which
 validates user-received `invitations` to participate
-in contract instances.
+in contract instances. It has a one-to-one association with a
```
katelynsills, line 130: This may be my personal quirk, but it seems wrong to me to use the words "one-to-one" when the relationship in one direction is "one-to-many". There are many invitations per issuer and per mint, so I think one-to-one gives the wrong connotations. Instead, we can say it shares the same brand as the invitation issuer and invitation mint.

main/zoe/api/zoe-contract-facet.md
katelynsills, line 166: I like the comment about the interface to the mint. See above for the one-to-one comment

katelynsills (changes requested):
The doc looks good in general, but I feel like the one-to-one comment is not right.

main/zoe/api/zoe-contract-facet.md
Chris-Hibbert: Did you mis-read this? I think it says the issuer and mint are one-to-one.  The mint is also "devoted to minting invitations" but those aren't described as one-to-one.

katelynsills: Ah, sorry. Yes, the issuer and mint are one-to-one.

katelynsills: Wait, it says "they"? Is it invitations or the issuer?

main/zoe/api/zoe-contract-facet.md
Chris-Hibbert: > Wait, it says "they"? Is it invitations or the issuer?

The only 'they's are down here, so I'm going to respond here.

I agree that one-to-one doesn't apply here. All invitations share a single unique issuer, but there are many invitations associated with many contracts. The single InvitationIssuer is tightly coupled with a single invitationMint, but only Zoe ever sees the invitationMint.

So let me change my suggestion to 

"All invitations come from a single invitation issuer and its mint, which are used
to mint invitations and validate their amounts."

Chris-Hibbert (changes requested):

main/zoe/api/zoe-contract-facet.md
tyg: Point. I blame writing on insufficient sleep. : -)

tyg: Although in this bit, I was referring to the one-to-one between the InvitationIssuer and the invitation mint, which is one-to-one. It does apply to the later bit though, and still rephrasing.

PR #283, opened 2020-09-13
Edit pass and date format update on Simple Exchange

tyg:
Should 

{ give: { Asset: simoleans(5) }, want: { Price: quatloos(3) } }

{ give: { Price: quatloos(8) }, want: { Asset: simoleans(3) } }

use simoleansAmountMath.make(5) and similar instead? Or is this just the format of the order book?

### DIFF
--- a/main/zoe/guide/contracts/simple-exchange.md
+++ b/main/zoe/guide/contracts/simple-exchange.md
@@ -2,7 +2,7 @@
 
 <Zoe-Version/>
 
-##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/f29591519809dbadf19db0a26f38704d87429b89/packages/zoe/src/contracts/simpleExchange.js) (Last updated: 12-SEP-2020)
+##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/f29591519809dbadf19db0a26f38704d87429b89/packages/zoe/src/contracts/simpleExchange.js) (Last updated: 2020-9-12)
 ##### [View all contracts on Github](https://github.com/Agoric/agoric-sdk/tree/master/packages/zoe/src/contracts)
 
 The "simple exchange" is a very basic, un-optimized exchange. It
@@ -24,8 +24,14 @@ accepted in either direction.
 { give: { Asset: simoleans(5) }, want: { Price: quatloos(3) } }
 { give: { Price: quatloos(8) }, want: { Asset: simoleans(3) } }
 ```
+Note: Here we used a shorthand for assets whose values are 5 simoleons, 3 
+quatloos, 8 quatloos, and 3 simoleons. Elsewhere this might have been done 
+by creating `amounts` inline (i.e. `quatloosAmountMath.make(8)`). Or by
+creating `amounts` outside the proposal and assigning them to variables.
+For example, `const quatloos8 = quatloosAmountMath.make(8);` and then using
+`quatloos8` as the value for `Price` in the second clause above.
 
-The `want` term is an exact amount to exchang, while the
+The `want` term is an exact amount to exchange, while the
 `give` term is a limit that may be improved on. This simple exchange does not
 partially fill orders.
 
@@ -116,7 +122,7 @@ const bobSeat = await E(zoe).offer(
 
 ## Payout
 
-When a match is made, the payout promise in a user's seat
+When a match is made, the payout promise from a user's seat
 resolves to a promise for payment. For Bob:
 
 ```js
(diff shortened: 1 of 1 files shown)


Chris-Hibbert:
My opinion is that it's kind of okay to write our test code using the common shorthand we've developed (`moola(5)`, etc.), but our docs and probably sample contracts should start out using the `moolaAmountMath(3)` so it's clear what we're doing. In demos and presentations, as long as we mention that it's a shorthand for amountMath, the short forms are probably okay.

main/zoe/guide/contracts/simple-exchange.md
```diff
@@ -2,36 +2,36 @@
 
 <Zoe-Version/>
 
-##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/f29591519809dbadf19db0a26f38704d87429b89/packages/zoe/src/contracts/simpleExchange.js) (Last updated: 9/12/2020)
+##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/f29591519809dbadf19db0a26f38704d87429b89/packages/zoe/src/contracts/simpleExchange.js) (Last updated: 12-SEP-2020)
```
Chris-Hibbert, line 5: for dates, I actually prefer 2020-09-12. (You're right that I used the US-only format. shame on me.)  It seems to be an ISO standard, and it seems to be uniquely recognizable both US and internationally.

main/zoe/guide/contracts/simple-exchange.md
```diff
@@ -2,36 +2,36 @@
...
+the two keywords symmetrically. New offers are created and existing offers
+accepted in either direction.
 
 ```js
 { give: { Asset: simoleans(5) }, want: { Price: quatloos(3) } }
 { give: { Price: quatloos(8) }, want: { Asset: simoleans(3) } }
 ```
 
-The `want` term is treated as an exact amount to be exchanged, while the
+The `want` term is an exact amount to exchang, while the
```
Chris-Hibbert, line 28:
```suggestion
The `want` term is an exact amount to exchange, while the
```

main/zoe/guide/contracts/simple-exchange.md
```diff
@@ -2,36 +2,36 @@
...
 `give` term is a limit that may be improved on. This simple exchange does not
 partially fill orders.
 
-The publicFacet is returned when the contract is started.
+The `publicFacet` is returned when the contract is started.
```
Chris-Hibbert, line 32: Do we need 'the' with the highlighted term?
```suggestion
`publicFacet` is returned when the contract is started.
```

main/zoe/guide/contracts/simple-exchange.md
```diff
@@ -116,8 +116,8 @@ const bobSeat = await E(zoe).offer(
 
 ## Payout
 
-When a match is made, the payout promise that the user can get from their seat
-is resolved to a promises for payment.
+When a match is made, the payout promise in a user's seat
```
Chris-Hibbert, line 119:
```suggestion
When a match is made, the payout promise from a user's seat
```

main/zoe/guide/contracts/simple-exchange.md
```diff
@@ -127,8 +127,8 @@ const bobPricePayout = await bobPricePayoutP;
 const bobSimGainAmount = await E(bobSimPurse).deposit(bobPricePayout);
 ```
 
-And Alice gets her payouts the same way. (The choice of getPayouts() vs
-getPayout(keyword) is purely for convenience).
+Alice gets her payouts the same way. (The choice of `getPayouts()` vs
+`getPayout(keyword)` is purely for convenience).
```
Chris-Hibbert, line 131: This didn't sound smooth when I wrote it, and now that I'm seeing it again, I want to improve it.
```suggestion
Alice gets her payouts the same way. (The choice between `getPayouts()` and
`getPayout(keyword)` is based on which is more convenient in each circumstance).
```

main/zoe/guide/contracts/simple-exchange.md
tyg: I added the "the" to emphasize the "what it is" that this is in addition to "name of object"  here. By emphasizing that "public facet" is a thing and not just a name, it does a small bit of encouragment to look it up and find out what a public facet is if one doesn't know.

main/zoe/guide/contracts/simple-exchange.md
tyg: Makes sense. Used 12-SEP-2020 because that's what's currently on the site wide variable (changed from our previous 9/12/2020 style, so it wasn't just you). I can argue for why day-month is usually the most relevant info first in these cases, but ISO standard beats that.

katelynsills (approved):
LGTM!

PR #285, opened 2020-09-14
docs: update barter link

Chris-Hibbert:
And add link for barter to README.


katelynsills (approved):
Looks good

tyg:
Could you add a line or two that differentiates this a bit more from the various swap contracts? I'm not clear on what's interesting/useful/unique about doing Barter Exchange in particular, and saying what/why would be good. Otherwise looks good.

tyg (commented):
Good addition. Approve to merge.

PR #294, opened 2020-09-18
Werent offerHandles retired an seats took their place?

zarutian:
Was using https://agoric.com/documentation/zoe/api/zoe-contract-facet when writing a smart contract and stumbled upon this small inconsistency.

Small fix.


erights (approved):
LGTM. Thanks!

PR #299, opened 2020-09-22
chore: turn on setting that adds last updated to all files

katelynsills:
chore: turn on setting that adds last updated to all files, based on git data for that file

### DIFF
--- a/main/.vuepress/config.js
+++ b/main/.vuepress/config.js
@@ -44,6 +44,7 @@ module.exports = {
 
   /* --- DEFAULT THEME CONFIG --- */
   themeConfig: {
+    lastUpdated: 'Last Updated', 
     logo: '/logo.svg',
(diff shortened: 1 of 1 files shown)


tyg:
Looks good. Is there an easy way of adding the current Zoe/ERTP/Agoric version as well, although that would have to possibly be potentially overridden by tips (i.e. "This doc has not yet been updated to the version seen above...")

tyg (approved):

katelynsills:
> Looks good. Is there an easy way of adding the current Zoe/ERTP/Agoric version as well, although that would have to possibly be potentially overridden by tips (i.e. "This doc has not yet been updated to the version seen above...")

We'd have to do editing of the functionality, I think, to add the version. That'd be a lot harder than just changing the config. I think the ZoeVersion component is our best bet currently.

PR #302, opened 2020-09-23
Add plugin to check internal links

katelynsills:
Links can be checked with `npm run check-links`, giving output below. Some may be false positives. Eventually, it would be nice to have this be part of CI before merging.

Screen Shot 2020-09-22 at 4 59 42 PM
Screen Shot 2020-09-22 at 4 59 33 PM

### DIFF
--- a/main/.vuepress/config.js
+++ b/main/.vuepress/config.js
@@ -42,6 +42,8 @@ module.exports = {
     ]
   ],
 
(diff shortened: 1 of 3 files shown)


Chris-Hibbert (approved):
I don't know anything useful about configuring vuepress. The format here seems consistent, and I can see that the dependencies all stack up.

PR #318, opened 2020-09-25
chore: update to latest vuepress

katelynsills:


katelynsills:
The Vuepress issue I filed points to this PR, so let's keep it up for a bit.

PR #322, opened 2020-09-29
Spellcheck should ignore code sections

### LINKED ISSUE #321: Spellcheck should ignore code sections
There's an `ignores` configuration setting that is set to `code` and `pre` currently. The spellchecker is currently running on the input markdown, but if I change it to run on the output html, it might ignore those sections successfully. I'll make an issue.

_Originally posted by @katelynsills in https://github.com/Agoric/documentation/pull/311#discussion_r496314415_

katelynsills:
This PR changes the spellcheck CI step to be part of the build testing step. Instead of running the spellcheck on the markdown files, we run it on the html files and ignore the `<code>` and `<pre>` tags. This should reduce the number of new words we must add to the wordlist. 

I will also change the required CI actions as part of the github settings now that spellcheck step isn't a separate action.

Fixes #321


Chris-Hibbert (approved):
good idea.

PR #324, opened 2020-09-29
chore: add scheduled cron job to check all links (external and internal)

katelynsills:
This PR adds a scheduled cron job to check all links (external and internal). This should run every day, if I'm reading the chron notation correctly. 

This uses the same config as Tendermint: https://github.com/tendermint/tendermint/blob/master/.github/workflows/linkchecker.yml


.github/workflows/test-all-links.yml
```diff
@@ -0,0 +1,12 @@
...
+  schedule:
+    - cron: '* */24 * * *'
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: gaurav-nelson/github-action-markdown-link-check@1.0.7
+        with:
+          folder-path: "main"
```
Chris-Hibbert, line 12: Is this happy without a final linefeed? It's usually safer to include it.

Chris-Hibbert (approved):

.github/workflows/test-all-links.yml
katelynsills: Oh good point, I will include

PR #326, opened 2020-09-30
chore: try other link checker

katelynsills:


katelynsills:
Appears to work but needs a lot of configuration. Will revisit.

PR #332, opened 2020-10-01
fix: fix all links except for router-links

### LINKED ISSUE #331: Replace `agoric.com/documentation` with `/`
We shouldn't be using the `https://agoric.com/documentation/` prefix for any linking between our documentation pages. There are a number of reasons: 1) our link checkers can check internal link consistency without actually hitting the page, so it makes our link checkers work harder for no reason, and 2) if the website domain ever changes it will be much harder to change everything.

katelynsills:
Fix all links except for router-links, as checked by `yarn check-links` and https://github.com/tcort/markdown-link-check with the following command:

```
find . -path ./node_modules -prune -o -name \*.md -exec markdown-link-check {} -v -c markdown-link-check.json \;   
```

Also removes manual-setup which was based on a repo that no longer exists (cosmic-swingset) so the instructions are definitely out of date. 

Adds `yarn check-links` to CI. Removes the other link checker (https://github.com/tcort/markdown-link-check) from CI because the settings are not quite right yet. However, it works locally with the above command.

Closes #333 
Closes #331


Chris-Hibbert (commented):
This looks good. I didn't scrutinize every single link, but I did run it locally, and verify quite a few.

PR #334, opened 2020-10-01
fix: change router-links to regular markdown. 

### LINKED ISSUE #330: Replace explicit <router-link> with traditional Markdown links
We should remove the `router-link` references and replace them with traditional Markdown links so that our link-checkers can operate on them.

katelynsills:
Also, fix new issues identified by the link checker now that the links are in a checkable format.

Closes #330 

Stacked on top of #332


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules/
 dist/
 .DS_Store
 _agstate
+mlc-config.json
```
Chris-Hibbert, line 5: needs a trailing newline.

main/ertp/guide/amounts.md
```diff
@@ -109,14 +109,14 @@ return a `brand`.
 
 Values are the "how many" part of an `amount`. There are no `value`
 methods, but two `amountMath` methods use or return them. 
-- <router-link to="../api/amount-math.html#amountmath-getvalue-amount">`amountMath.getValue(amount)`</router-link>
+- [`amountMath.getValue(amount)`](../api/amount-math.md#amountmath-getvalue-amount)
```
Chris-Hibbert, line 112: In several places (e.g. line 76 above) I see changes from relative path (../) to absolute (/ertp/) paths but here and below, they're left as relative. Is the change to absolute paths important or not? They seem to produce identical behavior in the generated pages.

Chris-Hibbert (approved):
Hmm.  I don't see the point of these changes. I've compared the current site with a locally generated one, and I don't see any differences in displayed text, links displayed when I hover, or destinations. The router-links were apparently more work and more verbose, but converting `.html` to `.md` on links or adding `.md` where it was missing before, or converting from relative (`../..`) to absolute (leading `/`) paths doesn't seem to make any difference to the site. 

What's the benefit of spending time on making these changes?

tyg:
> converting from relative (`../..`) to absolute (leading `/`) paths doesn't seem to make any difference to the site.

That may not make a difference to the site, but it definitely makes it quicker and easier to write links if we make it our standard. Being able to always link to a Glossary entry by going (/Glossary/README.md#foo) is a *lot* quicker/easier to do than having to figure out the relative path from, say, a file at /ERTP/api/foo/bar.md back to there, at least for me.

katelynsills:
> Hmm. I don't see the point of these changes. I've compared the current site with a locally generated one, and I don't see any differences in displayed text, links displayed when I hover, or destinations. The router-links were apparently more work and more verbose, but converting `.html` to `.md` on links or adding `.md` where it was missing before, or converting from relative (`../..`) to absolute (leading `/`) paths doesn't seem to make any difference to the site.
> 
> What's the benefit of spending time on making these changes?

The benefit is being conformant with the automated link checkers as mentioned in #330. Many of these links that I changed were simply wrong, connecting to ids that didn't exist anymore because method signatures have changed. We didn't know they were wrong until the automated link checker told me. You can think of it like eslint or prettier. Getting conformant is a pain, sure, but afterward it's great to keep you from making mistakes. Btw, I did the replacements with a regex so the time spent was minimal.

main/ertp/guide/amounts.md
katelynsills: Some of these were simply wrong, and it was easier to get them right when the path is absolute.

PR #341, opened 2020-10-05
Update docs (with questions) for SDK PR Vat shutdown patch 1798

tyg:
Per Chris' request, updated docs to account for agoric-sdk changes in PR 1798. There are some questions that need answering for me to make the final small additions. See also [Agoric/agoric-sdk#1798](https://github.com/Agoric/agoric-sdk/pull/1798)

### DIFF
--- a/main/zoe/api/zoe-contract-facet.md
+++ b/main/zoe/api/zoe-contract-facet.md
@@ -259,11 +259,19 @@ to manipulate the offer. The queries and operations are as follows:
         Price: moolaAmountMath.make(9)
       }
       ```
- - `exit()`
+ - `exit(completion)`
    - Returns: `void`
    - Causes the `seat` to exit, concluding its existence. All `payouts`, if any,
-     are made, and the `seat` object can no longer interact with the contract. .
- - `kickOut(msg)` (`msg` is optional)
+     are made, and the `seat` object can no longer interact with the contract.
+     The `completion` argument is usually a string, but this is not required. **tyg todo:
+     What else can it be and what is it used for/does it do?**
+ - `fail(msg)` (`msg` is optional)
+   - Returns: `void`
+   - The `seat` exits, displaying the `msg` string, if there is one, on the console.
+     This is equivalent to exiting, except that `exit` is for a successful transaction while
+     `fail()` aborts the transaction attempt and signals an error. The contract
+     still gets its current `allocation` and the `seat` can no longer interact with the contract.     
+ - `kickOut(msg)` (`msg` is optional) **Renamed fail(msg) as of 4-OCT-2020. DO NOT USE**
    - Returns: `void`
    - The `seat` exits, displaying the `msg` string, if there is one, on the console.
      This is equivalent to exiting, except that `exit` is for a successful transaction while
@@ -313,9 +321,14 @@ Returns the `amountMath` object associated with the `brand` argument.
 ```js
 const assetAmountMath = zcf.getAmountMath(assetAmount.brand);
 ```
-## zcf.shutdown()
+## zcf.stopAcceptingOffers()
+- The contract requests Zoe to not accept offers for this contract instance. **tyg todo: No way of 
+starting it back up and accepting again?** It can't be called from outside the contract 
+unless the contract explicitly makes it accessible.
 
-**Note**: Still in development, use at your own risk.
+## zcf.shutdown(completion)
+
+**Note**: Still in development, use at your own risk. **tyg todo: Is this warning still valid?**
 
 Shuts down the entire vat and gives payouts.
 
@@ -325,6 +338,9 @@ giving them their payouts.
 Call when:
 - You want nothing more to happen in the contract, and
 - You don't want to take any more offers
+
+The `completion` argument is usually a string, but this 
+is not required. **tyg todo: What else can it be and what is it used for/does it do?**
 
 ```js
 zcf.shutdown();
--- a/main/zoe/api/zoe-contract-facet.md
+++ b/main/zoe/api/zoe-contract-facet.md
@@ -265,14 +265,14 @@ to manipulate the offer. The queries and operations are as follows:
      are made, and the `seat` object can no longer interact with the contract.
      The `completion` argument is usually a string, but this is not required. Its
      only use is for the notification sent to the contract instance's `done()` function. 
-     Any still open seats or other outstanding promises are closed with a generic 'vat terminated' 
-     message.
+     Any other still open seats or outstanding promises and the contract instance continue.
  - `fail(msg)`
    - Returns: `void`
    - The `seat` exits, displaying the optional `msg` string, if there is one, on the console.
      This is equivalent to exiting, except that `exit` is successful while
      `fail()` signals an error occured while processing the offer. The contract
-     still gets its current `allocation` and the `seat` can no longer interact with the contract. 
+     still gets its current `allocation` and the `seat` can no longer interact with the contract.
+     Any other still open seats or outstanding promises and the contract instance continue.
  - `kickOut(msg)` **Renamed fail(msg) as of 4-OCT-2020. DO NOT USE**
    - Returns: `void`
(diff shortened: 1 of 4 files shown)


main/zoe/api/zoe-contract-facet.md
```diff
@@ -259,11 +259,19 @@ to manipulate the offer. The queries and operations are as follows:
...
       ```
- - `exit()`
+ - `exit(completion)`
    - Returns: `void`
    - Causes the `seat` to exit, concluding its existence. All `payouts`, if any,
-     are made, and the `seat` object can no longer interact with the contract. .
- - `kickOut(msg)` (`msg` is optional)
+     are made, and the `seat` object can no longer interact with the contract.
+     The `completion` argument is usually a string, but this is not required. **tyg todo:
+     What else can it be and what is it used for/does it do?**
```
Chris-Hibbert, line 267: Its only use is that it shows up in the seat's notifier's final state.

main/zoe/api/zoe-contract-facet.md
```diff
@@ -259,11 +259,19 @@ to manipulate the offer. The queries and operations are as follows:
...
+ - `fail(msg)` (`msg` is optional)
+   - Returns: `void`
+   - The `seat` exits, displaying the `msg` string, if there is one, on the console.
+     This is equivalent to exiting, except that `exit` is for a successful transaction while
+     `fail()` aborts the transaction attempt and signals an error. The contract
```
Chris-Hibbert, line 272:
```suggestion
     `fail()` signals an error. The contract
```

The only difference is the way it signals termination. In both cases, other functions called in the same turn will complete. in both cases the seat gets its final allocation, and in both cases there is no further interaction with the contract.

main/zoe/api/zoe-contract-facet.md
```diff
@@ -313,9 +321,14 @@ Returns the `amountMath` object associated with the `brand` argument.
 ```js
 const assetAmountMath = zcf.getAmountMath(assetAmount.brand);
 ```
-## zcf.shutdown()
+## zcf.stopAcceptingOffers()
+- The contract requests Zoe to not accept offers for this contract instance. **tyg todo: No way of 
+starting it back up and accepting again?** It can't be called from outside the contract 
+unless the contract explicitly makes it accessible.
```
Chris-Hibbert, line 327: This hasn't been added yet.  I suppose I should add this now.

main/zoe/api/zoe-contract-facet.md
```diff
@@ -313,9 +321,14 @@ Returns the `amountMath` object associated with the `brand` argument.
...
 
-**Note**: Still in development, use at your own risk.
+## zcf.shutdown(completion)
+
+**Note**: Still in development, use at your own risk. **tyg todo: Is this warning still valid?**
```
Chris-Hibbert, line 331: This warning should be dropped. Chip's PR has finished the implementation, so this now does everything it's intended to.

main/zoe/api/zoe-contract-facet.md
```diff
@@ -326,6 +339,9 @@ Call when:
 - You want nothing more to happen in the contract, and
 - You don't want to take any more offers
 
+The `completion` argument is usually a string, but this 
+is not required. **tyg todo: What else can it be and what is it used for/does it do?**
```
Chris-Hibbert, line 343: As above, it is used for the notification sent to the `done()` function for the contract instance. Any seats that are still open or other promises outstanding are closed with a generic 'vat terminated' message.

main/zoe/api/zoe.md
```diff
@@ -193,11 +194,27 @@ be different for different instances of the same contract, but the contract
...
 It returns a promise for a `StartInstanceResult` object. The object consists of:
+- `adminFacet` `{any}`
 - `creatorFacet` `{any}`
 - `publicFacet` `{any}`
 - `instance` `{Instance}`
 - `creatorInvitation` `{Payment | undefined}`
 
+The `adminFacet` has two methods:
+- `getVatShutdownPromise()
+  - Returns a promise that resolves (to reason or completion) when this newly started instance terminates. **tyg todo: Not sure what "resolves to reason" does? Thrown error?**
```
Chris-Hibbert, line 205: 'reason' is the conventional name for the Error generated with `fail(reason)`, while `completion` is the name passed with `exit(completion)`.

Chris-Hibbert (approved):
Still on hold, pending decision to release.

Chris-Hibbert:
Oh, please add "see also" link to https://github.com/Agoric/agoric-sdk/pull/1798 in the top comment. github can use that to make cross-links.

main/zoe/api/zoe-contract-facet.md
katelynsills: Yup, I agree with Chris. We need to take out the term `transaction` - it implies that there's such a thing, but nothing like it actually exists. How about:

```suggestion
     
     This is equivalent to exiting, except that `exit` is successful while `fail()` signals an error occurred while processing the offer.
```

main/zoe/api/zoe-contract-facet.md
katelynsills: Let's just table this particular documentation change into a later PR rather than change your PR, @Chris-Hibbert

katelynsills (changes requested):
I agree with Chris' feedback but otherwise looks good!

main/zoe/api/zoe-contract-facet.md
Chris-Hibbert: Nah, it's okay to include this. I've prepared a [PR that adds `stopAcceptingOffers()`](https://github.com/Agoric/agoric-sdk/pull/1841). 

Since Dean has requested that we coordinate before releasing these changes, we might as well have all the related changes ready at the same time.  (And Dean is the one who requested `stopAcceptingOffers()`.)

main/zoe/api/zoe-contract-facet.md
```diff
@@ -259,13 +259,23 @@ to manipulate the offer. The queries and operations are as follows:
...
+ - `exit(completion)`
    - Returns: `void`
    - Causes the `seat` to exit, concluding its existence. All `payouts`, if any,
-     are made, and the `seat` object can no longer interact with the contract. .
- - `kickOut(msg)` (`msg` is optional)
+     are made, and the `seat` object can no longer interact with the contract.
+     The `completion` argument is usually a string, but this is not required. Its
+     only use is for the notification sent to the contract instance's `done()` function. 
+     Any still open seats or other outstanding promises are closed with a generic 'vat terminated' 
+     message.
```
Chris-Hibbert, line 269: Sorry, I wasn't clear. 

This applies to `shutdown()`, but not to `exit()` or `fail()`. When you exit/fail an individual seat, that seat is done, but the contract continues, and any other seats continue. When you shutdown the vat & contract instance, any remaining seats have `fail()` called. Normally, one wouldn't call shutdown unless all the seats had already completed their work.

main/zoe/api/zoe.md
```diff
@@ -193,11 +194,28 @@ be different for different instances of the same contract, but the contract
...
+- `adminFacet` `{any}`
 - `creatorFacet` `{any}`
 - `publicFacet` `{any}`
 - `instance` `{Instance}`
 - `creatorInvitation` `{Payment | undefined}`
 
+The `adminFacet` has two methods:
+- `getVatShutdownPromise()
+  - Returns a promise that resolves to reason (the Error generated with `fail(reason)`) or 
+    completion (the name passed with `exit(completion)`) when this newly started instance terminates.
```
Chris-Hibbert, line 206:
```suggestion
  - Returns a promise that resolves to reason (the value passed to `fail(reason)`) or 
    completion (the value passed to `exit(completion)`) when this newly started instance terminates. 
```

Chris-Hibbert (changes requested):
My approval was pre-mature.

Chris-Hibbert (approved):

katelynsills (approved):
LGTM!

PR #342, opened 2020-10-05
Created (and rewrote) single file for how to contribute to the Documentation repo

tyg:
Combined (and added to) the various sources for our documentation process and how to work with the Documentation repo.


katelynsills:
Can you also delete the other file in this PR as well?

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+open in the same browser tab when clicked on, absolute links open a new tab. 
+
+Use this trick to make writing relative links easier. While it's easy enough to, say, write a relative link to something in
+the same folder as the file you're writing (something like `(./assets/my-diagram.svg)` to include an image), it can be
+tricky to remember/figure out what the right syntax is for linking to a file two folders up, in a different upper folder, and then
+two levels down from there on a different branch of the file structure. 
+
+Instead, `main` is considered the top of the file hierarchy. So you can always get to, say, a Glossary entry 
+by just linking to `(/glossary/#allocation)`; its path relative to `main`. Any path starting with just `/` is considered
+to start at `main`.
```
Chris-Hibbert, line 86: If this means that paths that start with `'/'` are considered to be relative paths, please say so explicitly. My unix-based instincts read paths that start with `'/'` as absolute.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+
+VuePress turns every header in a Markdown file to a named file location you can link to so clicking take you directly to
```
Chris-Hibbert, line 88:
```suggestion
VuePress turns every header in a Markdown file to an HTML anchor  you can link to so clicking take you directly to
```

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+that file location. At the end of the link, append a `#` to the file name, followed by the header text. The header text
+must be altered to be 1) all lower case. 2) All punctuation and spaces are replaced by hyphens (except there aren't any
```
Chris-Hibbert, line 90: Please correct this if I'm wrong about "punctuation and spaces" meaning "non-alphanumerics". I don't know what's included in "punctuation". Are multiple consecutive non-alphanumerics turned into a single dash or one-per-character?
```suggestion
must be altered to be 1) all lower case. 2) All non-alphanumerics are replaced by hyphens (except there aren't any
```

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+`js` appended and ending with a line consisting of three backquotes. 
 
+Rather, code snippets are actual development or test code from the Agoric-SDK repo,
+or code held to a similar standard of correctness. They should pass `lint` and run with no
+errors. This ensures our documents use real code that works
+with the current agoric-sdk version (whatever is on master) and is
+not outdated. **tyg todo: I'm not quite seeing this guaranteeing things,
+since we're not extracting the code directly from agoric-sdk. It'd seem
+if we don't check and manually update the snippets file after each upgrade, we risk
+falling out of sync with the agoric-sdk repo.**
```
Chris-Hibbert, line 138: you might say it "provides assurance" to use instead of "guarantees", for the reasons you cite.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+
+To import code snippets into the documentation, create a file
+under `Agoric/documentation/snippets/`. **tyg todo: It appears there should be only one 
+snippets folder for the whole doc repo?** Put your code in the new file.
```
Chris-Hibbert, line 142: I think @katelynsills  said one per directory.

Chris-Hibbert (commented):
I ran out of time. I'll take another look tomorrow.

CONTRIBUTING.md
katelynsills: There's only one snippets folder, but with multiple subfolders underneath in a similar structure to the main docs structure. This is similar to having one test directory.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+Remember, it must be able to run without errors.
```
katelynsills, line 143: We can do more than that. AVA is imported, so you can write specific tests as well. For instance if your snippet says:
```
// issuer === quatloosIssuer
```
you can write the test:

```
t.is(issuer, quatloosIssuer);
```

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
-  // #region install
-  const atomicSwapInstallation = await E(zoe).install(atomicSwapBundle);
-  // #endregion install
+// #region regionName
+...
+// #endregion regionName
 ```
+`regionName` in the above is any name you want to give this snippet.
+
+If you want the whole file to be a snippet, just put the `#region` / `#endregion`
```
katelynsills, line 155: The whole file will be imported if there's no region mentioned in the import or if the region doesn't exist (dangerous!), so no need to put it at top and bottom - just omit from the import.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+pair at the start and end of the file. You can define any number of snippet
+regions in a file, including defining one region inside of another. Just be sure to give
```
katelynsills, line 157: I don't know if you can define a region inside another. I haven't tried it.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+the file. 
 
-Write tests using AVA, and then run them with `yarn test`. The code
-files can also be linted with `yarn lint-fix`. Testing code snippets
-allows us to ensure that our documents are using real code that works
-with the current version of agoric-sdk (whatever is on master) and is
-not outdated.
+To test your snippets files:
+1. Write tests using AVA.
+2. Run the tests with `yarn test`. **tyg todo: (Do all the tests have to be in the snippets directory? And do you run `yarn test` from there?)**
```
katelynsills, line 168: The tests should be in the snippets directory. Actually the snippet files and the test files ended up being the same file because that was the cleanest way to go. The snippet files should all be named `test-original-filename.js` where `original-filename` is the markdown filename, like `amounts` of `ertp/guide/amounts.md`

Then, running `yarn test` anywhere will run the tests.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+3. Lint the files with `yarn lint-fix` **tyg todo: See above re: what directory you run this from**
```
katelynsills, line 169: anywhere, but usually root of the repo

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
 
-## Check Links
+**tyg todo: If they're being tested and linted during the merge to master, is this just how you personally test things during 
+development?**
```
katelynsills, line 172: yup!

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
 
-To check internal Vuepress links locally, run the following shell command. It does *not* check either external links or router-links. Output consists of the text of any broken links, what file they're in, and what line number they occur on.
+### Check Links
 
+To check internal VuePress links locally, run the shell command `yarn check-links` **tyg todo: From where?**
```
katelynsills, line 176: You can run it from anywhere in the root of the repo folder or below

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+    link: '/glossary/'
+  },
+```
+
+This is an abridged configuration of the `ERTP` entry and its submenu.
+Each submenu item has the same structure and properties as a top menubar item,
+except they do not have an `items` property. Note that for links to files, instead of folders, such as
+the ERTP Introduction, you leave off the file's suffix. So in that case, the
+link value is `'/getting-started/ertp-introduction/'` and not `'/getting-started/ertp-introduction.md'`
+All links must end with a trailing `/`, even if the link is to a file.
```
katelynsills, line 293: I didn't understand this. I think we should be able to link to an .md no matter what. It's just that in some cases the markdown file is README.md

katelynsills:
```
All links must end with a trailing `/`, even if the link is to a file.
```
I don't think this is right

katelynsills (changes requested):
I think a couple things aren't quite right, but this is overall looking good!

CONTRIBUTING.md
tyg: That's a good question about multiple consecutive non-alphanumerics. I'll have to do some experimenting.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+
+### README files
+
+VuePress converts Markdown files to same named HTML file. `README.md` files are an exception; they're 
+renamed `index.html`, since that's how web servers find a website's root file. Navigating 
+to `https://agoric.com/documentation/ertp/guide/` displays the VuePress processed `/main/ertp/guide/README.md`.
+While it may seem odd, VuePress expects multiple `README.md` files in a repo, one each for most folders in it.
+
+The root README.md file must start with a header. Any of H1, H2, or H3 (`#`, `##`, or `###` in Markdown) will do.
+This is needed to generate search indexes and sidebars.
```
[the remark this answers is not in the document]
tyg: VuePress tutorial: "The root README.md file must start with a header. It doesn't have to be an H1 (# as shown above) but it should be either H2 (##) or H3 (###) for proper generation of search indexes and sidebars."

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+
+A subdirectory is invisible to VuePress unless it has a `README.md`. Empty `README.md` files are acceptable.
+
```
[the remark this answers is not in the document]
tyg: Also from the VuePress tutorial: "Another property of README.md as mentioned before is that if you want to organize your site using subdirectories, a subdirectory is invisible to VuePress unless it has a README.md in it. Those README.md files can be blank."

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+VuePress turns Markdown links into HTML links. There are some quirks about how you should write Markdown links.
+
+First, our link checker does **not** check `router-link` style links. Please only use Markdown style links.
+
+Next, your Markdown links should be to the `.md` Markdown files in the Doc repo. VuePress processing will change
+both the `.md` files and links to them to be `.html`.
+
+Use relative links instead of absolute ones for any links to files or folders in the Documentation repo. Relative links
+open in the same browser tab when clicked on, absolute links open a new tab. 
+
```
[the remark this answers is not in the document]
tyg: I've written around it. I refer to the main (/...) start as a way around writing complicated relative links, wrapping up with say this type of link also opens in the same browser tab. I never actually call it a relative or absolute link.

CONTRIBUTING.md
```diff
@@ -1,75 +1,389 @@
...
+VuePress turns every header in a Markdown file to a named file location you can link to so clicking take you directly to
+that file location. At the end of the link, append a `#` to the file name, followed by the header text. The header text
+must be altered to be 1) all lower case. 2) All punctuation and spaces are replaced by hyphens (except there aren't any
+trailing hyphens at the end)
+
+So, for example, the header `E(zoe).getBrands(instance)` is linked to by `zoe.md#e-zoe-getbrands-instance` (note the last `)` was
+not turned into a hyphen) and the header
+`Contract and Contract Instance` by `glossary/#contract-and-contract-instance`. Note that no specific file is specified in
+the latter, as it defaults to the `glossary` folder's `README.md` file
```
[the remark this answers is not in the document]
tyg: Dunno about the link checker, but if you go to the site's Glossary and click on copy link address for one of the entry slugified links, you get https://agoric.com/documentation/glossary/#allocation

CONTRIBUTING.md
tyg: Not sure if I'm following you, since I've not used or looked at AVA. Does this mean a single file test-original-filename.js both has AVA test(s) and code snippet(s) [the latter denoted by #region and #endregion markers] in it? With the tests specifically testing the code snippet and that snippet only?

katelynsills: Yup, you can take a look at https://github.com/Agoric/documentation/blob/6763ee64d5fc66c80a7e3fb4c7a680c41db3a73e/snippets/ertp/guide/test-amounts.js and the other files on that branch to see how it's organized.

CONTRIBUTING.md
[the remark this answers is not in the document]
katelynsills: Ok, can we back up a bit? Is there any reason why someone wouldn't use an h1 header on a page in our documentation? I think we should not say from our perspective that they are all fine, when it's very likely that not using an h1 header is the wrong thing.

CONTRIBUTING.md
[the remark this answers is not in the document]
katelynsills: I just tested, and it is false. To try it yourself, add a new directory, then add a file in that directory with a name other than `README.md`. Then navigate to the produced html file in your local browser. 
Screen Shot 2020-10-06 at 6 36 28 PM

CONTRIBUTING.md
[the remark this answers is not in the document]
katelynsills: Ah, gotcha. We might also want to adopt the language here: https://mor10.com/html-basics-hyperlink-syntax-absolute-relative-and-root-relative/

CONTRIBUTING.md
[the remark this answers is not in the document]
katelynsills: Ah, good point. I see in our current documentation we use this and it passed the link checkers, so I think we're good.

CONTRIBUTING.md
```diff
@@ -1,75 +1,404 @@
...
 ```
+3. **Build**: To build the site as it will be built for production, run:
+```shell
+yarn docs:build
+```
+4. **Run**: To run a local server and see your changes in real time, run:
+```shell
+yarn docs:dev
+```
+Note that site config changes may require stopping and restarting this program.
```
Chris-Hibbert, line 215:
```suggestion
Most edit changes will be reflected immediately in the browser, but 
note that site config changes may require stopping and restarting this program.
```

CONTRIBUTING.md
katelynsills: Just tried it, it seems to work fine

CONTRIBUTING.md
```diff
@@ -1,75 +1,420 @@
...
+
+Instead, VuePress considers `main` the top of the file hierarchy. So you can always get to, say, a Glossary entry 
+by just linking to `(/glossary/#allocation)`; its path starting at `main`. Any path starting with just `/` starts
+at `main`. These links also open in the same browser tab. 
+
+VuePress turns every header in a Markdown file to an HTML anchor you can link to, so clicking such a link takes you directly to
+that file location (called *slugifying* by WordPress and other blogging platforms). At the end of the link, 
+append a `#` to the file name, followed by the header text. The header text
+must be altered to be 1) all lower case. 2) All non-alphanumerics, including spaces, are replaced by hyphens (except there aren't any
+trailing hyphens at the end).
```
Chris-Hibbert, line 105: the preface to the list doesn't apply equally to all elements. Please re-word for parallel construction or drop the numbering.

CONTRIBUTING.md
```diff
@@ -1,75 +1,420 @@
...
+should be stored in `main/zoe/guide/assets/process.svg` and appear in the page via an image link of `./assets/process.svg` 
+
+Note that `assets` should store all the auxiliary files for the files in its parent folder. Don't make an `assets` folder
+or similar for individual files/pages.
+
+### README files
+
+VuePress converts Markdown files to an HTML file with the same base name. `README.md` files are an exception; they're 
+renamed `index.html`, since that's the default file web servers expect to find in each directory. Navigating 
+
```
Chris-Hibbert, line 61: This blank line will turn into a paragraph break, right?

Chris-Hibbert (approved):
minor cleanups

CONTRIBUTING.md
tyg: Doesn't seem to be right. Fixed.

tyg: I think you're right.

katelynsills (approved):
LGTM!

Issue #297, opened 2020-09-20
starting a project: need yarn before yarn start in ui?


dckc:
on https://agoric.com/documentation/getting-started/start-a-project.html at step 8 `(cd ui && yarn start)` I get `/bin/sh: 1: parcel: not found`

`npm ls` gripes about `UNMET DEPENDENCY parcel-bundler@^1.12.4`. `yarn ls` says there is no `ls` thingy

the `README.md` here in the `ui` directory says to do `yarn` then `yarn start`. That seems to work.

katelynsills:
related: https://github.com/Agoric/dapp-encouragement/issues/63

PR #346, opened 2020-10-13
Install parcel in /ui

katelynsills:
Add instruction to do `yarn install` in /ui to install parcel

Closes https://github.com/Agoric/dapp-encouragement/issues/63
Closes https://github.com/Agoric/documentation/issues/297

### DIFF
--- a/main/getting-started/start-a-project.md
+++ b/main/getting-started/start-a-project.md
@@ -94,7 +94,7 @@ Cycle](/getting-started/development-cycle.md)
   </tr>
   <tr>
     <td>8</td>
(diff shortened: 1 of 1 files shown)


michaelfig:
> Add instruction to do `yarn install` in /ui to install parcel

I'd prefer for `agoric install` to do this, and in fact, that's necessary if the `/ui` directory is ever going to use any packages from `agoric-sdk` so that they are linked.

Please see: Agoric/agoric-sdk#1869

katelynsills:
Closed in favor of better solution

PR #359, opened 2020-10-27
Glossary patch 09 2020

tyg:
Added lots of new entries and edited old ones.


main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
 see the [ERTP Introduction](/getting-started/ertp-introduction.md),
 [ERTP Guide](/ertp/guide/), and [ERTP API](/ertp/api/).
 
+## Escrow
+
+To give assets for a possible transaction to an impartial third party, who keeps them until specified conditions are satisfied. 
+For example, Alice wants to sell Bob a ticket for $100. Alice escrows the ticket, and Bob escrows the $100, with Carol. Carol
+does not give Alice the $100 or Bob the ticket until she has both items. Since neither Alice nor Bob ever holds both items at
+once, they don't have to trust each other to do the transaction. Zoe automatically escrows payments for transaction offers.
+
```
katelynsills, line 166: I don't think we should define escrow in a general way with a untrusted third-party such as Carol, when users should use Zoe for escrow.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+## Eventual Send
+
+See [`E()`](#e) above.
+
+## Exit Rule
+
+Part of an [offer](#offer) specifying how the offer can be cancelled/exited. There are three values:
+- `onDemand:null`: (Default) The offering party can cancel on demand.
```
katelynsills, line 174: we should put a space here
```suggestion
- `onDemand: null`: (Default) The offering party can cancel on demand.
```

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+- `waived:null`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
```
katelynsills, line 175:
```suggestion
- `waived: null`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
```

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+- `afterDeadline`: The offer is automatically cancelled after a deadline, as determined by its timer and deadline properties.
+
 ## Facet
 
-A particular view or API of an object. An object can have many facets.
+A particular view or API of an object. An object can have many facets. Two Agoric uses are:
```
katelynsills, line 180: I realize this part didn't change, but I think it might be confusing to a new user. Instead maybe (needs work):

A facet is an object that exposes a certain API or particular view of some larger entity, which may be an object itself. Any number of facets of the larger entity can be made. In JavaScript, a facet of an object is often made by selecting the methods from the old object, either directly or by destructuring:

```js
const facet = {
  myMethod: oldObject.method,
```

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+## IBC
+The Inter-Blockchain Communication protocol, used to by blockchains to communicate with each other. A short article about IBC
+is available [here](https://www.computerweekly.com/blog/Open-Source-Insider/What-developers-need-to-know-about-inter-blockchain-communication).
+
+## Invitation
+To participate in a contract instance, one must hold an invitation to do so. Contracts often 
+return a creator invitation on their instantiation, in case the contract instantiator wants 
+to immediately participate. Otherwise, the contract instance must create any additional invitations. 
+These, or any invitation held by a party, can be distributed via any means the holder wishes. An
+invitation could be emailed directly to a friend, posted on a bulletin board, etc. When you receive
```
katelynsills, line 210: hmm, well, we don't have any way of emailing ERTP payments to a friend yet, so I think we shouldn't say that. The invitation can only be distributed one way: by sending it to someone else's wallet.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+## InvitationIssuer
+
+Since [invitations](#invitation) are special cases of [payments](#payment), invitations 
+must have a dedicated [issuer](#issuer), which is the InvitationIssuer.
+
+Zoe has a single `InvitationIssuer` for its entire lifetime. By having a reference to Zoe, 
+a user can get the `InvitationIssuer`. This lets them claim any invitation they 
+receive from someone else by calling `E(invitationIssuer).claim()` with the untrusted
+invitation as the argument. During the claiming process, the invitationIssuer validates
+the invitation. A successful claim also means that invitation is exclusively yours.
```
katelynsills, line 231: We should mention that depositing in a purse does the same thing, and this is what the wallet does.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
-Issuers are linked to a single [mint](#mint) and vice versa, so each issuer works
+Issuers are a one-to-one relationship with both a [mint](#mint) and a [brand](#brand), so each issuer works
 with one and only one asset type, such as only working with quatloos or only working
-with moola. They can create empty [purses](#purse) and [payments](#payment) for
-their asset type, but cannot mint new [amounts](#amounts). 
+with moola. This association cannot change to another type. 
 
-Issuers can also transform
+Issuers can create empty [purses](#purse) and [payments](#payment) for
+their asset type, but cannot mint new [amounts](#amount). Issuers can also transform
```
katelynsills, line 239: Nope, issuers cannot create empty payments.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
 
-## Vat
+A set of Zoe API methods made visible on a Zoe [facet](#facet). They can be called synchronously by contract code. 
+See the [ZCF API](/zoe/api/zoe-contract-facet.md).
 
-A vat is a *unit of synchrony*. This means that within a JavaScript vat, objects and functions can communicate with one another synchronously.
+## ZCFMint
 
-A vat runs a single *event loop*.
+See [Mint](#mint).
```
katelynsills, line 454: Let's define this separately from the ERTP mint.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+The overall place a party keeps their assets of all brands. For example, your wallet might contain 5 Quatloos
+[purses](#purse), 8 Moola purses, and 2 Simoleons purses. You can also keep [Issuers](#issuer) in a 
+wallet. [Offers](#offer) are associated with the wallets 
+from which their associated payments come from. See the [Wallet API](/wallet-api.md).
+
+## ZCF (Zoe Contract Facet)
 
-## Vat
+A set of Zoe API methods made visible on a Zoe [facet](#facet). They can be called synchronously by contract code. 
+See the [ZCF API](/zoe/api/zoe-contract-facet.md).
```
katelynsills, line 450:
```suggestion
The facet of Zoe that is exposed to contract code. The Zoe Contract Facet methods can be called synchronously by contract code. 
See the [ZCF API](/zoe/api/zoe-contract-facet.md).
```

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+episode [The Gamesters of Triskelion](https://en.wikipedia.org/wiki/The_Gamesters_of_Triskelion).
+
+## Reallocate/Reallocation
+
+When an [offer](#offer) exits due either to success or rejection, the associated [payments](#payment) that were [escrowed](#escrow) with Zoe are automatically
+appropriately reallocated to the offer participants. If the offer was rejected or otherwise failed, Zoe gives the offer-making party back what
+they escrowed. If the offer was accepted, Zoe allocates to each party what they said they wanted. There are cases where the reallocation both
+gives a party what they wanted and some of what they escrowed. For example, in an auction a party might have escrowed 10 Quatloos to make their
+highest bid if necessary, but they won the item with a bid of just 8 Quatloos. The reallocation would give them both the item they won and
+a refund of the 2 Quatloos that weren't needed to purchase it.
```
katelynsills, line 383: No, this is very confused. A reallocation is a transfer of amounts between seats within Zoe, a change in allocation. When a seat exits, the seat gets the allocation it currently has, as a payout. Exiting *never* triggers a reallocation. Exiting has nothing to do with reallocation. I think you are talking about allocations here.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+from one party to another. All assets of a payment must be of the same [brand](#brand).
+
+For more information, see the [ERTP Guide's Payments section](/ertp/guide/purses-and-payments.md#purses-and-payments)
+and the [ERTP API's Payments section](/ertp/api/payment.md).
 
-See the [ERTP Guide's Purses section](/ertp/guide/purses-and-payments.md#purses) and the
-[ERTP API's Purses section](/ertp/api/purse.md)
-for more information.
+## Payout
+The assets reallocated to a user when an [offer](#offer) exits, either successfully or not. If an offer isn't successful, a
```
katelynsills, line 333: No, not reallocated! 
```suggestion
The assets paid out to a user when an [offer](#offer) exits, either successfully or not. If an offer isn't successful, a
```

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+payout is a party's [escrowed](#escrow) assets. If an offer is successful, a party's payout is what they wanted, and possibly the 
+return of any escrowed assets in excess of the final cost. 
+
```
katelynsills, line 336: This is very confused. There's no concept of offer success, and the payout does not depend on it. The payout is always what the seat's current allocation is, regardless of how the seat exited (seat.exit() or seat.fail() or an error in the offerHandler). If there was a previous reallocation, it will be different than what the user escrowed. If there is no reallocation before the seat exits, it will be the same as what they escrowed.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+to be [escrowed](#escrow), and [payouts](#payout) to the user by serving as keys for key-value pairs in various 
+records with values of [amounts](#amount), [issuers](#issuer), etc.
 
 ## Mint
-In ERTP, mints create digital assets and are the only objects with the authority to do so. 
-Access to a mint gives you the power to create more digital assets of its type at will. Mints
-can only create one type of asset. 
+Agoric has two mint objects, *ERTP mints* and *Zoe Contract Facet mints (ZCFMints)*. They both create
+digital assets. Which mint type creates an asset doesn't matter; quatloos created by an ERTP mint are 
+indistinguishable from quatloos created by a ZCFMint.
```
katelynsills, line 257: I think this might cause more confusion than it clears up. Let's say that there is only one kind of mint, the ERTP mint, but ZCF provides a different interface, the ZCFMint interface, with an ERTP mint underneath. Both create ERTP assets.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
-In ERTP, mints create digital assets and are the only objects with the authority to do so. 
-Access to a mint gives you the power to create more digital assets of its type at will. Mints
-can only create one type of asset. 
+Agoric has two mint objects, *ERTP mints* and *Zoe Contract Facet mints (ZCFMints)*. They both create
+digital assets. Which mint type creates an asset doesn't matter; quatloos created by an ERTP mint are 
+indistinguishable from quatloos created by a ZCFMint. 
 
-Mints are [issuer's](#issuer) admin facets, and there is a one-to-one relationship between an issuer and
-its mint.
+- ERTP mints create digital assets asynchronously and are the only ERTP objects with the authority to do so.
```
katelynsills, line 258: No, minting is a synchronous method. This is not the distinction between ZCFMints and the underlying ERTP mint. The distinction is that the ZCFMint can mint assets that get associated with a seat without having to escrow payments, and burn assets that used to be associated with a seat without having to payout assets.

main/glossary/README.md
```diff
@@ -67,151 +102,362 @@ synchronously compared for structural equivalence.
...
+  can only create one type of asset and cannot change to create a different type.
 
-See the [ERTP Guide's Mint section](/ertp/guide/issuers-and-mints.md#mints) 
-and the [ERTP API's Mint section](/ertp/api/mint.md) for more information.
+  ERTP mints are [issuer's](#issuer) admin [facets](#facet), and there is a one-to-one relationship between an issuer and
+  its mint. ERTP mints are also in a one-to-one relationship with that issuer's associated [brand](#brand).
+
+- ZCFMints are synchronous Zoe mints, which mint and reallocate digital assets synchronously 
+  instead of asynchronously like ERTP mints. Similar to ERTP mints, they have one-to-one relationships
+  with an issuer and its associated brand.
```
katelynsills, line 267: No, first, objects are not synchronous or asynchronous. Second, the `mintPayment` method on the underlying ERTP mint is synchronous, so that is not the distinction. The distinction is that ZCFMints make it much easier to mint and escrow the payments to a particular seat.

katelynsills (changes requested):
This is good step in the right direction, but some corrections need to be made before merging.

main/glossary/README.md
tyg: How's this: 
[ERTP](#ertp) has a *mint* object, which creates digital assets. [ZCF](#zcf) provides a 
different interface to an ERTP mint, called a *ZCFMint*. It doesn't matter whether you
use the ERTP mint interface (henceforth called an *ERTP mint*) or the ZCFMint interface;
quatloos created by an ERTP mint are indistinguishable from quatloos created by a ZCFMint.

main/glossary/README.md
```diff
@@ -25,19 +29,20 @@ assign assets that weren't already in some allocation and it can't assign them t
...
+- `MathKind.STRING_SET`: Used with non-fungible assets. Amount values are strings.
+- `MathKind.SET`: Used with non-fungible assets. Amount values are objects or records with multiple properties.
 
-See the [ERTP Guide's AmountMath section](/ertp/guide/amount-math.md) 
-and the [ERTP API's AmountMath section](/ertp/api/amount-math.md) for more information.
+For more information, see the [ERTP Guide's AmountMath section](/ertp/guide/amount-math.md) 
+and the [ERTP API's AmountMath section](/ertp/api/amount-math.md).
 
-## Amounts
+## Amount
```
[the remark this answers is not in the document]
tyg: Duh. Yeah, occurred to me it was out of sync with the other similar entries, which are all singletons, but you're right, it's too established/common to easily change existing links.

main/glossary/README.md
tyg: Added "**Note**: Depositing into an invitation-branded purse also claims an invitation. This is
what the [wallet](#wallet) does."

main/glossary/README.md
tyg: I thought you could email an object reference? Now have "Otherwise, the contract instance must create any additional invitations. These, or any invitation held by a party, are distributed by sending it to someone's wallet. When you receive 
an invitation, you should validate it via the [InvitationIssuer](#invitationissuer)."

But if the only way is to send it to a wallet, which claims/validates by depositing it in a purse, when does anyone use the invitationIssuer to do validation of an invitation?

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+handles are unforgeable. This means the only way to know a handle identity is being given 
+an object reference, and no identity can be guessed and no fake identity will succeed. 
+
+## Harden
+A hardened object’s properties cannot be changed, so the only way to interact with a hardened object is through its methods.
+`harden()` is similar to `Object.freeze()` but more powerful. For more about `harden()`, see
+its [section in the JavaScript Distributed Programming Guide](/distributed-programming.md#harden)
+
+## IBC
+The Inter-Blockchain Communication protocol, used to by blockchains to communicate with each other. A short article about IBC
```
Chris-Hibbert, line 211:
```suggestion
The Inter-Blockchain Communication protocol, used by blockchains to communicate with each other. A short article about IBC
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
 ## Mint
-In ERTP, mints create digital assets and are the only objects with the authority to do so. 
-Access to a mint gives you the power to create more digital assets of its type at will. Mints
-can only create one type of asset. 
 
-Mints are [issuer's](#issuer) admin facets, and there is a one-to-one relationship between an issuer and
-its mint.
+[ERTP](#ertp) has a *mint* object, which creates digital assets. [ZCF](#zcf) provides a different interface to an ERTP mint, called a
+*ZCFMint*. It doesn't matter whether you use the ERTP mint interface (henceforth called an *ERTP mint*) or the ZCFMint interface;
+quatloos created by an ERTP mint are indistinguishable from quatloos created by a ZCFMint.
```
Chris-Hibbert, line 268: Except that any particular currency will only have one of those kinds of mints. I'm not sure this is a good way to phrase it. Maybe something more like "Assets and AssetHolders created using ZcfMints are usable in all the same ways as assets created by other ERTP Mints. They interact with Purses, Payments, Brands, and Issuers in the same ways."

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+
+Offers are a structured way of describing user intent. In Zoe, an offer consists of a [proposal](#proposal) (the 
+rules under which the party wants to exercise the offer) and [payments](#payment) corresponding to what the proposal specifies as what the
+party will give if the offer is satisfied. The payments are automatically [escrowed](#escrow) by Zoe, and appropriately reallocated when
+the offer exits with either success or rejection. See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord)`](https://agoric.com/documentation/zoe/api/zoe.html#e-zoe-offer-invitation-proposal-paymentkeywordrecord).
+
+## Offer Safety
+
+Zoe guarantees offer safety. When a user makes an [offer](#offer) and it is [escrowed](#escrow) with Zoe, Zoe guarantees that 
+the user either gets what they said they wanted, or gets back (refunded) what they originally offered and escrowed.
```
Chris-Hibbert, line 334:
```suggestion
the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and escrowed.
```

Chris-Hibbert (changes requested):

main/glossary/README.md
```diff
@@ -25,17 +29,20 @@ assign assets that weren't already in some allocation and it can't assign them t
...
+gives a new balance of 5 bucks. But, a deposit of a non-fungible theater ticket into a purse that already holds
+five tickets isn't done by numeric addition. Instead, you have to add a string or object to an array or record.
+
+AmountMath has a single set of polymorphic
 methods of three different kinds to deal with [fungible](#fungible) assets (values are natural numbers) and
 [non-fungible](#non-fungible) assets (values are an array or object). The three AmountMathKinds are
-- `MathKind.NAT`: Used with fungible assets. `amount` values are natural numbers (non-negative integers). Default value.
-- `MathKind.STRING_SET`: Used with non-fungible assets. `amount` values are strings.
-- `MathKind.SET`: Used with non-fungible assets. `amount` values are objects or records with multiple properties.
+- `MathKind.NAT`: Used with fungible assets. Amount values are natural numbers (non-negative integers). Default value.
```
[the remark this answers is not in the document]
tyg: Is it no longer the case it's the default? If so, there's some documentation I need to update...

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+For example, a realtor has a standard house selling agreement. The contract is the 
+code defining how that agreement works. When the realtor has a new house to sell, 
+they instantiate a new instance of their standard contract for that specific property. 
+If they have ten houses for sale, they have ten different contract instances.
+
+## CreatorInvitation
+
+An [invitation](#invitation) optionally returned by `startInstance()` that the contract instance
+creator can use. It is usually used in contracts where the creator immediately 
+sells something (auctions, swaps, etc.).
```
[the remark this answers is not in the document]
tyg: Honestly pretty sure I was told the deleted at some point. Is it definitely not the case, or should I maybe include both reasons?

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+## Eventual Send
+
+See [`E()`](#e) above.
+
+## Exit Rule
+
+Part of an [offer](#offer) specifying how the offer can be cancelled/exited. There are three values:
+- `onDemand: null`: (Default) The offering party can cancel on demand.
+- `waived: null`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
+- `afterDeadline`: The offer is automatically cancelled after a deadline, as determined by its timer and deadline properties.
```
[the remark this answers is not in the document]
tyg: My inclination is not to, both because those are terms whose technical definition here matches up with a standard definition, and because that's going a bit deeper than I think is needed in the glossary. Definitely in the detailed documentation, but I don't think it's worth it here.  (I was borderline on the specific rules, but decided they are specific enough terms and used as such to warrant inclusion).

main/glossary/README.md
tyg: Huh. Just to be sure, is it required there can be only one mint, whether ERTP or ZCF, for a given brand? If so, I should note that in the API docs.

main/glossary/README.md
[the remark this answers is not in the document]
Chris-Hibbert: I don't think "Default value" applies here. There are several methods for which `MathKind.NAT` is the default for some parameter, but that's a property of those methods, and not of MathKind itself. 

FYI, I can't think of a circumstance where we'd have one of the other types as a default, but I'm fairly certain that there are places with no default.

main/glossary/README.md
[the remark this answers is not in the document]
Chris-Hibbert: That is a circumstance in which the creatorInvitation is usually used, but it's not the case that most of the circumstances in which the creatorInvitation is available are because the creator will sell something immediately.  

Just a change in emphasis.  Most As are B, but that doesn't mean that most Bs are A.

"If the creator will usually want to immediately sell something, the contract should probably provide a creatorFacet."

main/glossary/README.md
Chris-Hibbert: > is it required there can be only one mint for a given brand? 
Absolutely. A mint can be shared, but there is only one for any given brand.

katelynsills:
Some github process notes: Whenever you think this is ready for another review, please request another review from me using the "re-request review" circular arrows button. That will put it in my inbox, so to speak. Also, if you think it's a candidate for merging, you can turn it into a real PR, not a draft PR.

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
 
-If references can only be obtained by creation, construction, or introduction, you may have a safe system. If they can be obtained in any other way, your system is unsafe.
+If references can only be obtained by creation, construction, or introduction, you may have a safe 
+system. If they can be obtained in any other way, your system is unsafe.
 
 For more information, see [Douglas Crockford on Object Capabilities](https://frontendmasters.com/courses/good-parts-javascript-web/object-capabilities/).
 
+## Offer
+
+Offers are a structured way of describing user intent. In Zoe, an offer consists of a [proposal](#proposal) (the
```
katelynsills, line 326: I think I started the "describing user intent" thing, but it's too vague for a glossary and not definitional enough. 
```suggestion
Users interact with contract instances by making offers. In Zoe, an offer consists of a [proposal](#proposal) (the 
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+system. If they can be obtained in any other way, your system is unsafe.
 
 For more information, see [Douglas Crockford on Object Capabilities](https://frontendmasters.com/courses/good-parts-javascript-web/object-capabilities/).
 
+## Offer
+
+Offers are a structured way of describing user intent. In Zoe, an offer consists of a [proposal](#proposal) (the 
+rules under which the party wants to exercise the offer) and [payments](#payment) corresponding to what the proposal specifies as what the
+party will give if the offer is satisfied. The payments are automatically [escrowed](#escrow) by Zoe, and appropriately reallocated when
+the offer exits with either success or rejection. See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord)`](https://agoric.com/documentation/zoe/api/zoe.html#e-zoe-offer-invitation-proposal-paymentkeywordrecord).
```
katelynsills, line 329: I feel like I've been repeating myself for months, but there's no general concept of an "offer being satisfied". We need to strike any such language post-haste! Also **no reallocation happens on offer exits!** I think I've said this a number of times. There is a fundamental misunderstanding of what reallocation means that we need to correct.

```suggestion
conditions under which the party makes the offer) and [payments](#payment) corresponding to what the party is willing to give in exchange for what they want. The payments are automatically [escrowed](#escrow) by Zoe, and reallocated according to the contract code. An offer gets a payout of some combination of what the party originally contributed and what others have contributed. The specific payout is determined by the contract code. See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord)`](https://agoric.com/documentation/zoe/api/zoe.html#e-zoe-offer-invitation-proposal-paymentkeywordrecord).
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+
+## Offer Safety
+
+Zoe guarantees offer safety. When a user makes an [offer](#offer) and it is [escrowed](#escrow) with Zoe, Zoe guarantees that 
+the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and escrowed.
```
katelynsills, line 334:
```suggestion
Zoe guarantees offer safety. When a user makes an [offer](#offer) and its payments are [escrowed](#escrow) with Zoe, Zoe guarantees that 
the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and escrowed.
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
-of a purse must be of the same [brand](#brand).
-
-See the [ERTP Guide's Payments section](/ertp/guide/purses-and-payments.md#payments)
-and the [ERTP API's Payments section](/ertp/api/payment.md#payment-getallegedbrand) 
-for more information.
-
-## Presence
-A local version of a remote object that serves as a proxy for the remote object. 
+
+Payments hold assets issued by [Mints](#mint). Specifically assets intended for transfer
```
katelynsills, line 338:
```suggestion
Payments hold assets created by [Mints](#mint). Specifically assets intended for transfer 
```

It's confusing to say mints are the issuer.

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
-
-See the [ERTP Guide's Payments section](/ertp/guide/purses-and-payments.md#payments)
-and the [ERTP API's Payments section](/ertp/api/payment.md#payment-getallegedbrand) 
-for more information.
-
-## Presence
-A local version of a remote object that serves as a proxy for the remote object. 
+
+Payments hold assets issued by [Mints](#mint). Specifically assets intended for transfer 
+from one party to another. All assets of a payment must be of the same [brand](#brand).
```
katelynsills, line 339:
```suggestion
from one party to another. All assets of a payment are of the same [brand](#brand).
```
`must be` sounds like we are instructing people to not do something wrong, but there is no way in which this can be otherwise.

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+
+Proposals are records with `give`, `want`, and `exit` keys. [Offers](#offer) must include a proposal, which states
+what asset you want, what asset you will give for it, and how/when the offer maker can cancel the offer
+(see [Exit Rule](#exit-rule) for details on the last). For example:
+```
+const myProposal = harden({
+  give: { Asset: quatloosAmountMath.make(4)},
+  want: { Price: moolaAmountMath.make(15) },
+  exit: { 'onDemand'
+})
```
katelynsills, line 373:
```suggestion
const myProposal = harden({
  give: { Asset: quatloosAmountMath.make(4)},
  want: { Price: moolaAmountMath.make(15) },
  exit: { 'onDemand' },
})
```

This code wouldn't run due to the missing `}`. Can we move this into a tested code file and import it to prevent things like this?

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+[ERTP API's Purses section](/ertp/api/purse.md).
 
-See the [ERTP Guide's Purses section](/ertp/guide/purses-and-payments.md#purses) and the
-[ERTP API's Purses section](/ertp/api/purse.md)
-for more information.
+## Quatloos
+An imaginary currency Agoric documentation uses in examples. For its origins, see the Wikipedia entry for the Star Trek 
+episode [The Gamesters of Triskelion](https://en.wikipedia.org/wiki/The_Gamesters_of_Triskelion).
+
+## Reallocate/Reallocation
```
katelynsills, line 390: For the purposes of the link, let's only use one of these. If we need another entry for the other title, we can point them to this one.

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+
+A transfer of [amounts](#amounts) between [seats](#seat) within Zoe; i.e. a change in their [allocations](#allocation). When a seat exits, it gets its
+current allocation as a [payout](#payout). 
 
 ## Seat
 
-Zoe uses seats to represent offers, and has two seat [facets](#facet)  a ZCFSeat and a UserSeat.
+Zoe uses seats to represent [offers](#offer), and has two seat [facets](#facet); a `ZCFSeat` and a `UserSeat`.
```
katelynsills, line 397:
```suggestion
Zoe uses a seat to represent an [offer](#offer) in progress, and has two seat [facets](#facet) representing two views of the same seat; a `ZCFSeat` and a `UserSeat`. The `UserSeat` is what is returned to the user who made an offer, and the `ZCFSeat` is the argument passed to the `offerHandler` in the contract code. The two seat facets have slightly different methods but represent the same seat and offer in progress.
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+The term comes from the expression "having a seat at the table" with regards to participating in a
+negotiation.
 
 Seats represent active offers and let contracts and users interact with them. ZCFSeats are used 
 within contracts and with `zcf.` methods. User seats represent offers external to Zoe and the 
 contract. The party who exercises an invitation and sends the `offer()` message to Zoe 
 gets a UserSeat that can check payouts' status or retrieve their results.
```
katelynsills, line 404: Hmm, this is probably superfluous with my changes above. Can we merge the two?

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
 
+For more details, see the [ZCFSeat documentation](/zoe/api/zoe-contract-facet.md#zcfseat-object) and 
+the [UserSeat documentation](/zoe/api/zoe.md#userseat-object).
+
+## SeatStagings
+`seatStagings` are associations of seats with reallocations.
```
katelynsills, line 410:
```suggestion
`seatStagings` are associations of seats with new allocations. `seatStagings` are passed to `zcf.reallocate`.
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+more details. 
+
+## Simoleons
+An imaginary currency Agoric documentation uses in examples.
+
+## Terms
+Contract instances have associated terms, gotten via `E(zoe).getTerms(instance)`,
+which include the instance's associated [issuers](#issuer), [brands](#brand), and any custom terms. For 
+example, you might have a general auction contract. When someone instantiates it,
+they provide terms applicable only to that instance. For example, for some instances of
```
katelynsills, line 427:
```suggestion
they provide terms applicable only to that instance. For some instances of 
```
Already have one 'for example'

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+
 ## Value
 
 Values are the part of an [amount](#amounts) that describe the value of something
 that can be owned or shared: How much, how many, or a description of a unique asset, such as
-Pixel(3,2), $3, or ‘Right to occupy on Tuesdays’. [Fungible](#fungible) values are usually 
+Pixel(3,2), $3, or 'Seat J12 for the show September 27th at 9:00pm'.
+[Fungible](#fungible) values are usually 
 represented by natural numbers. Other values may be represented as strings naming a particular
 right, or an arbitrary object representing the rights at issue. The latter two examples
```
katelynsills, line 438:
```suggestion
represented by natural numbers. Other values may be represented as an array of strings naming particular
rights, or an array of arbitrary objects representing the rights at issue. The latter two examples 
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+[Fungible](#fungible) values are usually 
 represented by natural numbers. Other values may be represented as strings naming a particular
 right, or an arbitrary object representing the rights at issue. The latter two examples 
 are usually [non-fungible](#non-fungible) assets. Values must be [Comparable](#comparable).
 
-See the [ERTP Guide's Value section](/ertp/guide/amounts.md#values) for more information.
+For more information, see the [ERTP Guide's Value section](/ertp/guide/amounts.md#values).
+
+## Vat 
+A vat is a unit of isolation. To paraphrase the Las Vegas advertising slogan, what happens in a vat stays in that vat.
```
katelynsills, line 444: I like the snap of this, but I'm not sure it's helpful. What happens in a vat can drastically affect objects in other vats.

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+Objects and functions in a JavaScript vat can communicate synchronously with one another. Vats and their contents can
+communicate with other vats and their objects and functions, but have to manage asynchronous messages and responses.
```
katelynsills, line 446:
```suggestion
communicate with other vats and their objects and functions, but can only communicate asynchronously. 
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+
+For more information, see the [Vat section in the Distributed JS Programming Guide](/distributed-programming.md#vats)
+
+## Wallet
+
+The overall place a party keeps their assets of all brands. For example, your wallet might contain 5 Quatloos
+[purses](#purse), 8 Moola purses, and 2 Simoleons purses. You can also keep [Issuers](#issuer) in a 
+wallet. [Offers](#offer) are associated with the wallets 
+from which their associated payments come from. See the [Wallet API](/wallet-api.md).
```
katelynsills, line 455: This is a weird phrasing. 
```suggestion
wallet. Dapps can propose offers to a wallet. If a user accepts the offer proposal, the wallet makes an offer on the user's behalf and deposits the payout in the user's purses. See the [Wallet API](/wallet-api.md).
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+Contract Facet methods can be called synchronously by contract code. 
+
+See the [ZCF API](/zoe/api/zoe-contract-facet.md).
 
-## Vat
+## ZCFMint
 
-A vat is a *unit of synchrony*. This means that within a JavaScript vat, objects and functions can communicate with one another synchronously.
+ZCFMints are Zoe mints. Similar to ERTP mints, they have one-to-one relationships
+with an [issuer](#issuer) and its associated [brand](#brand). A ZCFMint can mint assets that get associated with a [seat](#seat)
```
katelynsills, line 466: See the above language under mint for fixing `Zoe mint` and `similar to ERTP mints`.

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+without having to [escrow](#escrow) [payments](#payment), and [burn](#burn) assets that used 
+to be associated with a seat without having to payout assets.
+  
+ZCFMints and ERTP mints do **not** have the same methods. Do not try to use ERTP methods on a ZCFMint or vice versa.
+However, issuers, brands, and [amountMaths](#amountmath) associated with either an ERTP mint or a ZCFMint do have the same methods.
```
katelynsills, line 471: Same change as under mint for this language

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
 
-A vat runs a single *event loop*.
+See [Mint](#mint).
 
-A physical machine can run one or several vats. A blockchain can run one or several communicating vats.
+## Zoe Helpers
 
-Different vats can communicate by sending asynchronous messages to other vats.
+A set of API helper methods for writing contracts. These methods extract common contract code and
```
katelynsills, line 477:
```suggestion
A set of helper methods for writing contracts. These methods extract common contract code and
```

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+patterns into reusable helpers. See the [Zoe Helpers API](/zoe/api/zoe-helpers.md).
 
-A vat is the moral equivalent of a Unix Process.
+## Zoe Service
 
+A set of API methods for deploying and working with smart contracts. See [Zoe Service API](/zoe/api/zoe.md).
```
katelynsills, line 482:
```suggestion
The on-chain entity that runs contract instances and lets users interact with contract instances. See [Zoe Service API](/zoe/api/zoe.md).
```

katelynsills (changes requested):

main/glossary/README.md
```diff
@@ -67,151 +104,379 @@ synchronously compared for structural equivalence.
...
+To participate in a contract instance, one must hold an invitation to do so. Contracts often 
+return a creator invitation on their instantiation, in case the contract instantiator wants 
+to immediately participate. Otherwise, the contract instance must create any additional invitations. 
+These, or any invitation held by a party, are distributed by sending it to someone's wallet. When you receive 
+an invitation, you should validate it via the [InvitationIssuer](#invitationissuer). Note that 
+the invitation is a special case of [`Payment`](#payment), and so is associated with a specific [`Issuer`](#issuer).
+
+To participate in a contract instance by making an [offer](#offer), an invitation to that instance must accompany the offer.
+
+An `invitation`'s properties are:
```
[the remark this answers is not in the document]
tyg: Changed, but I'm confused. I thought these properties were off an invitation object, not its amount, and everywhere else we've been defining an "amount" as a value and a brand, so adding properties like contract installation and instance in a single case seems a pretty big deal and potentially confusing in general?

main/glossary/README.md
tyg: Made the suggested change, but I'm still not totally clear on the key point. Let me ask which of the following are correct:

a) For any given brand, such as Quatloos, there can only be either a ZCFMint or an ERTP mint that creates it, not both kinds of mints.

b) If you have an ERTP mint that creates Quatloos, you can associate a ZCFMint with it (how?) such that the ZCFMint serves as an interface for creating Quatloos. However, the actual creation is done by the ERTP mint.

c) The same as b), but the ZCFMint for Quatloos is created first and then an ERTP minit is somehow associated with it to be able to create Quatloos.

main/glossary/README.md
Chris-Hibbert: I wouldn't put it that way, since any particular mint is going to be of one kind or another, and there won't be duplicate mints for any brand.

assets, brands, and issuers associated with a mint (retrieved with `getIssuerRecord()`) created within a contract are indistinguishable from those created using makeIssuerKit.

main/glossary/README.md
Chris-Hibbert: Only (a) is true.  

The zcfMint is a special case so that code inside a contract can issue a new currency, distinct from any other. Having created it that way, the Issuer and brand need to be made available to anyone that will be dealing with the associated assets. The Mint will be closely held within the contract. In the case of a normal Mint, it would be closely held outside a contract, and the same statements are true about the brand, issuer, and purses and payments.

main/glossary/README.md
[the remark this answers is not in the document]
katelynsills: It's in the `value` of the amount. There's no big change here. Non-fungible assets have custom properties that can be viewed within the amount.

main/glossary/README.md
katelynsills: @Chris-Hibbert I don't think that's right. There is always an ERTP mint. It's just that in some cases, there is also a ZCFMint wrapper. So if there is a ZCFMint, there is an ERTP mint within Zoe actually minting.

a) is wrong - if there is a ZCFMint there is also an ERTP mint, just within Zoe and not available to anyone else.
b) is wrong - you cannot take an ERTP mint that already exists and create the wrapper ZCFMint after the fact.
c) is wrong - you cannot get the ERTP mint from a ZCFMint. The ERTP mint is within Zoe and not available to anyone.

The actual correct version is: If you want to be able to mint digital assets and assign them to a seat directly, within contract code, you can use Zoe to make a ZCFMint. Making a ZCFMint creates an ERTP Mint within Zoe. Calls on the ZCFMint object make calls on the ERTP Mint within Zoe, but they do some extra steps, such as escrowing the digital assets within Zoe without an offer and associating the assets with a seat. The associated ERTP Mint of the ZCFMint wrapper cannot be accessed outside of Zoe, and all minting or burning must happen by the contract code calling methods on the ZCFMint.

Chris-Hibbert: I like your explanation, but I'd de-emphasize the ERTP MInt: "The ZCFMint encapsulates an ERTP Mint, and ensures that all assets created by it are assigned to a zcfSeat."  

The point that Tom is trying to clarify is that the `zcf.makeZCFMint()` produces an issuerKit that clients interact with in the same way as issuerKits from `makeIssuerKit()`. The brands, issuers, and purse and payments work the same way, and are (re-)usable by the same code.

katelynsills (approved):
Looks good! Just one suggestion

main/glossary/README.md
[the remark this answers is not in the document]
tyg: Interesting. I don't think that's mentioned anywhere in the docs. Should something be added to the Amount's Value section in the ERTP Guide about it?

main/glossary/README.md
tyg: OK, I think I'm clear on there ultimately being an ERTP mint that mints Quatloos. It may or may not be encapsulated by a ZCFMint.  But it's still not quite clear to me if/when a mint is unique. My question now could be expressed as, let's say the asset brand is Quatloos. In a given contract instance, is it possible to mint new Quatloos using both:
 mint.mintPayment(newAmount) (the ERTP method) or:

 const mySynchronousMint = await zcf.makeZCFMint('MySyncMint', 'set');
 mySynchronousMint.mintGains({ MyKeyword: amount }, seat); (the ZCFMint method)

Or do you have to set up only one or the other in a contract instance, and once one is created, whether explicitly ERTP or ZCF-encapsulating-ERTP, only that one will ever be able to create Quatloos. Or to phrase it another way, is it the case that exactly one object can execute a command to make new Quatloos, or could both of those objects above mint Quatloos?

Chris-Hibbert: If a contract ever had both at the same time, they would necessarily be distinct mints. They might even claim to have the same name, but they would create payments and purses that were incompatible with one another.

katelynsills: > if/when a mint is unique

In a given contract instance, you cannot ever get the ERTP mint and the ZCFMint that encapsulates it, because when a ZCFMint is created, the ERTP mint does not escape Zoe. It's not accessible to the contract.

> Or do you have to set up only one or the other in a contract instance

There's no way to setup both the ZCFMint and the ERTP mint within it in an accessible way, because the ERTP mint that a ZCFMint wraps doesn't leave Zoe.

> once one is created, whether explicitly ERTP or ZCF-encapsulating-ERTP, only that one will ever be able to create Quatloos.

Only an ERTP mint ever actually makes Quatloos. The ZCFMint provides a method for the contract to mint. That method calls `mint.mintPayment` internally in Zoe. Another way to say this is that ZCFMints are only able to mint because they call the `mintPayment` method on the ERTP mint within Zoe that they are associated with.

> exactly one object can execute a command to make new Quatloos, or could both of those objects above mint Quatloos?

Exactly one object can execute the actual command that makes quatloos, the ERTP Mint. Zoe sets up the ZCFMint such that `myZCFMint.mintGains(...` internally calls `myERTPMint.mintPayment(...`

main/glossary/README.md
katelynsills: The language above isn't right, but the language used in the document "**Note**: Depositing into an invitation-branded purse also validates an invitation. This is
what the [wallet](#wallet) does." is correct.

katelynsills (approved):
Looks good to me!

main/glossary/README.md
```diff
@@ -46,18 +54,50 @@ abstract right to participate in a particular exchange.
...
 ## AssetHolder
-[Purses](#purse) and [payments](#payment) are AssetHolders. These are objects that contain [amounts](#amounts).
+[Purses](#purse) and [payments](#payment) are AssetHolders. These are objects that contain
+digital assets in the quantity specified by an [amount](#amounts).
+
+## Board (Agoric Board)
+
+The Board is a shared, on-chain location where users can post a value and make it 
+accessible to others. When a user posts a value, they receive a unique Id for the value. 
+Others can get the value just by knowing the Id. You can make an Id known by any
```
Chris-Hibbert, line 75:
```suggestion
accessible to others. When a user posts a value, they receive a unique ID for the value. 
Others can get the value just by knowing the ID. You can make an ID known by any 
```

main/glossary/README.md
```diff
@@ -67,151 +107,376 @@ synchronously compared for structural equivalence.
...
+- ERTP mints create digital assets and are the only ERTP objects with the authority to do so. 
+  Access to an ERTP mint gives you the power to create more digital assets of its type at will. Mints
+  can only create one type of asset and cannot change to create a different type.
+
+  ERTP mints are [issuer's](#issuer) admin [facets](#facet), and there is a one-to-one relationship between an issuer and
+  its mint. ERTP mints are also in a one-to-one relationship with that issuer's associated [brand](#brand).
+
+- ZCFMints give contract code a simpler interface to interact with an ERTP mint. Because ZCFMints encapsulate
+  an internal ERTP mint, they have the same one-to-one relationships
+  with an issuer and its associated brand. A ZCFMint can mint assets that get associated with a seat without having to escrow
```
Chris-Hibbert, line 283:
```suggestion
  with an issuer and its associated brand. A ZCFMint can mint assets and assign them to a seat without having to escrow
```

main/glossary/README.md
```diff
@@ -67,151 +107,376 @@ synchronously compared for structural equivalence.
...
-as they are not interchangeable (and may even have different prices). See also [fungible](#fungible).
+as they are not interchangeable (and may have different prices). See also [fungible](#fungible).
 
 ## Notifier
 
-You can track updates to contract state using a notifier. The notifier provides a
-stream of updates describing changes to the state of an offer.
-
-See more: [Notifier](/distributed-programming.md)
+A notifier provides a stream of updates describing changes to the state of an [offer](#offer).
```
Chris-Hibbert, line 305:
```suggestion
A notifier provides a stream of updates describing changes to the state of an [offer](#offer) or other object.
```

main/glossary/README.md
```diff
@@ -67,151 +107,376 @@ synchronously compared for structural equivalence.
...
+If references can only be obtained by creation, construction, or introduction, you may have a safe 
+system. If they can be obtained in any other way, your system is unsafe.
 
 For more information, see [Douglas Crockford on Object Capabilities](https://frontendmasters.com/courses/good-parts-javascript-web/object-capabilities/).
 
+## Offer
+
+Users interact with contract instances by making offers. In Zoe, an offer consists of a [proposal](#proposal) (the 
+conditions under which the party makes the offer) and [payments](#payment) corresponding to what the party is willing
+to give in exchange for what they want. The payments are automatically [escrowed](#escrow) by Zoe, and reallocated
```
Chris-Hibbert, line 333: a proposal has three parts: the exit conditions, the give, and the want. The give's structure matches that of the payments. 

```suggestion
Users interact with contract instances by making offers. In Zoe, an offer consists of a [proposal](#proposal) (what the party making the offer is willing to give up and what they want in exchange) and [payments](#payment) corresponding to the amount the proposal said they're willing
to give. The payments are automatically [escrowed](#escrow) by Zoe, and reallocated 
```

My wording may be too verbose, but it's worth pointing out the parallelism.

main/glossary/README.md
```diff
@@ -67,151 +107,376 @@ synchronously compared for structural equivalence.
...
+conditions under which the party makes the offer) and [payments](#payment) corresponding to what the party is willing
+to give in exchange for what they want. The payments are automatically [escrowed](#escrow) by Zoe, and reallocated 
+according to the contract code. An offer gets a payout of some combination of what the party originally contributed
+and what others have contributed. The specific payout is determined by the contract code. 
+See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord)`](https://agoric.com/documentation/zoe/api/zoe.html#e-zoe-offer-invitation-proposal-paymentkeywordrecord).
+
+## Offer Safety
+
+Zoe guarantees offer safety. When a user makes an [offer](#offer) and its payments are [escrowed](#escrow) with Zoe, Zoe guarantees that 
+the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and escrowed.
```
Chris-Hibbert, line 341:
```suggestion
the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and
escrowed. One reason this is possible is that if a proposal doesn't match what the contract expects to do, it
can immediately cause the seat to exit, getting back the amount it offered.
```

main/glossary/README.md
```diff
@@ -67,151 +107,376 @@ synchronously compared for structural equivalence.
...
+
+The two seat facets have slightly different methods but represent the same seat and offer in progress. 
+The term comes from the expression "having a seat at the table" with regards to participating in a negotiation.
+
+For more details, see the [ZCFSeat documentation](/zoe/api/zoe-contract-facet.md#zcfseat-object) and 
+the [UserSeat documentation](/zoe/api/zoe.md#userseat-object).
+
+## SeatStagings
+`seatStagings` are associations of seats with new [allocations](#allocation). `seatStagings` are
+passed to `zcf.reallocate`.
```
Chris-Hibbert, line 417:
```suggestion
passed to `zcf.reallocate()`.
```

main/glossary/README.md
```diff
@@ -67,151 +107,376 @@ synchronously compared for structural equivalence.
...
+Objects and functions in a JavaScript vat can communicate synchronously with one another. Vats and their contents can
+communicate with other vats and their objects and functions, but can only communicate asynchronously. 
+
+For more information, see the [Vat section in the Distributed JS Programming Guide](/distributed-programming.md#vats)
+
+## Wallet
+
+The overall place a party keeps their assets of all brands. For example, your wallet might contain 5 Quatloos
+[purses](#purse), 8 Moola purses, and 2 Simoleons purses. You can also keep [Issuers](#issuer) in a 
+wallet. Dapps can propose [offers](#offers) to a wallet. If a user accepts the offer proposal,
```
Chris-Hibbert, line 461:
```suggestion
[purses](#purse), 8 Moola purses, and 2 Simoleons purses. A wallet can distinguish between [Issuers](#issuer). Dapps can propose [offers](#offers) to a wallet. If a user accepts the offer proposal, 
```

Chris-Hibbert (approved):
This is looking very good. I have a handful of suggestions, but none of them are mandatory. If you're happier with your wording, I'm fine with this doc as it is.

Issue #206, opened 2020-06-05
Minor mistake in installation instructions


FUDCo:
(Actually, I'm not entirely sure if this is a documentation problem or an issue with the installer software itself, but I'll treat it as a doc issue for now as it will lead to doc changes however it's dealt with).

Step 9 of https://agoric.com/documentation/getting-started/before-using-agoric.html says "Select a directory to install the Agoric CLI in. For example, `yarn link-cli ~/bin/agoric`"  However, the thing you are specifying is not actually a directory name but the name of the Agoric CLI utility itself.  I was thinking I wanted to put it in my `bin` directory, so I typed `yarn link-cli ~/bin` and it failed with the complaint that the directory `/Users/chip/bin` already existed.  So I tried again specifying `~/bin/agoric` and that created an `agoric` file in my `bin` directory; that file is a shell script, not a directory.

Either the doc should be changed to tell you to name the file for your Agoric CLI utility or the `yarn` script should treat the command line arg as the name of the directory to put `agoric` into.  (Actually, I'd argue for the latter rather than just changing the doc because it makes a lot more sense and having the name of the utility be standardized will make the other docs more straightforward.  Though if we do that the doc will _still_ need to be changed to have the example not suggest `~/bin/agoric` but rather just `~/bin`, which does make more sense.

michaelfig:
The doc should be updated.

If we choose to break compatibility with the existing command in the future, we should change its name, too, to avoid confusion.

PR #397, opened 2020-11-30
Expanded and clarified explanation of what's happening in the final s…

tyg:
…tep per doc issue #206

For why I did this, see https://github.com/Agoric/documentation/issues/206
Not really necessary though, if you agree the explanation is clearer and/or more informative now.

### DIFF
--- a/main/getting-started/before-using-agoric.md
+++ b/main/getting-started/before-using-agoric.md
@@ -90,9 +90,11 @@ software, you need to install the following.
       <td>9</td>
       <td>Install the Agoric CLI by: <code>yarn link-cli 
         &lt;<i>agoric script location</i>&gt;</code></td>
-      <td>Select a location for the Agoric CLI program. For example,
+      <td>Select a location for the Agoric CLI program to be called from. For example,
         <code>yarn link-cli /usr/local/bin/agoric</code> (or if that fails
         with permission problems, <code>sudo yarn link-cli /usr/local/bin/agoric</code>)
+        This creates a shell script named <code>agoric</code> in the specified directory.
+        The script runs the <code>agoric</code> commands that make up the Agoric CLI.
         </td>
     </tr>
   </tbody>
--- a/main/getting-started/before-using-agoric.md
+++ b/main/getting-started/before-using-agoric.md
@@ -90,12 +90,13 @@ software, you need to install the following.
       <td>9</td>
       <td>Install the Agoric CLI by: <code>yarn link-cli 
         &lt;<i>agoric script location</i>&gt;</code></td>
-      <td>Select a location for the Agoric CLI program to be called from. For example,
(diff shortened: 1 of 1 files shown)


main/getting-started/before-using-agoric.md
```diff
@@ -90,9 +90,11 @@ software, you need to install the following.
       <td>9</td>
       <td>Install the Agoric CLI by: <code>yarn link-cli 
         &lt;<i>agoric script location</i>&gt;</code></td>
-      <td>Select a location for the Agoric CLI program. For example,
+      <td>Select a location for the Agoric CLI program to be called from. For example,
```
michaelfig, line 93:
```suggestion
      <td>Select a location for the Agoric CLI wrapper script. For example,
```

main/getting-started/before-using-agoric.md
```diff
@@ -90,9 +90,11 @@ software, you need to install the following.
...
         <code>yarn link-cli /usr/local/bin/agoric</code> (or if that fails
         with permission problems, <code>sudo yarn link-cli /usr/local/bin/agoric</code>)
+        This creates a shell script named <code>agoric</code> in the specified directory.
+        The script runs the <code>agoric</code> commands that make up the Agoric CLI.
```
michaelfig, line 97:
```suggestion
        This creates a <code>/usr/local/bin/agoric</code> shell script that you can run to access the SDK's Agoric CLI.
```

michaelfig (changes requested):
It was a little bit inaccurate, so I tried to elaborate further.  Feel free to make more edits.

Thanks!

main/getting-started/before-using-agoric.md
tyg: I'm a little bit confused about the correction. In the example case established on lines 94-95, yes it's creating /usr/local/bin/agoric. But my understanding is that it'll create the agoric file in whatever directory is specified in the command, thus my current phrasing above in line 96. If it's always created in /usr/local/bin/agoric, I need to also change something in the previous column's text. Also, is it "can run"? If this is what the script does, I'd prefer the more definitive "shell script that when run accesses the.."

michaelfig: It doesn't have to be called "agoric".  If you say `/usr/local/bin/fororo`, then it will create a file called "fororo".

If you prefer "shell script that when run..." we can leave it that way.

tyg: How about "This creates a shell script that when run accesses the SDK's Agoric CLI. We strongly recommend putting it at <code>/usr/local/bin/agoric</code>, but its specific location and name is up to you."    Which covers both a possible different directory than /usr/local/bin and a possible different name, but gives strong guidance as to a best practice for it.

michaelfig: I like it!

PR #398, opened 2020-11-30
Markdown links don't work when in HTML code. Fixed.

tyg:
Also redirected the links to the Intros to ERTP and Zoe instead of the much more overwhelming Guides to each.


katelynsills (approved):
LGTM!

PR #404, opened 2020-12-09
Clarified that offerTo can be between different contracts.

tyg:


katelynsills (approved):
Ok, I thought I had already covered this with `any contract instance to any other contract instance` but stating it again in different words sounds like a good idea.

PR #406, opened 2020-12-10
Added getAllegedBrand() and getCurrentAmountNotifier()

tyg:
Will edit Guide after this is approved.

### DIFF
--- a/main/ertp/api/purse.md
+++ b/main/ertp/api/purse.md
@@ -162,15 +162,10 @@ purse.getCurrentAmount();
 ## purse.getAllegedBrand()
 - Returns: `{Brand}`
 
-Returns an alleged brand, indicating what kind of digital asset the purse purports to hold.
-Since a purse is not trusted, this result should be treated with suspicion. Either
-verify the returned brand with its issuer, or check the result when you use it. Any successful 
-operation on the purse that involves branded assets (`withdraw`, `deposit`, `getCurrentAmount`, 
-`depositFacet.receive`) verifies the brand for that purse.
+Returns an alleged brand (Note: a `Brand`, not a `string` as `allegedName()` methods do), 
+indicating what kind of digital asset the purse purports to hold. This can identify the 
+purse's brand if the purse was made by a trusted issuer using `issuer.makeEmptyPurse()`.
 
 ```js
 const purseBrand = quatloosPurse.getAllegedBrand();
 ```
-
-
-
--- a/main/ertp/api/purse.md
+++ b/main/ertp/api/purse.md
@@ -53,6 +53,22 @@ const quatloosPurse.deposit(quatloos5);
 const currentBalance = quatloosPurse.getCurrentAmount();
 ```
 
+## purse.getCurrentAmountNotifier()
+- Returns: `{Notifier<Amount>}`
+
+Returns a lossy notifier for changes to this purse's balance. For more details,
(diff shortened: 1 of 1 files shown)


main/ertp/api/purse.md
```diff
@@ -142,3 +158,19 @@ issuer.getAmountOf(withdrawalPayment);
...
 purse.getCurrentAmount();
 ```
+
+## purse.getAllegedBrand()
+- Returns: `{Brand}`
+
+Returns an alleged brand, indicating what kind of digital asset the purse purports to hold.
+Since a purse is not trusted, this result should be treated with suspicion. Either
+verify the returned brand with its issuer, or check the result when you use it. Any successful 
+operation on the purse that involves branded assets (`withdraw`, `deposit`, `getCurrentAmount`,
```
katelynsills, line 168: Hmm, do we have this language elsewhere? I am pondering whether this is true. 

If "verifying the brand" means "this purse was made with the issuer with a one-to-one correspondence to the brand", then a successful operation like (`withdraw`) doesn't show that, since a fake purse that I whip up could appear successful but do nothing to the issuer balances.

If "verifying the brand" means "this purse safely deals with assets of this brand", I don't think a successful operation like `withdraw` shows that either. A later call or a call of a different method could fail if the `purse` isn't made correctly. 

I think the only way that we should recommend for people to verify that a purse is of a particular brand is if the purse was made via `issuer.makeEmptyPurse()`.

katelynsills (commented):
Question about the explanation under purse.getAllegedBrand()

main/ertp/api/purse.md
tyg: And yep, that's why I wanted to wait on updating the Guide; I wasn't at all sure about this. All that's in the types file is "Returns the purse's alleged brand". Noting that it's actually returning a Brand and not a name/string, I thought it should have a bit more than that about what it is and how it's used. Doing a search, the other allegedBrand was for payments, and the above is a modification of what we've got for that. So my question now is if the allegedBrand isn't useful for verification 1) What is the point/use of getting it? and 2) Should the payment.getAllegedBrand description also be modified?

katelynsills: `purse.getAllegedBrand()` is used to tell you what kind of purse it is supposed to be, of the various purses that you have, assuming those purses have been made from issuers using `makeEmptyPurse`.

For the payments, in the docs we have "Any successful operation by the `issuer` for that `brand` done on the `payment` verifies the `payment`." This is correct, because the issuer is the correct authority to verify payments. Purses work differently, though. An issuer operates on a payment (`issuer.isLive(payment)`), but does not operate on a purse (nothing like `issuer.isLive(purse)` exists). A purse has methods that are called directly. This is important, because it changes the trust relationship. In `issuer.isLive(payment)` we trust the issuer and do not trust the payment. In `purse.getCurrentAmount()`, we trust the purse to give us the balance, but only because earlier we made it with `issuer.makeEmptyPurse()`.

As an aside, `purse.getAllegedBrand()` could maybe be `purse.getBrand()` as long as we can always guarantee that a purse is made properly.

I think what we can write is something like: "`purse.getAllegedBrand()` can be used to identify the brand of a purse among purses made with trusted issuers using `issuer.makeEmptyPurse()`."

katelynsills (approved):
LGTM! Thanks!

PR #420, opened 2021-01-27
Addressed missing dependency

tet123:
I ran into this issue when I setup the env several months ago and it happened again while I was helping a friend who is not versed in the JS-ecosystem. We both run Ubuntu 20.04 LTS.

Please let me know if you rather have an issue rather than a PR.


katelynsills (approved):
Thank you for adding this!

PR #430, opened 2021-02-07
Update wallet-api.md

zarutian:
Change to reflect that an wallet bridge facet .getPursesNotifier() only gets attenuated purses state.

Just one word change btw.

(This is due to the authority leakage fix I made in svelte-wallet-dapp in agoricSDK)


zarutian:
It is ready for review.

Pinging @katelynsills or anyone that has docu duties.

katelynsills (approved):
LGTM! Thank you for the PR!

PR #437, opened 2021-02-15
Ses initial draft review

tyg:


dckc:
I'm new to reviewing docs. The "files changed" view here has lots of distracting markup and doesn't make clear what context is available to the reader.

CONTRIBUTING.md says

```
4. Check your changed docs in a local copy of the repo and local documentation 
   server.
```

but I don't see details on how to start a local documentation server.

The `yarn.lock` suggests `yarn install` should get dependencies, and it runs without error, but then the conventional `yarn start` loses with `error Command "start" not found.` Inspecting `package.json` suggests `yarn docs:dev` but that fails with `vuepress: not found`.

Suggestions?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+- `console` is available to help with debugging. Since all JavaScript implementations
+  add it, you may be surprised it’s not in the official spec. So leaving it out would cause 
+  too much confusion. Note that `console log`’s exact behavior is up to the host program; display 
+  to the operator is not guaranteed. Also, it is difficult at best to write to `process.stdout` and
+  so that should not be done.
+
+- `lockdown()` and `harden()` both an object’s API surface (enumerable data properties). 
+  A hardened object’s properties cannot be changed, so the only way to interact with a 
+  hardened object is through its methods. `harden()` is similar to `Object.freeze()` but more 
+  powerful. See the individual [`lockdown()`](#lockdown) and [`harden()`](#harden) sections
```
dckc, line 192: The word "powerful" is sort of taken in this context. Perhaps "more thorough"?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+
+- `console` is available to help with debugging. Since all JavaScript implementations
+  add it, you may be surprised it’s not in the official spec. So leaving it out would cause 
+  too much confusion. Note that `console log`’s exact behavior is up to the host program; display 
+  to the operator is not guaranteed. Also, it is difficult at best to write to `process.stdout` and
+  so that should not be done.
+
+- `lockdown()` and `harden()` both an object’s API surface (enumerable data properties). 
+  A hardened object’s properties cannot be changed, so the only way to interact with a 
+  hardened object is through its methods. `harden()` is similar to `Object.freeze()` but more
```
dckc, line 191: well, the properties of a hardened object are still readable.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+  to the operator is not guaranteed. Also, it is difficult at best to write to `process.stdout` and
+  so that should not be done.
+
+- `lockdown()` and `harden()` both an object’s API surface (enumerable data properties). 
+  A hardened object’s properties cannot be changed, so the only way to interact with a 
+  hardened object is through its methods. `harden()` is similar to `Object.freeze()` but more 
+  powerful. See the individual [`lockdown()`](#lockdown) and [`harden()`](#harden) sections
+  below. 
+
+- `HandledPromise` is also a global.
```
dckc, line 195: Is `HandledPromise` part of SES? or tildot? Not by my understanding.

Are we trying to document SES or the Agoric vat code runtime environment here?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+JavaScript code runs in the context of 
+a [*Realm*](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-code-realms). A 
+realm is the set of *primordials* (objects and functions of the standard library 
+like `Array.prototype.push`) and a global object. In a web browser, an iframe is a realm. 
+In Node.js, a Node process is a realm.
+
+For historical reasons, the ECMAScript specification requires the *primordials*
+to be mutable (`Array.prototype.push = yourFunction` is valid ECMAScript but not 
+recommended). With SES you can turn the current realm into an *immutable realm*; 
+a realm within which the primordials are deeply frozen.
```
dckc, line 226: tell us how.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+
+After calling lockdown, the harden function ensures that every object in the transitive 
+closure over property and prototype access starting with that object has been frozen by 
+`Object.freeze()`. This means that the object can be passed among programs and none of 
+those programs will be able to tamper with the surface of that object graph. They can 
+only read the surface data and call the surface functions.
+
+You have to harden a class before you harden any of its instances; i.e. it takes two separate 
+steps to harden both a class and its instances. Harden a base class before hardening classes 
+that inherit from it. harden()` does transitive freezing by following the object’s own
```
dckc, line 369: missing ` here?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+properties (as opposed to properties it inherited), and the objects whose own properties refer 
+to them, and so forth.
+
+When you freeze an object via `harden()`, it ensures any external callers 
+can only interact with it through functions in the object’s API. `harden()` 
+is an enhanced transitive version of `Object.freeze()`, which only locks up an 
+object's own properties.
+
+All objects that are transferred across a trust boundary must have their API
```
dckc, line 378: Didn't we say this above, but with "should" rather than "must"?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
+# SES Guide
+
+This document is a guide to understanding *SES (Secure ECMAScript)*. In addition
```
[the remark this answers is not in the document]
tyg: I'm afraid it is needed. Kate's mentioned that there's been feedback about duplication in the doc set. The problem is that the same person is likely to be two different audiences; first, someone getting started, who'll potentially want a good grounding in detail about what's going on and why (this doc) to best really understand things. But then they'll come back three months later and just want to find out as quickly as possible what options lockdown() has (for example). And that's what the SES-Reference doc will be. So we need to have a short introduction in both the Guide and the Reference explaining what each one is about and that the other one exists and why you might want to use it instead.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+background on the reasons why functionality was added or removed from JavaScript to do so.
+This is intended for initial reading when starting to use or learn about Agoric. For 
+a more limited description of how to use SES without much explanation, see 
+the [SES Reference](./ses-reference.md).
+
+## What is SES
+
+SES (*Secure ECMAScript*) is a JavaScript runtime library for safely running 
+third-party code inside a featherweight compartment. It addresses JavaScript’s 
+lack of security, which is particularly significant in that JavaScript applications
```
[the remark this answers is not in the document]
tyg: I'm not really seeing the problem with saying JavaScript has a lack of security; isn't the point of SES to be a secure JavaScript? In this case, I'm referring to "JavaScript" as, well, what you get when you get "all" of JavaScript with SES being a subset of JavaScript with some additions not in JavaScript.  To go back to a previous language I've used, Lisp would be the whole "concept" of Lisp, but Common Lisp would be its own version of Lisp, as would Interlisp. Does anyone else have a problem with the phrasing?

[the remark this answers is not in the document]
tyg: I did add in the second paragraph (slightly reworded to tighten). The diagram's worth thinking about, but I'll have to redo it to avoid adding in the several other areas shown in it such as Jessie and JS-Strict that, currently at least, aren't being mentioned in these docs.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+use and rely on a lot of third-party code (modules, packages, libraries, 
+user-provided code for extensions and plug-ins, etc.). 
+
+SES is a safe deterministic subset of "strict mode" JavaScript. This means 
+it does not include any IO objects that provide 
+[*ambient authority*](https://en.wikipedia.org/wiki/Ambient_authority). SES 
+also removes non-determinism by modifying a few built-in objects. It also uses
```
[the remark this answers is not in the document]
tyg: They get mentioned when we start looking at each SES aspect in depth. Specifics like that are distractions in the high level overview/intro this is meant to be.

main/javascript-modifications/ses-guide.md
tyg: Yep. This is something I was aware of with respect to the review. Some of this is from a paper of Brian's that I edited, and I'm not sure which parts were SES specific and which a different part of Agoric's development environment/platform. I figured y'all would be able to tell me which bits don't belong in the SES docs. : -)

main/javascript-modifications/ses-guide.md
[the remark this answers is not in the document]
tyg: No, this does need to be here. The problem is we've gotten feedback that some folk are bothered by duplication in the doc set. But some amount is necessary because the same person is likely to be two different audiences. When starting out, they'll hopefully want to read this, SES-Guide, which goes into detail about how and why SES is like it is to develop a good understanding of SES. But three months from now, they'll, for example, just want to quickly find what options go with lockdown(). Which is what the SES-Reference will be for; a stripped down version of the Guide with just "here's what this does and how to use it". So we need an intro paragraph in each explaining what this SES doc is for, and that the other's available if you want that version of SES documentation

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+presented as names in the global scope (some also added to Node.js). None are available in a 
+SES environment. The most surprising removals include `atob`, `TextEncoder`, and `URL`.
+
+`debugger` is a first-class JavaScript statement, and behaves as expected.
+
+## What does SES add to standard Javascript
+
+- `console` is available to help with debugging. Since all JavaScript implementations
+  add it, you may be surprised it’s not in the official spec. So leaving it out would cause 
+  too much confusion. Note that `console log`’s exact behavior is up to the host program; display
```
kriskowal, line 186:
```suggestion
  too much confusion. Note that `console.log`’s exact behavior is up to the host program; display 
```

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+
+`debugger` is a first-class JavaScript statement, and behaves as expected.
+
+## What does SES add to standard Javascript
+
+- `console` is available to help with debugging. Since all JavaScript implementations
+  add it, you may be surprised it’s not in the official spec. So leaving it out would cause 
+  too much confusion. Note that `console log`’s exact behavior is up to the host program; display 
+  to the operator is not guaranteed. Also, it is difficult at best to write to `process.stdout` and
+  so that should not be done.
```
kriskowal, line 188:
```suggestion
  to the operator is not guaranteed. Use the console for debug information only. The console is not obliged to write to the POSIX standard output and is a poor substitute for expressly threading that capability into a compartment.
```

main/javascript-modifications/ses-guide.md
kriskowal: Concur: `HandledPromise` is specific to the Agoric programming environment, and will also be part of the Endo programming environment, and is not SES.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+  A hardened object’s properties cannot be changed, only read, so the only way to interact with a 
+  hardened object is through its methods. `harden()` is similar to `Object.freeze()` but more 
+  thorough. See the individual [`lockdown()`](#lockdown) and [`harden()`](#harden) sections
+  below. 
+
+- `HandledPromise` is also a global. 
+  The [`E` wrapper (`E(target).method-name(args)`)](https://agoric.com/documentation/distributed-programming.html#communicating-with-remote-objects-using-e) 
+  can be imported as `import { E } from '@agoric/eventual-send`. These two are defined by the 
+  TC39 [Eventual-Send Proposal](https://github.com/tc39/proposal-eventual-send). In addition, "tildot" syntax (`target~.methodname(args)`) 
+  is expanded as the code is first bundled, so it can be used (but not inside `eval()` calls or new compartments).
```
kriskowal, line 200: non-ses

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+
+- `Compartment` (a [part of SES](https://github.com/Agoric/SES-shim/tree/SES-v0.8.0/packages/ses#compartment)) is 
+  a global. Code runs inside a `Compartment` and can create sub-compartments to host other 
+  code (with different globals or transforms).
+
+  Note that these child compartments get `harden` and `Compartment`, but you have to explicitly 
+  provide any other JS additions (including `console` and `HandledPromise`) as “endowments” since 
+  they won’t be present otherwise. If the parent compartment is metered, its child compartments 
+  are always metered too. Child compartments will *not* be frozen by default: 
+  see [Frozen globalThis](#frozen-globalthis) below for details.
```
kriskowal, line 210: remove some non-ses:

```suggestion
  Note that these child compartments get `harden` and `Compartment`.
```

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+JavaScript code runs in the context of 
+a [*Realm*](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-code-realms). A 
+realm is the set of *primordials* (objects and functions of the standard library 
+like `Array.prototype.push`) and a global object. In a web browser, an iframe is a realm. 
+In Node.js, a Node process is a realm.
+
+For historical reasons, the ECMAScript specification requires the *primordials*
+to be mutable (`Array.prototype.push = yourFunction` is valid ECMAScript but not 
+recommended). With SES you can turn the current realm into an *immutable realm*; 
+a realm within which the primordials are deeply frozen.
```
kriskowal, line 227:
```suggestion
recommended). By using the SES shim and calling `lockdown()`, you can turn the current realm into an *immutable realm*; 
a realm within which the primordials are deeply frozen. 
```

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+
+SES also lets programs create *Compartments*. These are "mini-realms".
+A Compartment has its own dedicated global object and environment, but 
+it inherits the primordials from their parent realm. Components are described
+in detail in the next section. 
+
+Agoric deploy scripts and smart contract code run in an immutable
+realm with Compartments providing just enough authority to create
+useful and secure contracts. But not enough authority to do anything
+unintended or harmful to the participants of the smart contract.
```
kriskowal, line 237: This could be an Agoric-specific aside.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+  'dependency': dependency.module('./main.js'),
+}, {
+  resolveHook,
+  importHook,
+});
+```
+
+Compartments greatly simplify using SES for large applications. We will build tools that use 
+compartments to load and bundle existing libraries and then execute them with a compartment
+for each package. **tyg todo Has this happened? If so, where/what are the tools? If not, any ETA?**
```
kriskowal, line 327: This material was good for the blog but we should just omit it from the guide. This will be Endo.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+This will provide a seamless experience as simple as <script src=“app.js”> or node app.js, but 
+with the safety of SES.
+  
+**tyg todo: My problem with this section is that while I think I understand what compartments are
+and what they do, I'm not clear on just how/when they are used. This (and the SES reference) could
+use what amounts to a paragraph or two like "When writing a contract, you use compartments like this. 
+This is when you want to create a compartment. This is when you don't want to. Do you ever destroy a
+compartment? If so, how?**
```
kriskowal, line 335: Compartments provide a low-level loader API for JavaScript modules. Your code might run in compartments, but they are an implementation detail of tools and runtimes.

The Agoric runtime uses compartments to isolate contracts within a Vat. MetaMask’s LavaMoat uses a Compartment for every module, to create boundaries between application code and third-party dependencies.

The lifetime of a compartment is bounded by garbage collection and the lifetime of the realm that contains them. There is no ceremony for tearing down a compartment.

main/javascript-modifications/ses-guide.md
kriskowal: I’m not well-versed at making that distinction. “should” seems appropriate since it is possible to pass unfrozen objects over a trust boundary, but “must” seems appropriate since doing so would invalidate the notion of trust over the boundary.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+The general rule is that if you make a new object and give it to someone else 
+(and don't immediately forget it yourself), you should give them `harden(obj)` 
+instead of the raw object. This prevents someone from adding/deleting the properties
+or prototypes of that object. Being hardened doesn't preclude an object from having 
+access to mutable state (`harden(new Map())` still behaves like a normal mutable `Map`), 
+but it means their methods stay the same and can't be surprisingly changed by someone else
+
+Defined objects (`mint`, `issuer`, `zcf`, etc.) shouldn't need hardening as their 
+constructors should do that work. It's mainly records, callbacks, and ephemeral 
+objects that need hardening.
```
kriskowal, line 394: This could be a breakout using the Agoric environment as an example of uses for `harden`. Would need to rework this to make that clear:
```suggestion
For example, the Agoric programming environment for smart contracts defines objects (`mint`, `issuer`, `zcf`, etc.) that shouldn't need hardening as their 
constructors should do that work. It's mainly records, callbacks, and ephemeral 
objects that need hardening.
```

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+
+You can send a message to a hardened object. If it's a record, you can access 
+its properties and their values.
```
kriskowal, line 397: This is out-of-place since we aren’t discussing eventual-send.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+try adding `/* global harden */` to the top of the file.
+
+You use `harden()` like this:
+```js
+const o = {a: 2};
+o.a  = 12;
+console.log(o.a); // 12 because o is still mutable
+harden(o);
+o.a  = 37; // throws a TypeError because o is now hardened
+```
```
kriskowal, line 412: Is it possible to make this a breakout? Perhaps with Markdown’s `>`?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+properties cannot be changed. The only way to interact with frozen objects is through 
+their methods. Their differences are what objects you use them on, and when you use them.
+
+`lockdown()` **must** be called first. It hardens JavaScript's built-in *primordials* 
+(implicitly shared global objects) and enables `harden()`. If you call `harden()` 
+before `lockdown()` excutes, it throws an error.
+
+`lockdown()` works on objects created by the JavaScript language itself as part of 
+its definition. Use `harden()` to freeze objects created after `lockdown()`was called;
+i.e. objects created by programs written in JavaScript.
```
kriskowal, line 425: This appears to be a duplicate in this document.

main/javascript-modifications/ses-reference.md
```diff
@@ -0,0 +1,148 @@
...
+- [Features from browser environments](https://developer.mozilla.org/en-US/docs/Web/API) presented as names in the global scope including `atob`, `TextEncoder`, and `URL`.
+- HTML comments
+- Dynamic `import` expressions
+- Direct evals
+
+## Added/Changed by SES summary
+
+SES adds the following to JavaScript or changes them significantly: 
+- console
+- `HandledPromise()`
```
kriskowal, line 25: HandledPromise is an Agoric contract environment addition (and will be in Endo)

kriskowal (commented):
I’ve made a partial crawl. Will continue.

main/javascript-modifications/ses-guide.md
tyg: I've put the whole thing in for now, but do our developers need to know the info either in all of the second sentence (The console is not...) or the second part (is a poor substitute)? It feels a little deep for users as opposed to internal devs?

kriskowal: We can omit “is a poor substitute” and beyond.

main/javascript-modifications/ses-guide.md
tyg: Changed last sentence to "You will not ever have to tear
down or delete one. " Is that OK?

tyg: Is each compartment automatically a separate vat?

main/javascript-modifications/ses-guide.md
tyg: Unless it's very long, my recollection is that most of Agoric Eng said they didn't like the compact/breakout style for docs (by "breakout" I think you mean "compact down to one line when you turn an arrow one way, fully expanding when you turn it the other" Did you mean something else?

main/javascript-modifications/ses-guide.md
tyg: Not sure what you mean? Some information's repeated here to make clear what the differences between lockdown and harden, but just enough to make clear how they are different and when to use one or the other?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+  to the bootstrap vat, which can share it with other vats)
+* `global`: Is not defined. Use `globalThis` instead (and remember that it is frozen).
+* `process`: Is not available, e.g. no `process.env` to access the process's environment variables, 
+  or `process.argv` for the argument array.
+* `URL` and `URLSearchParams`: Are not available.
+* `WebAssembly`: Is not available.
+* `TextEncoder` and `TextDecoder`: Are not available.
+
+Some names look like globals, but are really part of the module-defining tools: imports, exports, and 
+metadata. Modules start as files on disk, but then are bundled together into an archive before being
```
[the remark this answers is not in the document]
tyg: Changed to "Modules start as separate files, but then...".  OK?

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,505 @@
...
+removed before execution. 
+
+- `import` and `export` syntax are allowed in ESM-style modules ( preferred over CommonJS). These are 
+  not globals per se, but top-level syntax that defines the module graph.
+- `require`, `module`, `module.exports`, and `exports` are allowed in CommonJS-style modules, and 
+  should work as expected. However, new code should be written as ESM modules. They are either 
+  consumed by the bundling process, provided (in some form) by the execution environment, or 
+  otherwise rewritten to work sensibly
+- `__dirname` and `__filename` are not provided
+- The dynamic import expression (`await import('name')`) is currently prohibited in vat code,
```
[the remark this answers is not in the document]
tyg: Yep. Like I wrote, I need help in pulling out the non-SES changes to JS for elsewhere.

main/javascript-modifications/ses-guide.md
tyg: Previous instance changed from "should" to "must".

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,506 @@
...
+the needed SES features) for use in writing secure smart contracts in JavaScript 
+(Several Agoric engineers are on the relevant standards committees and are responsible
+for aspects of SES).
+
+## Installation and importing
+
+**tyg todo Not sure what the current situation is with respect to what needs to go into
+what code to get and use SES in one's code**
+
+## JavaScript globals and primordials
```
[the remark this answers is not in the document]
tyg: Changed to "What SES does to JavaScript". OK?

main/javascript-modifications/ses-guide.md
tyg: No, but shouldn't they know what's possible to do with a hardened object?

main/javascript-modifications/agoric-overview.md
```diff
@@ -100,32 +128,61 @@ There are two queues: the *IO queue* (accessed by `setImmediate`), and the *Prom
...
-
-- `import` and `export` syntax are allowed in ESM-style modules ( preferred over CommonJS). These are not globals per se, but top-level syntax that defines the module graph.
-- `require`, `module`, `module.exports`, and `exports` are allowed in CommonJS-style modules, and should work as expected. However, new code should be written as ESM modules. They are either consumed by the bundling process, provided (in some form) by the execution environment, or otherwise rewritten to work sensibly
+Some names look like globals, but are really part of the module-defining tools: imports, 
+exports, and metadata. Modules start as files on disk, but then are bundled together 
+into an archive before being loaded into a vat. The bundling tool uses several standard 
+functions to locate other modules that must be included. These are not a part of SES, but 
+are allowed in module source code, and are translated or removed before execution. 
+
+- `import` and `export` syntax are allowed in ESM-style modules ( preferred over CommonJS).
```
kriskowal, line 137:
```suggestion
- `import` and `export` syntax are allowed in ESM-style modules (preferred over CommonJS). 
```

main/javascript-modifications/agoric-overview.md
```diff
@@ -100,32 +128,61 @@ There are two queues: the *IO queue* (accessed by `setImmediate`), and the *Prom
...
-- `import` and `export` syntax are allowed in ESM-style modules ( preferred over CommonJS). These are not globals per se, but top-level syntax that defines the module graph.
-- `require`, `module`, `module.exports`, and `exports` are allowed in CommonJS-style modules, and should work as expected. However, new code should be written as ESM modules. They are either consumed by the bundling process, provided (in some form) by the execution environment, or otherwise rewritten to work sensibly
+Some names look like globals, but are really part of the module-defining tools: imports, 
+exports, and metadata. Modules start as files on disk, but then are bundled together 
+into an archive before being loaded into a vat. The bundling tool uses several standard 
+functions to locate other modules that must be included. These are not a part of SES, but 
+are allowed in module source code, and are translated or removed before execution. 
+
+- `import` and `export` syntax are allowed in ESM-style modules ( preferred over CommonJS). 
+  These are not globals per se, but top-level syntax that defines the module graph.
```
kriskowal, line 138: Do we avoid Latin as a rule?

main/javascript-modifications/agoric-overview.md
```diff
@@ -152,61 +211,97 @@ function foo(code) {
...
-The SES shim cannot correctly emulate a direct eval. If it tried, it would perform an indirect eval. This could be pretty confusing, because the code might not actually use objects from the local scope. You might not notice the problem until some later change altered the behavior.
+The SES shim cannot correctly emulate a direct eval. If it tried, it would perform 
+an indirect eval. This could be pretty confusing, because the code might not actually 
+use objects from the local scope. You might not notice the problem until some later 
+change altered the behavior.
 
-To avoid this confusion, the shim uses a regular expression to reject code that looks like it is performing a direct eval. This regexp is not complete (you can trick it into performing a direct eval anyway), but  that’s safe. Our goal is just to guide people away from confusing behaviors early in their development process.
+To avoid this confusion, the shim uses a regular expression to reject code that 
+looks like it is performing a direct eval. This regexp is not complete (you can 
+trick it into performing a direct eval anyway), but  that’s safe. Our goal is
```
kriskowal, line 233:
```suggestion
trick it into performing a direct eval anyway), but that’s safe. Our goal is 
```

main/javascript-modifications/agoric-overview.md
```diff
@@ -152,61 +211,97 @@ function foo(code) {
...
+and only then load any untrusted code into it.
 
 ### Frozen Primordials
 
-SES freezes *primordials*; built-in JavaScript objects such as `Object`, `Array`, and `RegExp`, and their prototype chains. This prevents malicious code from changing their behavior (imagine `Array.prototype.push` delivering a copy of its argument to an attacker, or ignoring certain values). It also prevents using, for example, `Object.heyBuddy` as an ambient communication channel.
+SES freezes *primordials*; built-in JavaScript objects such as `Object`, `Array`, 
+and `RegExp`, and their prototype chains. This prevents malicious code from 
+changing their behavior (imagine `Array.prototype.push` delivering a copy of 
+its argument to an attacker, or ignoring certain values). It also prevents 
+using, for example, `Object.heyBuddy` as an ambient communication channel.
```
kriskowal, line 258: `Object.heyBuddy` makes me laugh. Smells like @warner’s humor.

kriskowal (approved):

main/javascript-modifications/ses-guide.md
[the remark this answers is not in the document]
kriskowal: Perhaps “JavaScript’s lack of internal security”. JavaScript’s claim to fame is that it runs arbitrary code off the internet safely on behalf of a user, so it would be too harsh to say security is entirely absent. It’s in a league of its own compared to C, Python, or…Node.js.

main/javascript-modifications/ses-guide.md
kriskowal: Vats use compartments and can use multiple compartments. Compartments are not vats.

main/javascript-modifications/ses-guide.md
kriskowal: I’ll buy it, but it’s like “frozen turkeys are also a munition”. The fact might be more useful in the documentation of eventual send / handled promise in the Agoric environment.

main/javascript-modifications/ses-guide.md
kriskowal, line 364:
```suggestion
> Tip: If your text editor/IDE complains about `harden()` not being defined or imported, 
> try adding `/* global harden */` to the top of the file.
>
> You use `harden()` like this:
> ```js
> const o = {a: 2};
> o.a  = 12;
> console.log(o.a); // 12 because o is still mutable
> harden(o);
> o.a  = 37; // throws a TypeError because o is now hardened
> ```
```

See the preview for the effect. There would be no [disclosure button](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/disclosure-controls/).

main/javascript-modifications/agoric-overview.md
```diff
@@ -64,33 +69,56 @@ The vat environment has four significant objects not part of standard JavaScript
...
+  and give it to someone else (and don't immediately forget it yourself), 
+  you should give them `harden(obj)` instead of the raw object. 
+
+  You must harden a class before you harden any of its instances; i.e. it 
+  takes two separate steps to harden both a class and its instances. Harden 
+  a base class before hardening classes that inherit from it. `harden()` does 
+  transitive freezing by following the object’s own properties (as opposed to 
+  properties it inherited), and the objects whose own properties refer to 
+  them, and so forth.
+
```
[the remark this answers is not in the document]
tyg: So is it correct to say that "hardening a class also hardens its instances"?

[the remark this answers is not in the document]
kriskowal: No, the reverse. Hardening an instance hardens the class. This was not preciously the case

kriskowal (approved):

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,454 @@
...
+const c = new Compartment({
+  print: harden(console.log),
+});
+
+c.evaluate(`
+  print('Hello! Hello?');
+`);
+```
+This new compartment has a different global object than the start compartment **tyg todo: Not sure what "start compartment"
+means? The initial compartment? One from the Zoe start() method?** . The
```
kriskowal, line 228: We posit that all JavaScript executes in a realm and compartment. Every realm has distinct intrinsics, whereas every compartment shares intrinsics. The initial realm and compartment are not constructed with `new Realm()` or `new Compartment()` but that’s invisible to the code running: they could just as well be running within a constructed realm or compartment. We call the one compartment in a realm that was not expressly constructed the *start compartment*. The start compartment receives some ambient authorities from the host, often access to timers and IO that are denied to other compartments. Running lockdown does not erase these powerful objects, but puts the program running in the start compartment on a footing where it is possible to carefully delegate powers to child compartments.

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+When you create a new `Compartment` object, you must decide if it supports OCaps security.
+If it does, run `harden(compartment.globalThis)` on it before loading any untrusted code into it.
+
+A single compartment can run a JavaScript program in the locked-down SES environment.
+However, most interesting programs have multiple modules. So, each compartment also has
+its own module system. The next SES release will include support for ECMAScript modules,
+**tyg todo: Has this happened already? If not, is there an ETA?**
+a relatively new system supported by many browsers, and officially released in Node.js 14.
```
kriskowal, line 265:
```suggestion
its own module system. SES version 0.8.0 adds support for ECMAScript modules,
a relatively new system supported by many browsers, and officially released in Node.js 14.
```

main/javascript-modifications/ses-guide.md
```diff
@@ -0,0 +1,454 @@
...
+
+Compartments can be linked, so one compartment can export a module that another compartment
+imports. Each compartment may have its own rules for how to resolve import specifiers and 
+how to locate and retrieve modules. In this example, we use the compartment constructor to 
+create two compartments: one for the application and another for its dependency.
+
+**tyg todo: Will people understand this code or do we need to explain any/all methods in it
+in more detail**
```
kriskowal, line 273:
```suggestion
The `resolveHook` is synchronous and determines how to compute the full module specifier for a partially resolved module specifier in ESM source text, like `import "./even.js"` as it appears in `./math/odd.js` corresponds to `./math/even.js` in a Node.js program.
The `importHook` is asynchronous and responsible for for locating, retrieving, and parsing modules. Retrieving is getting the source text from the web, archive, or database based on its location. Converting a module specifier to a location is an internal concern of the `importHook` and the particular storage medium for the module texts but should generally be a URL and may appear in stack traces. The `importHook` may use the `ModuleStaticRecord` constructor to create a reusable, parsed representation of the module text.
```

main/javascript-modifications/ses-guide.md
kriskowal: Please let me know if you need further clarifications to settle this section.

main/javascript-modifications/ses-reference.md
```diff
@@ -0,0 +1,286 @@
...
+(preventing some man in the middle attacks). Also, no program can use them to pass notes to parties
+that haven't been expressly introduced (preventing some covert communication channels).
+
+Lockdown *freezes* all JavaScript defined objects accessible to any program in the realm. The frozen 
+accessible objects include but are not limited to: **tyg todo: Is there a short English description for
+the 2nd-4th items here?**
+- `globalThis`
+- `[].__proto__`
+- `{}.__proto__`
+- `(() => {}).__proto__ (async () => {}).__proto__`
```
kriskowal, line 43:
```suggestion
- `(() => {}).__proto__`
- `(async () => {}).__proto__`
```

main/javascript-modifications/ses-reference.md
```diff
@@ -0,0 +1,286 @@
...
+
+## `lockdown()`
+
+`lockdown()` tamper-proofs all of the JavaScript intrinsics, so no program can subvert their methods
+(preventing some man in the middle attacks). Also, no program can use them to pass notes to parties
+that haven't been expressly introduced (preventing some covert communication channels).
+
+Lockdown *freezes* all JavaScript defined objects accessible to any program in the realm. The frozen 
+accessible objects include but are not limited to: **tyg todo: Is there a short English description for
+the 2nd-4th items here?**
```
kriskowal, line 39: - `[].__proto__` is the array prototype, equivalent to `Array.prototype` in a pristine JavaScript environment.
- `{}.__proto__` is `Object.prototype` 〃
- `(() => {}).__proto__` is `Function.prototype` 〃
- `(async () => {}).__proto__` is the prototype of all asynchronous functions and has no alias in the global scope of a pristine JavaScript environment.

main/javascript-modifications/ses-reference.md
```diff
@@ -0,0 +1,286 @@
...
+lockdown({ stackFiltering: 'concise' }); // Preserve important deep stack info
+// vs
+lockdown({ stackFiltering: 'verbose' }); // Console shows full deep stacks
+```
+
+## `overrideTaming` Options
+
+**tyg todo: Honestly not fond of this. If you can do a better job of giving a very brief
+summary of what the settings just do without having to explain a lot of background, please
+do.**
```
kriskowal, line 261:
```suggestion
We recommend explicitly selecting `overrideTaming` of `min` when beginning a project because debugging will be more pleasant. You may need to relax it back to the default of `moderate` third-party shimming code interferes with `lockdown`. By way of background, these are two treatments to address a language design mistake that can interfere with the SES shim.
```

I defer to @erights to verify the correctness of this summary or propose a better alternative.

kriskowal (commented):
This review covers all of the `**tyg` blocks I could find. Please as always let me know if I missed a question!

main/javascript-modifications/lockdown.md
```diff
@@ -0,0 +1,546 @@
...
+
+**tyg todo: I'm not finding a good definition for what the "concise" value does.
+In the sample code, the comment is "Preserve important deep stack info". Does
+this mean it should be something like "The 'concise' value causes stack traces to
+only show the most important distributed computation related values/information.
+[And what sort of things is that?]."?**
+
+Either `stackFiltering` setting is safe. Stack information will
+or will not be available from error objects according to the `errorTaming`
+option and the platform error behavior.
```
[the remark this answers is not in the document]
erights: @kriskowal your description fits `errorTaming: 'safe'` vs `'unsafe'`. The `stackFiltering: 'concise'` vs `'verbose'` is about trying to reduce the noise appearing on the console's reported stack traces. `'concise'` reduces the noise but at the risk of throwing out the signal you were really looking for, which is why it's a configurable option.

At the bottom of the https://github.com/endojs/endo/blob/992f35fdf4679faa8fb7d1c7e265c6b654156399/packages/ses/lockdown-options.md#stackfiltering-options section you'll see two triangles for

Expand for { stackFiltering: 'concise' } log output

and

Expand for { stackFiltering: 'verbose' } log output

expand them to see an example of the difference this option makes.

main/javascript-modifications/lockdown.md
```diff
@@ -0,0 +1,543 @@
...
+All of the "locale methods" have, by design, global behavior not fully determined by the
+JavaScript spec. Their behavior varies with location and culture. 
+
+However, by placing this information of shared primordial prototypes, **(tyg todo: I'm not parsing the
+previous phrase. What does "placing this information of shared prototypes" mean, or is there a typo
+or missing text in there?)** 
+it cannot differ per Comparment. So one Compartment cannot virtualize the locale for code
+running in another Compartment. Worse, on some engines the methods' behavior may change at runtime as 
+the machine is "moved" between different locales, i.e., if the operating system's locale is
+reconfigured while JavaScript code is running.
```
kriskowal, line 211:
```suggestion
The behavior of all of these methods varies depending on the user's locale and can change while a program is running. This breaks determinism since a program run in one compartment might not have the same result in another compartment or at another time. Revealing the user's locale also helps unscrupulous programs to track even identify the user without their consent.

The behavior of methods like `toLocaleString` might be acceptable if they could be governed on a per-compartment basis ("virtualized"), but because these methods appear on the prototype of shared intrinsic objects like the String prototype, there is no safe alternative design.

Instead, the hosting program can choose to reveal the locale to all compartments by setting `localeTaming` to `unsafe`, or inject a `locale` object into selected compartments with the powers of these methods.
```

PR #439, opened 2021-02-16
DRAFT: Still erroring with `Cannot initialize @agoric/assert, missing globalThis.assert`

katelynsills:
Still erroring with `Cannot initialize @agoric/assert, missing globalThis.assert`

To replicate:

Make sure your agoric-sdk is on master
within agoric-sdk, `yarn install`
within agoric-sdk, `yarn build`
in documentation `agoric install`
in documentation `agoric test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "eslint-plugin-jsdoc": "^30.4.2",
     "eslint-plugin-prettier": "^3.1.2",
     "prettier": "^1.18.2",
-    "ses": "^0.10.1",
+    "ses": "^0.12.2",
     "vuepress": "^1.5.4",
     "vuepress-plugin-check-md": "0.0.2"
   },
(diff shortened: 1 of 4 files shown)


dckc:
you get that error in response to what?

`yarn docs:dev` seems to work for me.

`yarn test` gives me very different errors... `Cannot find module '@agoric/eventual-send/shim'`

katelynsills:
Replied elsewhere, but for anyone following, I think this error (`yarn test gives me very different errors... Cannot find module '@agoric/eventual-send/shim'`) is fixed by `agoric install`

dckc:
for the full log of diagnostics, see https://gist.github.com/dckc/4ccfb54128f0b3fe0bc2956c1e4e9e28

oh... and node 15 might be necessary to reproduce this

PR #440, opened 2021-02-17
chore: fix tests. Works under Node 14

katelynsills:
@michaelfig, I believe this should fix any issues you were seeing in https://github.com/Agoric/agoric-sdk/pull/2441#issuecomment-780113461. 

This does not work under Node 15, but I believe our CI doesn't use that yet.


Chris-Hibbert (approved):
looks good!

PR #441, opened 2021-02-19
chore: update the lockfile

katelynsills:
I believe this fixes the [Node 15 issues](https://github.com/Agoric/agoric-sdk/issues/2454) we've been seeing.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,20 +2,17 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


dckc (approved):
bingo!

```
agoric/documentation$ git checkout update-lockfile
agoric/documentation$ agoric install
agoric/documentation$ yarn test
...
  37 tests passed
Done in 8.51s.
```

PR #443, opened 2021-02-22
chore: update for Nat 4.0.0

katelynsills:
This PR updates the documentation for agoric-sdk using Nat 4.0.0 (https://github.com/Agoric/agoric-sdk/pull/2475).

Things changed in https://github.com/Agoric/agoric-sdk/pull/2475:
* Nat values such as `moola3` are [BigInts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) rather than Numbers: `{ brand, value: 3n}` 
* Expiration dates and deadlines are also BigInts. E.g:
```
const myProposal = harden({
  give: { Asset: quatloos(4)},
  want: { Price: moola(15) },
  exit: { afterDeadline: {
    timer,
    deadline: 100n,
  }}
})
```

Still to do:
- [ ] Although the snippet tests have been updated, we should check any code that is written as text for the above changes.
- [ ] An short upgrade guide for anyone who has already made a dapp
- [ ] An explanation of what BigInts are, their benefits, and how to use them. (A good start is here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)

#agoric-sdk-branch: 325-nat-4.0.0-zoe

### DIFF
--- a/main/distributed-programming.md
+++ b/main/distributed-programming.md
@@ -3,6 +3,10 @@
 Agoric's platform lets you write secure smart contracts in JavaScript. The platform itself is mainly written in JavaScript. However, we've made several Agoric-specific additions to general JavaScript programming that you should know about and understand before programming on the platform. Some are *concepts*, others are *Agoric library additions*, and some are at the *syntax level*. All changes at the language level are in process to become official standards.
 
 Extensions covered in this document are:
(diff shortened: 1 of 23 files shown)


katelynsills (commented):
Looks good. Thanks @tyg

Issue #426, opened 2021-02-02
Fix lack of auto-generated Sidebar menu for various docs.


tyg:
Several pages should have an auto-generated sidebar menu with items equating to h2 and h3 headers in the source. They don't, and this is to fix them.

The pages mentioned here have no individual page Sidebar menu at all.

- [X] Glossary https://agoric.com/documentation/glossary/
Fixed in : https://github.com/Agoric/documentation/pull/444
- [X] Platform  https://agoric.com/documentation/platform/
Fixed in : https://github.com/Agoric/documentation/pull/444

Trickier ones

- [x] https://agoric.com/documentation/getting-started/agoric-cli-guide.html
- [x] https://agoric.com/documentation/dapps/
- [x] https://agoric.com/documentation/wallet-api.html
- [x] https://agoric.com/documentation/distributed-programming.html

Other changes:
- [x] Add H3 headers to https://agoric.com/documentation/getting-started/agoric-cli-guide.html 

In most of these cases, plus the REPL directory, these are in separate files at the documentation/main/ top level.  As part of the fix/cleanup, I wanted to move these (plus their associated files from main/assets) into a new main/guides directory, then establish a new    '/guides/': [. ... ]   entry in .vuepress/config.js under sidebar: { ... } using their new paths.

PR #446, opened 2021-02-25
chore: show two ways of adding config to show a sidebar

katelynsills:
@tyg, I think this should take care of #426 if you use these methods for any other pages that need sidebars. I'll let you take this PR from here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,7 @@
+---
+sidebar: auto
+---
+
 # Contributing to Agoric's Documentation Repo
 
 Agoric's public-facing technical documentation is mostly in the [Agoric Documentation GitHub repo](https://github.com/Agoric/documentation).
@@ -78,6 +82,14 @@ Lines with no special treatment are converted into standard HTML paragraph tags.
 VuePress automatically builds search functionality and individual page menus from `h1`, `h2`, and `h3` headers (i.e. Markdown's `#`, `##`, and `###` commands). 
 You must have only **one** `h1` per `.md` file. Be careful not to have too many `h2` and `h3` level headers 
 on one page and that they aren't too long. Otherwise the sidebar menu for the page will be cluttered and hard to read and use.
+
+There's a glitch (or something we don't understand) that sometimes causes individual pages not to display a sidebar menu for their headers.
+To force an individual page sidebar menu, add the following YAML at the top of a page's source file (this file has this YAML at the top):
+```js
+---
+sidebar: auto
+---
+```
 
 ## Writing Links
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,15 +80,52 @@ Lines with no special treatment are converted into standard HTML paragraph tags.
 ### Sectioning Pages
 
 VuePress automatically builds search functionality and individual page menus from `h1`, `h2`, and `h3` headers (i.e. Markdown's `#`, `##`, and `###` commands). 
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -79,6 +83,14 @@ VuePress automatically builds search functionality and individual page menus fro
...
 on one page and that they aren't too long. Otherwise the sidebar menu for the page will be cluttered and hard to read and use.
 
+There's a glitch (or something we don't understand) that sometimes causes individual pages not to display a sidebar menu for their headers.
+To force an individual page sidebar menu, add the following YAML at the top of a page's source file (this file has this YAML at the top):
+```js
+---
+sidebar: auto
+---
+```
+
```
katelynsills, line 93: No, this is not a glitch. To show a sidebar, you must either put it in the config or add that to the top. It will not show up unless we do that. Is there a reason to think it's a glitch, rather than the rules?

katelynsills (commented):
Looks good other than the explanation of it being a glitch. There's no glitch that I know of. It works if we follow the rules listed in the Vuepress documentation

CONTRIBUTING.md
tyg: From the [VuePress manual](https://vuepress.vuejs.org/theme/default-theme-config.html#nested-header-links)

"The sidebar automatically displays links for headers in the current active page, nested under the link for the page itself."

This seems to say that we shouldn't have had to do anything for the individual page sidebars header menus to show up. That we did would seem to be a glitch.

katelynsills: Ah, I can see why it would seem to imply that it automatically happens without more configuration, but that isn't true. There's no technical glitch; it's that their documentation isn't as clear as it could be. For the record, how VuePress works is that by default, no sidebars appear. Sidebars will appear if you 

1) turn it on for the whole site (note that VuePress includes this in their docs, showing that it does NOT happen by default):
```
// .vuepress/config.js
module.exports = {
  themeConfig: {
    sidebar: 'auto'
  }
}
```
2) Turn it on for a single page by adding YAML to the top of that page:

```
---
sidebar: auto
---
```

or 3) Add a config entry, either as part of a group or as an individual page.

```
// .vuepress/config.js
module.exports = {
  themeConfig: {
    sidebar: [
      {
        title: 'Group 1',   // required
        path: '/foo/',      // optional, link of the title, which should be an absolute path and must exist
        collapsable: false, // optional, defaults to true
        sidebarDepth: 1,    // optional, defaults to 1
        children: [
          '/'
        ]
      },
      {
        title: 'Group 2',
        children: [ /* ... */ ],
        initialOpenGroupIndex: -1 // optional, defaults to 0, defines the index of initially opened subgroup
      }
    ]
  }
}
```

So, in summary, let's make sure to remove any mention of a glitch. There is no glitch. Instead, let's include the above information in CONTRIBUTING.md, and if we have time, file a PR on VuePress documentation to make sure it's clear for the next person.

katelynsills: Here's a PR to the VuePress docs. https://github.com/vuejs/vuepress/pull/2805 Let's see what they say!

katelynsills:
Just added a commit to turn all the sidebars on by default.

katelynsills:
@tyg can we take out the part about a glitch? Once we do that, I think we can merge this

tyg:
> @tyg can we take out the part about a glitch? Once we do that, I think we can merge this

Done.

katelynsills (commented):
LGTM!

PR #454, opened 2021-03-15
Allocations

tyg:
Updates remaining allocation getters to note that post-exit they return the allocation values at exit.


katelynsills (approved):
LGTM! Thank you!

PR #457, opened 2021-03-23
Sidebar patch 21 03 21

tyg:
This cleans up the sidebars to reflect, as best as is possible, overall navigation when on a page, particularly SES and the other JavaScript related docs.

As I'm understanding the VuePress docs, there's a fundamental issue that sidebars are pretty tightly coupled with folder and file organization in the repo. While one can include a file in more than one sidebar, when that file is displayed, only the sidebar that matches its location in the doc repo file structure is shown. For example, while Documentation Guide appears in both the Getting Started and Guides sidebars, if you click on it in either one and display it, it will always show the Getting Started sidebar since the documentation-guide.md file is in the getting-started folder, and not the guides folder. We have issues because it's very possible for our files to "belong" in more than one place. For example, the ERTP and Zoe introductions belong to both the Getting Started "track" and their respective ERTP and Zoe project track. The main fallout right now is that since it's /main/repl/ instead of /main/guides/repl, when anything in the repl folder is displayed, its sidebar is limited to only the repl folder contents. This potentially could be fixed by moving the repl folder from /main/repl to /main/guides/repl. Chris has written he's fine either way with this. 

Note that while working on this, I realized that somehow between merging the SES branch and the reorg branch, a new agoric-js-overview.md file had managed to vanish. I went ahead and used this branch/PR to get it back in.

### DIFF
--- a/main/.vuepress/config.js
+++ b/main/.vuepress/config.js
@@ -78,6 +78,30 @@ module.exports = {
             '/guides/agoric-cli/commands',
             '/guides/agoric-cli/starting-multiuser-dapps',
           ]
+        }
+        {
+          title: 'ERTP Guide',
+          path: '/ertp/guide/',
+          collapsible: false,
+          children: [
+            '/ertp/guide/',
+            '/ertp/guide/amounts',
+            '/ertp/guide/amount-math',
+            '/ertp/guide/issuers-and-mints',
+            '/ertp/guide/purses-and-payments',
+          ]
+        },
+        {
+          title: 'Zoe Guide',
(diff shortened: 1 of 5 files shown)


katelynsills (approved):
These changes look fine. Re: how sidebars work, my guess is that if a page is mentioned more than once in a sidebar, when displaying the page, the code either finds the first mention in the sidebar config, or the last mention in the sidebar config. I don't think it has anything to do with the file location, but this should be easy to test. Bottom line is, I don't think we need to move files around to change which sidebar appears on a page.

PR #467, opened 2021-04-05
fix: update to eslint 7.23.0, ses 0.12.7

erights:


erights:
We've decided not to move forward with these at this time.

PR #482, opened 2021-04-15
MutableQuote methods

tyg:


main/zoe/guide/price-authority.md
```diff
@@ -46,6 +46,11 @@ This means that the `priceAuthority` asserts that when `timestamp` according to
 `amountIn` and `amountOut` are ERTP amounts for the `brandIn` and `brandOut` you
 requested.
 
+## Mutable price quotes
+
+A `MutableQuote` object is the same as a `PriceQuote`, except that you can cancel it or
+change its trigger levels to specified values without requiring a second `Promise`.
```
Chris-Hibbert, line 52: A `PriceQuote` is a record (has fields rather than methods), while `MutableQuote` is an object (methods, no fields).  `MutableQuote`'s method `getPromise()` returns a promise for a `PriceQuote`, which is the same Promise returned by `quoteWhenLTE()` and its siblings.

main/zoe/api/contract-support/price-authority.md
```diff
@@ -136,3 +136,68 @@ const quote = E(pAuthority).quoteWhenLTE(
     amountMath.make(brands.Out, 974n),
   );
 ```
+## `MutableQuote` object
+
+**tyg todo: When do you want to use a mutable quote vs. a regular quote, and vice versa?**
```
Chris-Hibbert, line 141: Use a MutableQuote when you expect to make multiple calls, replacing the trigger value. If you just need a single quote, and won't change the trigger level, then quoteWhen will work fine.

main/zoe/api/contract-support/price-authority.md
```diff
@@ -136,3 +136,68 @@ const quote = E(pAuthority).quoteWhenLTE(
...
+
+There are four *mutable quote* methods, which return a `MutableQuote` object with the methods:
+- `cancel(e)`
+  - `e``{ String }`
+  - Causes the `Promise` to reject with the message `e`.  **tyg todo: How is the message used; where does it go?**
+- `getPromise()`
+  - Returns the same `Promise` as the `quoteWhen...()` methods.
```
Chris-Hibbert, line 148:
```suggestion
  - Returns: `{ Promise<PriceQuote> }`
```

main/zoe/api/contract-support/price-authority.md
```diff
@@ -136,3 +136,68 @@ const quote = E(pAuthority).quoteWhenLTE(
...
+- `updateLevel(newAmountIn, newAmountOutLimit)`
+  - `newAmountIn` `{ Amount }`
+  - `newAmountOutLimit` `{ Amount }`
```
Chris-Hibbert, line 151:
```suggestion
  - `newAmountIn` `{ Amount }`; The brand must match the original `amountIn`
  - `newAmountOutLimit` `{ Amount }`; The brand must match the original `amountOutLimit`
```

main/zoe/api/contract-support/price-authority.md
```diff
@@ -136,3 +136,68 @@ const quote = E(pAuthority).quoteWhenLTE(
...
   );
 ```
+## `MutableQuote` object
+
+**tyg todo: When do you want to use a mutable quote vs. a regular quote, and vice versa?**
+
+There are four *mutable quote* methods, which return a `MutableQuote` object with the methods:
+- `cancel(e)`
+  - `e``{ String }`
+  - Causes the `Promise` to reject with the message `e`.  **tyg todo: How is the message used; where does it go?**
```
Chris-Hibbert, line 146: Hmm. There's nothing in our doc about how to use promises, other than a reference to javascript.info's docs. That'll have to do, I guess. When someone makes use of the promise with a `.then()`, they'll be notified of rejections. That's where it is received.

main/zoe/api/contract-support/price-authority.md
```diff
@@ -136,3 +136,73 @@ const quote = E(pAuthority).quoteWhenLTE(
...
+## `MutableQuote` object
+
+Use a `MutableQuote` when you expect to make multiple calls, replacing the trigger
+value. If you just need a single quote, and won't change the trigger level, then use
+a non-mutable quote.
+
+There are four *mutable quote* methods, which return a `MutableQuote` object with the methods:
+- `cancel(e)`
+  - `e``{ String }`
+  - Causes the `Promise` to reject with the message `e`. When the promise is used with a `.then()`
```
Chris-Hibbert, line 148: @erights said yesterday that we should pervasively replace .then with `E.when()`. Might as well start here.

main/zoe/guide/price-authority.md
```diff
@@ -46,6 +46,16 @@ This means that the `priceAuthority` asserts that when `timestamp` according to
 `amountIn` and `amountOut` are ERTP amounts for the `brandIn` and `brandOut` you
 requested.
 
+## Mutable price quotes
+
+A `PriceQuote` is a record (has fields rather than methods), while `MutableQuote` is
+an object (methods, no fields). `MutableQuote`'s method `getPromise(`) returns 
+a `Promise` for a `PriceQuote`, which is the same `Promise` returned by the `quoteWhenLTE()`
+API method and its siblings.
```
Chris-Hibbert, line 54: Sorry, these words were me explaining to you, not me providing wording for the doc. I don't think the comparison to `PriceQuote` as a record belongs here.
```suggestion
A `MutableQuote` is an object. `MutableQuote`'s method `getPromise(`) returns 
a `Promise` for a `PriceQuote`, which serves the same purpose as the return value from `quoteWhenLTE()` etc. 
```

tyg: I dunno. I think this might be significant enough to be worth including to make sure they realize this, as it affects how they think of/use the two? At the very least, I don't see any harm in including it, and it might do some good?

Chris-Hibbert: The description of priceQuote as a record is redundant with the paragraph just above: "Receiving a quote".  I'd prefer "is a record with an amount and a payment" over " is an amount, payment pair".  This section is about MutableQuotes, which are a different way of getting a priceQuote. It might make more sense to say that you can use quoteWhenGT and its ilk to get a priceQuote directly, or mutableQuoteWhenGT to get a reusable object.

katelynsills:
I'll take myself off the review path for this one, as Chris seems to have it handled and I'm not as familiar. Thanks for including me.

main/zoe/guide/price-authority.md
```diff
@@ -48,13 +48,11 @@ requested.
...
-a `Promise` for a `PriceQuote`, which is the same `Promise` returned by the `quoteWhenLTE()`
-API method and its siblings. 
-
-You can cancel a `MutableQuote` or change its trigger levels to specified values without
-requiring a second `Promise`. 
+`MutableQuote`'s method `getPromise(`) returns a `Promise` for a `PriceQuote`, 
+which is the same `Promise` returned by the `quoteWhenLTE()` API method and its siblings. 
+Effectively, the non-mutable price quote methods return a static `PriceQuote`, while
+the mutable price quote methods return a reusable price quote object which can manipulated 
+by changing its trigger levels or cancelling it.
```
Chris-Hibbert, line 55: `MutableQuote`'s method `getPromise()` returns a Promise for a `PriceQuote`,  which is the same as what is returned by the `quoteWhenLTE()` API method and variants.  Effectively, the non-mutable price quote methods return a single `PriceQuote`, while the mutable price quote methods return a reusable object which can be manipulated by changing its trigger levels or by cancelling it.


The backquote after getPromise shouldn't be inside the parens.

Chris-Hibbert (approved):

Chris-Hibbert:
One more small change

PR #483, opened 2021-04-16
Far() documentation

tyg:
Please note this is written for external developers and as much as possible covers how and when they should use far(). It does not go into detail about how far() and Remotables interact with presences and promises and the like, as it does not appear to me external devs need to know in depth what we're doing with Remotables "on the other side" as it were. If you think external devs do need to know this sort of in depth implementation, please let me know in comments. Thanks.

### DIFF
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -15,7 +15,7 @@ To call a method on an object from another vat or machine, you must
 use [`E()`](./eventual-send.html#remote-object-communication-with-e). For example, getting
 an `Issuer`'s `brand` would look like `E(issuer).getBrand()`.
 
-Objects intended to be used in other vats are called *remotables*. Remote messages sent to
+Objects intended to be used from other vats are called *remotables*. Remote messages sent to
 remotables must only contain *passable* arguments and return *passable* results. 
 Passables includes all things that can be passed as arguments in messages. 
 
@@ -31,8 +31,6 @@ be passable.
 ### Rules for creating remotables
 - All property values must be functions. 
   - They cannot be accessors.
-  - Note: If you wish to send data, send pass-by-copy data such as pass-by-copy records, 
-    pass-by-copy arrays, strings, numbers, etc. or some other passable.
 - You must wrap the object with `Far()`.
 
 **Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`, as are
@@ -50,12 +48,12 @@ be passable.
 -  Returns: A `Remotable` object.
 
 The `farName` parameter gives the `Remotable` an *interface name* for debugging purposes, which only shows
-up when logged through `console.log`. 
+up when logged through the `console`, for example with `console.log`. 
 
-The `object-with-methods` parameter includes a record with definitions of all the object's 
-property functions. See the example code below.
+The `object-with-methods` parameter should be an object whose properties are the functions serving 
+as the object's methods. See the example code below.
 
-`Far()` function marks an object as remotable.  `Far()` also:
+The `Far()` function marks an object as remotable.  `Far()` also:
 - Hardens the object.
   - Both `harden()` and `Far()` function harden the object. 
   - Only hardened objects are passable.
@@ -69,11 +67,11 @@ You should call `Far()` on an object if it both:
     you should run `Far()` on it after creating it.
 - Has methods called on it, as opposed to just effectively storing data.
 
-As defined, only passables can be either passed as arguments or returned as results,
+Only passables can be passed as arguments or returned as results,
 and they must be hardened. If the passable is a remotable, it must be hardened with `Far()`.
 Otherwise, it must be hardened with [`harden()`](./ses/ses-guide.md#harden).
 
-There's no harm in using `Far()` on an object, even if it never leaves its vat. An error
+An error
 is thrown if you call `Far()` on a record, instead of an object, which doesn't have function
 values. However, if object `foo` should never be exposed to other vats, you should make it
 a point **not** to use `Far()` on it. If `foo` is not marked as `Remotable` but is accidentally
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -1,7 +1,7 @@
 
 # `Far()` and Remotable Objects
 
-## Remotable objects
+## Remotable and passable objects
 
 In Agoric smart contracts and dapps, you can call methods on objects from other vats or machines. 
 For example, a purse for an ERTP issuer actually lives in the issuer's vat. But code in your off-chain
@@ -15,23 +15,31 @@ To call a method on an object from another vat or machine, you must
 use [`E()`](./eventual-send.html#remote-object-communication-with-e). For example, getting
 an `Issuer`'s `brand` would look like `E(issuer).getBrand()`.
 
-Objects intended to be used in other vats are called `Remotables`. In particular, note that
-every object returned from a smart contract, such a `publicFacet` or 
-`creatorFacet`, must be `Remotable`. All objects used in your contract's external API must
-be `Remotable`
+Objects intended to be used in other vats are called *remotables*. Remote messages sent to
+remotables must only contain *passable* arguments and return *passable* results. 
+Passables includes all things that can be passed as arguments in messages. 
 
-**Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`. You do 
-not need to do anything to make them remotable; i.e. you do not need to call `Far()` on an ERTP object.
-**tyg todo: Are UserSeat and ZCFSeat both also automatically created as `Remotable`?**
+There are three kinds of passables:
+   * Remotables, objects with methods that can be called remotely using `E()`.
+   * Pass-by-copy data, such as numbers or hardened records
+   * Promises for passables.
 
-### Rules for Creating Remotables
+In particular, note that every object returned from a smart contract, such a `publicFacet` or 
+`creatorFacet`, must be passable. All objects used in your contract's external API must
+be passable.
+
+### Rules for creating remotables
 - All property values must be functions. 
   - They cannot be accessors.
-  - Note: If you wish to send data, send a pass-by-copy record instead, or add a function that returns a pass-by-copy record.
+  - Note: If you wish to send data, send pass-by-copy data such as pass-by-copy records, 
+    pass-by-copy arrays, strings, numbers, etc. or some other passable.
 - You must wrap the object with `Far()`.
 
-### Using Remotables
-- Call a `Remotable`'s method by first wrapping the `Remotable` object with `E`, such as `E(issuer).getBrand();`
+**Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`, as are
+`UserSeats` and `ZCFSeats`. 
+
+### Using remotables
+- Call a remotable's method by first wrapping the remotable object with `E`, such as `E(issuer).getBrand();`
 - Handle the resulting promise. Calling `E()` always results in a `Promise`.
 
 ## Using `Far()`
@@ -48,9 +56,11 @@ The `object-with-methods` parameter includes a record with definitions of all th
 property functions. See the example code below.
 
 `Far()` function marks an object as remotable.  `Far()` also:
-- Runs `harden()` on the object.
-- Checks for the property 
-and value requirements above. If they are not met, it throws an error.
+- Hardens the object.
+  - Both `harden()` and `Far()` function harden the object. 
+  - Only hardened objects are passable.
+- Checks for the property and value requirements above. 
+  If they are not met, it throws an error.
 - Records the object's interface name. 
 
 You should call `Far()` on an object if it both:
@@ -59,10 +69,9 @@ You should call `Far()` on an object if it both:
     you should run `Far()` on it after creating it.
 - Has methods called on it, as opposed to just effectively storing data.
 
-`Far()` automatically [hardens](./ses/ses-guide.md#harden) its object argument. 
-If you make a `Remotable` object with `Far()`, you don't need to also call `harden()` 
-on it. Since `Far()` is not used on objects used to send data, you must still use
-`harden()` on them.
+As defined, only passables can be either passed as arguments or returned as results,
+and they must be hardened. If the passable is a remotable, it must be hardened with `Far()`.
+Otherwise, it must be hardened with [`harden()`](./ses/ses-guide.md#harden).
 
 There's no harm in using `Far()` on an object, even if it never leaves its vat. An error
 is thrown if you call `Far()` on a record, instead of an object, which doesn't have function
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -21,8 +21,6 @@ every object returned from a smart contract, such a `publicFacet` or
 be `Remotable`
 
 ### Rules for Creating Remotables
-- All property names must be strings. 
-  - Property names must not be `Symbols`. **tyg todo: Do we need to have the `Symbol.asyncIterator` exception here?**
 - All property values must be functions. 
   - They cannot be accessors.
   - Note: If you wish to send data, send a pass-by-copy record instead, or add a function that returns a pass-by-copy record.
@@ -40,7 +38,7 @@ be `Remotable`
 -  Returns: A `Remotable` object.
 
 The `interface-name` parameter gives the `Remotable` an *interface name*, which only shows
-up in `console.log` when 
+up when logged through `console.log`. 
 
 The `object-with-methods` parameter includes a record with definitions of all the object's 
 property functions. See the example code below.
@@ -50,10 +48,10 @@ Use the `Far()` function to mark an object as `Remotable`.  `Far()` also:
 - Checks for the property name and value requirements above. If they are not met, it throws an error.
 - Records the object's interface name. 
 
-You should call `far()` on an object if it both:
+You should call `Far()` on an object if it both:
 - Will be sent out of its native vat.
   - If it might ever appear as the `foo` in [`E(foo)`](./eventual-send.md),  
-    you should run `far()` on it after creating it.
+    you should run `Far()` on it after creating it.
 - Has methods called on it, as opposed to just effectively storing data.
 
 ```js
@@ -68,7 +66,7 @@ const countRemotable = Far('counter', {
 
 `Far()` automatically [hardens](./ses/ses-guide.md#harden) its object argument. 
 If you make a `Remotable` object with `Far()`, you don't need to also call `harden()` 
-on it. Since `far()` is not used on objects used to send data, you must still use
+on it. Since `Far()` is not used on objects used to send data, you must still use
 `harden()` on them.
 
 
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -74,7 +74,7 @@ Otherwise, it must be hardened with [`harden()`](./ses/ses-guide.md#harden).
 An error
 is thrown if you call `Far()` on a record, instead of an object, which doesn't have function
 values. However, if object `foo` should never be exposed to other vats, you should make it
-a point **not** to use `Far()` on it. If `foo` is not marked as `Remotable` but is accidentally
+a point **not** to use `Far()` on it. If `foo` is not marked as a remotable but is accidentally
 exposed, an error is thrown. This prevents any vulnerability from such accidental exposure.
 
 ```js
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -20,6 +20,10 @@ every object returned from a smart contract, such a `publicFacet` or
 `creatorFacet`, must be `Remotable`. All objects used in your contract's external API must
 be `Remotable`
 
+**Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`. You do 
+not need to do anything to make them remotable; i.e. you do not need to call `Far()` on an ERTP object.
+**tyg todo: Are UserSeat and ZCFSeat both also automatically created as `Remotable`?**
+
 ### Rules for Creating Remotables
 - All property values must be functions. 
   - They cannot be accessors.
@@ -43,7 +47,7 @@ up when logged through `console.log`.
 The `object-with-methods` parameter includes a record with definitions of all the object's 
 property functions. See the example code below.
 
-Use the `Far()` function to mark an object as `Remotable`.  `Far()` also:
+`Far()` function marks an object as `Remotable`.  `Far()` also:
 - Runs `harden()` on the object.
 - Checks for the property name and value requirements above. If they are not met, it throws an error.
 - Records the object's interface name. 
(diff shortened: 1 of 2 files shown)


main/guides/js-programming/far.md
```diff
@@ -0,0 +1,73 @@
+# `Far()` and Remotable Objects
+
+## Remotable objects
+
+Some objects have methods that can be invoked remotely (across contracts and vats) using
```
katelynsills, line 5: I think as written, this is very confusing for an outsider. I think we need to back up and (if necessary) dial down the precision in order to achieve something understandable. 

How about starting out with something like the following instead, corrected for any mistakes:

---
In Agoric smart contracts and dapps, you can call methods on objects that originate in another vat or on another machine. For example, a purse for an ERTP issuer actually lives in the vat in which the issuer was created, but you can use the purse in your off-chain wallet or in another contract. In other words, Agoric encapsulates all of the inter-machine and inter-vat communication so that at the level of a smart contract or dapp, objects from other vats or machines can be treated almost  as if they are local. 

In order to call a method on an object from another vat or other machine, you must use `E()`, such as `E(issuer).getBrand()`. 

Objects (like ERTP issuers) which are intended to be used in other vats are called `Remotables`. Importantly, any object returned from a smart contract, such an `offerResult`, `publicFacet` or `creatorFacet`, must be `Remotable`.

## Rules for Creating Remotables
1. All property names must be strings. Property names must not be Symbols.
3. All property values must be functions. (Note: If you wish to send data, send a pass-by-copy record instead, or add a function that returns a pass-by-copy record.)
4. The object must be wrapped with `Far`. (Note: `Far` also hardens the object.)

## Rules for Using Remotables
1. Call the method by first wrapping the object with `E`, such as `E(issuer).getBrand()`;
2. Handle the resulting promise. Calling `E()` always results in a promise.

## How to use Far to Correctly Wrap a Remotable
...

katelynsills (changes requested):
I think we need to make this more accessible to the audience of developers who will be reading this. Suggestions in comments.

main/guides/js-programming/far.md
erights: I like @katelynsills ' version.

@katelynsills ' one thing seems off. An `offerResult` just needs to be passable. I don't see and reason it needs to be remotable. This is true for the `publicFacet` and `creatorFacet` as well, but less important since they are much more likely to be remotables in practice.

@tyg Do we have a glossary somewhere where we try to explain our terminology precisely? For a lot of what I explained on that github thread, I think @katelynsills is probably right to omit them here. But these terms should be explained precisely somewhere.

katelynsills: > One thing seems off. An offerResult just needs to be passable. I don't see any reason it needs to be remotable. This is true for the publicFacet and creatorFacet as well, but less important since they are much more likely to be remotables in practice.

Oh, good point. That's right. The `offerResult` could be a pass-by-copy object or not an object at all.

tyg: @erights We have the general glossary (visible at https://agoric.com/documentation/glossary/, in GitHub at documentation/main/glossary, but I'm not sure that syncs up with what you're looking for; it consists of relatively short definitions intended more to give users an initial idea of what we're talking about. Your sort of thing is more like the deep dives Allison's using in your Foresight book and are fairly detailed and precise discussions as much as definitions, IMO.

warner: Would a Purse, Payment, or Offer be a better example of a Remotable? In particular, are they the most-well-represented objects in our docs and sample code?

I think the `publicFacet` and `creatorFacet` will generally *not* be Remotable, since they'll have all sorts of random data as properties.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Using `Far()`
+
+`Far(interface-name, object-with-methods)`
+- `interface-name` `{ String }`
+- `object-with-methods` ` { Object }` `[remotable={}]`
+-  Returns: A `Remotable` object.
+
+The `interface-name` parameter gives the `Remotable` an *interface name*, which only shows
+up in `console.log` when
```
warner, line 43: dangling sentence

maybe "only shows up when logged through `console.log()`"

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,75 @@
...
+
+The `object-with-methods` parameter includes a record with definitions of all the object's 
+property functions. See the example code below.
+
+Use the `Far()` function to mark an object as `Remotable`.  `Far()` also:
+- Runs `harden()` on the object.
+- Checks for the property name and value requirements above. If they are not met, it throws an error.
+- Records the object's interface name. 
+
+You should call `far()` on an object if it both:
```
warner, line 53: Capitalization inconsistency.. we need to use capital-F `Far()` everywhere.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,75 @@
...
+use [`E()`](./eventual-send.html#remote-object-communication-with-e). For example, getting
+an `Issuer`'s `brand` would look like `E(issuer).getBrand()`.
+
+Objects intended to be used in other vats are called `Remotables`. In particular, note that
+every object returned from a smart contract, such a `publicFacet` or 
+`creatorFacet`, must be `Remotable`. All objects used in your contract's external API must
+be `Remotable`
+
+### Rules for Creating Remotables
+- All property names must be strings.
```
warner, line 24: @erights should confirm (and I know we talked about this somewhere, maybe in chat?), but looking at the code, I think I was wrong when I made this claim.

Let's remove this "all property names must be strings" clause, and the subsidiary "property names must not be symbols" clause.

main/guides/js-programming/far.md
katelynsills: > I think the publicFacet and creatorFacet will generally not be Remotable, since they'll have all sorts of random data as properties.

We don't have any publicFacets or creatorFacets that aren't Remotable. By design, they must be Remotable, but I don't believe we enforce this anywhere at the Zoe level. 

> Would a Purse, Payment, or Offer be a better example of a Remotable? 

A purse is a good example of using a Remotable, but not of creating your own Remotable object, since that's done for you in ERTP. A payment isn't a good example of either using or creating, because you don't usually call methods on a payment. An offer proposal is a pass-by-copy record. The UserSeat that you get back when you make an offer could be a good example of Remotable usage, but not creation, since the user doesn't make the object.

erights:
See https://github.com/Agoric/agoric-sdk/pull/2909 for draft detailed semantics of `Passable`

main/guides/js-programming/far.md
tyg: I'm a little confused at this point. Since methods are called using payment as an argument, do payments need to be far()ed when used as arguments to a method on an Issuer object? And other than that, it seems that anything you get from an IssuerKit needs to be far()ed? But Amounts don't need to be far()ed, since they're now made locally via amountMath.make()? Also, if all the IssuerKit results will need far(), is there any reason not to just call far() on them internally just before returning them so the developer doesn't have to?

And is there any harm in far()ing an object that doesn't need to be? If so, is there any sort of error or warning that will catch doing so?

katelynsills: ERTP does all of the `Far` calls that are necessary, so there's no need to wrap a payment with `Far` - issuer.js already does that. 

> anything you get from an IssuerKit needs to be far()ed?

Nope, already handled for you.

> But Amounts don't need to be far()ed, since they're now made locally via amountMath.make()

Amounts are records with no functions as values so they are never wrapped with `Far`

> Also, if all the IssuerKit results will need far(), is there any reason not to just call far() on them internally just before returning them so the developer doesn't have to?

IssuerKit results don't need `Far` - that's handled internally.

> And is there any harm in far()ing an object that doesn't need to be? If so, is there any sort of error or warning that will catch doing so?

There's no harm in wrapping an object with `Far` even if it never leaves the vat. There's no error or warning, as there's nothing wrong with this. Wrapping an object that is actually a record and doesn't have function values will error, I believe. I can check this.

erights: If an object, `foo`, should never be exposed to other vats, there is a benefit to not declaring it `Far`, i.e., not making it a Remotable. If `foo` is not marked but is accidentally exposed, that will cause an error, suppressing the vulnerability that might have followed from such accidental exposure.

We don't have this benefit until `Far` or `Remotable` are mandatory, which we're close to. Certainly, the documentation should assume that it is mandatory.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+as if they are local.
+
+To call a method on an object from another vat or machine, you must 
+use [`E()`](./eventual-send.html#remote-object-communication-with-e). For example, getting
+an `Issuer`'s `brand` would look like `E(issuer).getBrand()`.
+
+Objects intended to be used in other vats are called `Remotables`. In particular, note that
+every object returned from a smart contract, such a `publicFacet` or 
+`creatorFacet`, must be `Remotable`. All objects used in your contract's external API must
+be `Remotable`
```
erights, line 21:
```suggestion
Objects intended to be used in other vats are called *remotables*. Remote messages sent to remotables must only contain *passable* arguments and return *passable* results. In particular, note that
every object returned from a smart contract, such a `publicFacet` or 
`creatorFacet`, must be passable. All objects used in your contract's external API must
be passable.

There are essentially three kinds of passable:
   * Remotables, objects with methods that can be called remotely using `E()`.
   * Pass-by-copy data such as numbers or hardened records
   * Promises for passables.
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`. You do 
+not need to do anything to make them remotable; i.e. you do not need to call `Far()` on an ERTP object.
+**tyg todo: Are UserSeat and ZCFSeat both also automatically created as `Remotable`?**
```
erights, line 25: The code that creates the object is the one who needs to call `Far` on it when they create it. "You" do not need to use `Far` on purses because you are not writing the code that creates these objects.

Yes, `UserSeat` and `ZCFSeat` are remotables.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Rules for Creating Remotables
+- All property values must be functions. 
+  - They cannot be accessors.
+  - Note: If you wish to send data, send a pass-by-copy record instead, or add a function that returns a pass-by-copy record.
```
erights, line 30: Any form of pass-by-copy data will do here, including pass-by-copy records, pass-by-copy arrays, strings, numbers, etc. This is implied by the general rule, which should be stated explicitly, that only passables may be passed as arguments or returns as results from remote messages. Again, this includes other remotables, pass-by-copy data, and promises for passables.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+- You must wrap the object with `Far()`.
+
+### Using `Remotables`
```
erights, line 33:
```suggestion
### Using Remotables
```

Please use code font for stuff that appears in code, such as `Far` when referring to the function, or `Remotable` when referring to the type. Here, we're just using the name of the concept to talk about the objects in that category.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+- Call a `Remotable`'s method by first wrapping the `Remotable` object with `E`, such as `E(issuer).getBrand();`
+- Handle the resulting promise. Calling `E()` always results in a `Promise`.
+
+## Using `Far()`
+
+`Far(interface-name, object-with-methods)`
+- `interface-name` `{ String }`
+- `object-with-methods` ` { Object }` `[remotable={}]`
```
erights, line 41: Are you trying to use TypeScript syntax even when it isn't clear what it means? I would recommend against that. What convention have you generally been using to document APIs?

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+-  Returns: A `Remotable` object.
+
+The `interface-name` parameter gives the `Remotable` an *interface name*, which only shows
+up when logged through `console.log`. 
+
+The `object-with-methods` parameter includes a record with definitions of all the object's 
+property functions. See the example code below.
+
+`Far()` function marks an object as `Remotable`.  `Far()` also:
+- Runs `harden()` on the object.
```
erights, line 51:
```suggestion
- Hardens the object.

Both the `harden` function and the `Far` function harden the object. Only hardened objects are passable.
```
 Notice where I do and do not use code font in saying that. The user shouldn't care whether the `Far` function calls `harden`. They care that the `Far` function hardens the object, i.e., that the resulting object is hardened.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+**tyg todo: Are UserSeat and ZCFSeat both also automatically created as `Remotable`?**
+
+### Rules for Creating Remotables
+- All property values must be functions. 
+  - They cannot be accessors.
+  - Note: If you wish to send data, send a pass-by-copy record instead, or add a function that returns a pass-by-copy record.
+- You must wrap the object with `Far()`.
+
+### Using `Remotables`
+- Call a `Remotable`'s method by first wrapping the `Remotable` object with `E`, such as `E(issuer).getBrand();`
```
erights, line 34:
```suggestion
- Call a remotable's method by first wrapping the remotable object with `E`, such as `E(issuer).getBrand();`
```

Only use code font for code. Here we're using the term "remotable" to talk about these objects. When such new terms are introduced and defined, they should be italics and not code font.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+- Handle the resulting promise. Calling `E()` always results in a `Promise`.
+
+## Using `Far()`
+
+`Far(interface-name, object-with-methods)`
+- `interface-name` `{ String }`
+- `object-with-methods` ` { Object }` `[remotable={}]`
+-  Returns: A `Remotable` object.
+
+The `interface-name` parameter gives the `Remotable` an *interface name*, which only shows
```
erights, line 44:
```suggestion
The `interface-name` parameter gives the `Remotable` an *interface name* for debugging purposes, which only shows
```

The actual parameter name cannot be `interface-name` because that is not a valid JavaScript variable name. Either use correct variable names and keep code font, or remove the code font.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+property functions. See the example code below.
+
+`Far()` function marks an object as `Remotable`.  `Far()` also:
+- Runs `harden()` on the object.
+- Checks for the property name and value requirements above. If they are not met, it throws an error.
+- Records the object's interface name. 
+
+You should call `Far()` on an object if it both:
+- Will be sent out of its native vat.
+  - If it might ever appear as the `foo` in [`E(foo)`](./eventual-send.md),
```
erights, line 57:
```suggestion
  - If it might ever appear as the `foo` in [`E(foo)`.method(args)](./eventual-send.md),  
```

We should never show `E(foo)` by itself as an expression. It should always be shown as the base of a method call.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+You should call `Far()` on an object if it both:
+- Will be sent out of its native vat.
+  - If it might ever appear as the `foo` in [`E(foo)`](./eventual-send.md),  
+    you should run `Far()` on it after creating it.
+- Has methods called on it, as opposed to just effectively storing data.
+
+`Far()` automatically [hardens](./ses/ses-guide.md#harden) its object argument. 
+If you make a `Remotable` object with `Far()`, you don't need to also call `harden()` 
+on it. Since `Far()` is not used on objects used to send data, you must still use
+`harden()` on them.
```
erights, line 64: This is all oddly put. Better to start by explaining that only passables may be passed as arguments or returned as results, and passable must be hardened. If the passable is a remotable, it must be hardened with `Far`. Otherwise it must be hardened using `harden`.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+- `object-with-methods` ` { Object }` `[remotable={}]`
+-  Returns: A `Remotable` object.
+
+The `interface-name` parameter gives the `Remotable` an *interface name*, which only shows
+up when logged through `console.log`. 
+
+The `object-with-methods` parameter includes a record with definitions of all the object's 
+property functions. See the example code below.
+
+`Far()` function marks an object as `Remotable`.  `Far()` also:
```
erights, line 50:
```suggestion
`Far()` function marks an object as remotable.  `Far()` also:
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+`Far()` automatically [hardens](./ses/ses-guide.md#harden) its object argument. 
+If you make a `Remotable` object with `Far()`, you don't need to also call `harden()` 
+on it. Since `Far()` is not used on objects used to send data, you must still use
+`harden()` on them.
+
+There's no harm in using `Far()` on an object, even if it never leaves its vat. An error
+is thrown if you call `Far()` on a record, instead of an object, which doesn't have function
+values. However, if object `foo` should never be exposed to other vats, you should make it
+a point **not** to use `Far()` on it. If `foo` is not marked as `Remotable` but is accidentally
+exposed, an error is thrown. This prevents any vulnerability from such accidental exposure.
```
erights, line 70: Good.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,85 @@
...
+
+```js
+import { Far } from '@agoric/marshal';
+let counter;
+const countRemotable = Far('counter', {
+  increment() { counter++; },
+  decrement() { counter--; },
+  read() { return counter; },
+});
```
erights, line 79: Better to put the example much earlier. Use `read` as an example of a method intended to be called remotely that returns pass-by-copy data. Specifically, a number.

erights (changes requested):

main/guides/js-programming/far.md
tyg: Pretty much this one per being told it was standard for JS method documentation. A line with the parameter name followed by its type in { }s

main/guides/js-programming/far.md
tyg: I want to keep source code associated with the reference section on how to use Far(); if someone's coming here just to check/refresh their memory on how to Far(), they're probably not going to look at/read the rest of the page.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+
+This is possible because Agoric encapsulates inter-machine and inter-vat communication. At
+the smart contract or dapp level, objects from other vats or machines can almost be treated 
+as if they are local.
+
+To call a method on an object from another vat or machine, you must 
+use [`E()`](./eventual-send.html#remote-object-communication-with-e). For example, getting
+an `Issuer`'s `brand` would look like `E(issuer).getBrand()`.
+
+Objects intended to be used in other vats are called *remotables*. Remote messages sent to
```
erights, line 18:
```suggestion
Objects intended to be used from other vats are called *remotables*. Remote messages sent to
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+
+In particular, note that every object returned from a smart contract, such a `publicFacet` or 
+`creatorFacet`, must be passable. All objects used in your contract's external API must
+be passable.
+
+### Rules for creating remotables
+- All property values must be functions. 
+  - They cannot be accessors.
+  - Note: If you wish to send data, send pass-by-copy data such as pass-by-copy records, 
+    pass-by-copy arrays, strings, numbers, etc. or some other passable.
```
erights, line 35: I think this bullet is more confusing than clarifying. The point it tries to make is already better made above. Please delete it.
```suggestion
```

main/guides/js-programming/far.md
erights: I don't get this. But consistency with what we're doing elsewhere is the overriding concern. No change suggested.

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+- `object-with-methods` ` { Object }` `[remotable={}]`
+-  Returns: A `Remotable` object.
+
+The `farName` parameter gives the `Remotable` an *interface name* for debugging purposes, which only shows
+up when logged through `console.log`. 
+
+The `object-with-methods` parameter includes a record with definitions of all the object's 
+property functions. See the example code below.
+
+`Far()` function marks an object as remotable.  `Far()` also:
```
erights, line 58:
```suggestion
The `Far()` function marks an object as remotable.  `Far()` also:
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Using `Far()`
+
+`Far(farName, object-with-methods)`
+- `farName` `{ String }`
+- `object-with-methods` ` { Object }` `[remotable={}]`
+-  Returns: A `Remotable` object.
+
+The `farName` parameter gives the `Remotable` an *interface name* for debugging purposes, which only shows
+up when logged through `console.log`.
```
erights, line 53:
```suggestion
up when logged through the `console`, for example with `console.log`. 
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+`Far(farName, object-with-methods)`
+- `farName` `{ String }`
+- `object-with-methods` ` { Object }` `[remotable={}]`
+-  Returns: A `Remotable` object.
+
+The `farName` parameter gives the `Remotable` an *interface name* for debugging purposes, which only shows
+up when logged through `console.log`. 
+
+The `object-with-methods` parameter includes a record with definitions of all the object's 
+property functions. See the example code below.
```
erights, line 56:
```suggestion
The `object-with-methods` parameter should be an object whose properties are the functions serving as the object's methods. See the example code below.
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+  If they are not met, it throws an error.
+- Records the object's interface name. 
+
+You should call `Far()` on an object if it both:
+- Will be sent out of its native vat.
+  - If it might ever appear as the `foo` in [`E(foo).method(args)`](./eventual-send.md),  
+    you should run `Far()` on it after creating it.
+- Has methods called on it, as opposed to just effectively storing data.
+
+As defined, only passables can be either passed as arguments or returned as results,
```
erights, line 72:
```suggestion
Only passables can be passed as arguments or returned as results,
```

main/guides/js-programming/far.md
```diff
@@ -0,0 +1,95 @@
...
+and they must be hardened. If the passable is a remotable, it must be hardened with `Far()`.
+Otherwise, it must be hardened with [`harden()`](./ses/ses-guide.md#harden).
+
+There's no harm in using `Far()` on an object, even if it never leaves its vat. An error
```
erights, line 76:
```suggestion
An error
```
Below you correctly explain a harm so we shouldn't start out saying there is none, and then contradict ourselves a moment later.

erights (changes requested):
We are converging!

main/guides/js-programming/far.md
```diff
@@ -69,11 +67,11 @@ You should call `Far()` on an object if it both:
...
-As defined, only passables can be either passed as arguments or returned as results,
+Only passables can be passed as arguments or returned as results,
 and they must be hardened. If the passable is a remotable, it must be hardened with `Far()`.
 Otherwise, it must be hardened with [`harden()`](./ses/ses-guide.md#harden).
 
-There's no harm in using `Far()` on an object, even if it never leaves its vat. An error
+An error
 is thrown if you call `Far()` on a record, instead of an object, which doesn't have function
 values. However, if object `foo` should never be exposed to other vats, you should make it
 a point **not** to use `Far()` on it. If `foo` is not marked as `Remotable` but is accidentally
```
erights, line 77: The term "remotable" here is a category for talking about objects. It is not itself an identifier that would appear in code.

```suggestion
a point **not** to use `Far()` on it. If `foo` is not marked as a remotable but is accidentally
```

erights (approved):
Only one correction left to do.

LGTM

PR #484, opened 2021-04-20
Automatic deployment to web site

michaelfig:
This CI task deploys the built site from `main` via an SSH Github action to our web site.

Tested with this PR's branch, verified, and then removed the current PR's branch from CI.


katelynsills (approved):
LGTM! Should we remove the "Deploy to gh-pages" step?

michaelfig:
> LGTM! Should we remove the "Deploy to gh-pages" step?

Only after our migration to the new site is solid (we're sure we don't want to flip back).

PR #486, opened 2021-04-21
Fix broken link in Zoe Contracts

barbaraliau:
Link was broken because there was an extra space. Removed space.


katelynsills (approved):
Awesome, thanks for finding and fixing this!

PR #492, opened 2021-05-03
Cleaned up method headers so all formatted in code font and put Seat and ZCFMint methods at third header level.

tyg:


katelynsills (approved):
Looks good to me, thanks!

Issue #494, opened 2021-05-04
zoe docs should recommend  throw seat.fail(Error)


Chris-Hibbert:
See https://github.com/Agoric/agoric-sdk/issues/3026

The docs for `zcfSeat.fail` should recommend 
```
   throw seat.fail(Error('you did it wrong');
```
when exiting a seat with an error condition. It might even be helpful to cross-direct there from `seat.exit()` to say that error conditions shouldn't use `exit()`.

tyg:
Fixed in [PR 495](Fix for issue 494; elaborate on how ZCF.fail is used #495)

PR #495, opened 2021-05-04
Fix for issue 494; elaborate on how ZCF.fail is used

tyg:
[Issue 494](https://github.com/Agoric/documentation/issues/494) described how to elaborate on how to correctly use ZCFSeat.exit() and .fail().


katelynsills (approved):
Looks good, thanks!

Chris-Hibbert (approved):

PR #497, opened 2021-05-07
chore: make some more changes to AssetKind branch

katelynsills:
@tyg, I finally figured out what I want to call assertUsesNatMath :)


tyg (approved):
The text changes look good, but it looks like a file may have been changed in CosmicSwingset causing it not to pass 'yarn lint'. When that's fixed, good to go.

katelynsills:
> but it looks like a file may have been changed in CosmicSwingset causing it not to pass 'yarn lint'. When that's fixed, good to go.

I think that's unrelated to this PR. I'd like to merge this into AssetKind, rebase that onto main, and get that reviewed. Separately, I'll look into the cosmic-swingset change and make a PR for that off main if needed.

PR #500, opened 2021-05-07
Asset kind

tyg:
#agoric-sdk-branch: zoe-beta-1


katelynsills (changes requested):
Let's change some of the language in amount-math.md but after that, I think we're ready to go!

katelynsills (approved):
The changes look good to me!

PR #505, opened 2021-05-21
Updates to JS programming overview page

mhofman:
- Dynamic import is now stage 4
- Make clear that dynamic `import` would bypass compartment shim because of its syntax status
- Microtask and task queuing nits
- Add Set and WeakSet as allowed globals for completeness
- Clarify why incomplete eval regex is safe

### DIFF
--- a/main/guides/js-programming/agoric-js-overview.md
+++ b/main/guides/js-programming/agoric-js-overview.md
@@ -196,7 +196,7 @@ sneaky = import
 // tricky comment to obscure function invocation
 (modulename);
 ```
-There are also problems with “import” being near a parenthesis inside a comment. 
+There are also problems when “import” is near a parenthesis inside a comment. 
 
 ### Direct vs. indirect eval expressions
 
--- a/main/guides/js-programming/agoric-js-overview.md
+++ b/main/guides/js-programming/agoric-js-overview.md
@@ -33,8 +33,8 @@ defines a subset of the globals defined by the baseline JavaScript language spec
 - `Object`
 - `Array`
 - `Number`
-- `Map`
-- `WeakMap`
+- `Map` / `Set`
+- `WeakMap` / `WeakSet`
 - `Number`
 - `BigInt`
 - `Intl`
--- a/main/guides/js-programming/agoric-js-overview.md
+++ b/main/guides/js-programming/agoric-js-overview.md
@@ -176,11 +176,11 @@ containing either of the strings `<!--` or `-->`, even if neither marks a commen
 
 ### Dynamic import expressions
 
-The "dynamic import" feature (`import('path')`) enables code to load dependencies at runtime.
-While it takes the form of an expression returning a Promise, it's actually not a normal
-function call, but is instead JavaScript syntax. As such it would let vat code bypass the
-shim's `Compartment`'s module map. For safety, the SES shim rejects code that looks like it
-uses the dynamic import syntax.
+The "dynamic import expression" (`import('path')`) enables code to load dependencies at
+runtime. It returns a promise resolving to the module namespace object. While it takes
+the form of a function call, it's actually not a function call, but is instead JavaScript
+syntax. As such it would let vat code bypass the shim's `Compartment`'s module map.
+For safety, the SES shim rejects code that looks like it uses a dynamic import expression.
 
 The regular expression for this pattern is safe and should never allow any use of
 dynamic import, however obfuscated the usage is. Because of this, it may be confused
--- a/main/guides/js-programming/agoric-js-overview.md
+++ b/main/guides/js-programming/agoric-js-overview.md
@@ -107,13 +107,10 @@ additions and not part of JavaScript.
 Most Node.js-specific [global objects](https://nodejs.org/dist/latest-v14.x/docs/api/globals.html) 
 are unavailable within a vat including:
 
-* `queueMicrotask`
+* `queueMicrotask`: You can 
+  replace `queueMicrotask(fn)` with `Promise.resolve().then(_ => fn())`.
 * `Buffer` (consider using `TypedArray` instead, but see below)
-* `setImmediate`/`clearImmediate`: Not available, but you can generally 
-  replace `setImmediate(fn)` with `Promise.resolve().then(_ => fn())` to 
-  defer execution of `fn` until after the current event/callback finishes 
-  processing. But be aware it won't run until after all *other* ready 
-  Promise callbacks execute. 
+* `setImmediate`/`clearImmediate`: Not available
 
 There are two queues: the *IO queue* (accessed by `setImmediate`), and 
 the *Promise queue* (accessed by Promise resolution). SES code can only 
--- a/main/guides/js-programming/agoric-js-overview.md
+++ b/main/guides/js-programming/agoric-js-overview.md
@@ -219,16 +219,18 @@ authorities as you like.
 
 The most common way to invoke an indirect eval is `(1,eval)(code)`.
 
-The SES shim cannot correctly emulate a direct eval. If it tried, it would perform 
-an indirect eval. This could be pretty confusing, because the code might not actually 
-use objects from the local scope. You might not notice the problem until some later 
-change altered the behavior.
+The SES proposal does not change how direct and indirect eval work. However, the SES shim
+cannot correctly emulate a direct eval. If it tried, it would perform an indirect eval.
+This could be pretty confusing, because the evaluated code would not use objects from
+the local scope as expected. Furthermore, in the future when SES is natively implemented
+by JavaScript engines, the behavior would revert to direct eval, allowing access to
+anything in scope.
 
-To avoid this confusion, the shim uses a regular expression to reject code that 
-looks like it is performing a direct eval. This regexp is not complete (you can 
-trick it into allowing a direct eval), but that’s safe because it really performs
-an indirect eval. Our goal is just to guide people away from confusing behaviors
-early in their development process.
+To avoid this confusion and compatibility risk, the shim uses a regular expression to
+reject code that looks like it is performing a direct eval. This regexp is not complete
+(you can trick it into allowing a direct eval), but that’s safe because it really performs
+an indirect eval. Our goal is just to guide people away from confusing and non-compliant
+behaviors early in their development process.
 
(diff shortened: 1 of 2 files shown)


main/guides/js-programming/agoric-js-overview.md
```diff
@@ -226,8 +226,9 @@ change altered the behavior.
 
 To avoid this confusion, the shim uses a regular expression to reject code that 
 looks like it is performing a direct eval. This regexp is not complete (you can 
-trick it into performing a direct eval anyway), but that’s safe. Our goal is 
-just to guide people away from confusing behaviors early in their development process.
+trick it into allowing a direct eval), but that’s safe because it really performs
+an indirect eval. Our goal is just to guide people away from confusing behaviors
+early in their development process.
```
erights, line 231: What's here looks good. But please also mention that SES as natively implemented will support direct eval. Thus, this prohibition helps avoid code that will behave differently between the SES shim and SES as natively implemented.

erights (approved):
LGTM with the additional point made.

Thanks for fixing this!

main/guides/js-programming/agoric-js-overview.md
mhofman: Updated. PTAL @erights

main/guides/js-programming/agoric-js-overview.md
```diff
@@ -107,13 +107,10 @@ additions and not part of JavaScript.
 Most Node.js-specific [global objects](https://nodejs.org/dist/latest-v14.x/docs/api/globals.html) 
 are unavailable within a vat including:
 
-* `queueMicrotask`
+* `queueMicrotask`: You can 
+  replace `queueMicrotask(fn)` with `Promise.resolve().then(_ => fn())`.
```
erights, line 111: Are you sure that's true? Does `queueMicrotask` use the promise queue, or (as I'd expect) does it use another queue with a lower priority, like `setImmediate`, etc?

main/guides/js-programming/agoric-js-overview.md
```diff
@@ -179,10 +176,7 @@ containing either of the strings `<!--` or `-->`, even if neither marks a commen
 
 ### Dynamic import expressions
 
-One active JS feature proposal would add a "dynamic import" expression: `await import('path')`. 
-If implemented (or if someone decides to be an early adopter and adds it to an engine), 
-and your JS engine has this, vat code might be able to bypass the `Compartment`'s module 
-map. For safety, the SES shim already rejects code that looks like it uses this feature. 
+The "dynamic import" feature (`await import('path')`) is not a function call but JavaScript syntax. As such it would allow vat code to bypass the shim's `Compartment`'s module map. For safety, the SES shim rejects code that looks like it uses this feature.
```
erights, line 179: I see that both the old and new says `await import('path')` rather than just `import('path')`. This is wrong. Instead just say that the dynamic import expression returns a promise. If you think it would be helpful, show the longer expression below as a suggested usage pattern.

I would say "dynamic import expression" rather than "feature". The fact that it is an expression is important, to more clearly distinguish it from import declarations.

main/guides/js-programming/agoric-js-overview.md
```diff
@@ -225,15 +219,18 @@ authorities as you like.
...
+This could be pretty confusing, because the evaluated code would not use objects from
+the local scope as expected. Furthermore, in the future when SES is natively implemented
+by JavaScript engines, the behavior would revert to direct eval, allowing access to
+anything in scope.
+
+To avoid this confusion and compatibility risk, the shim uses a regular expression to
+reject code that looks like it is performing a direct eval. This regexp is not complete
+(you can trick it into allowing a direct eval), but that’s safe because it really performs
+an indirect eval. Our goal is just to guide people away from confusing and non-compliant
+behaviors early in their development process.
```
erights, line 233: Excellent

erights (approved):
LGTM with some corrections. If you'd like another round of review after making those corrections, please let me know.

main/guides/js-programming/agoric-js-overview.md
mhofman: 100% sure: https://twitter.com/mhofman/status/1327729938640773120

main/guides/js-programming/agoric-js-overview.md
mhofman: I wanted to highlight that the bypass risk is due to the syntax nature of dynamic import. The fact that it can be used as an expression matters in that it makes it different in usage from static import declarations. If dynamic `import` was not a syntax but a function call, we would be able to trap them without compartments, and they wouldn't be a hazard.

I'll try to rewrite to make both dynamic expression and syntax more clear.

tyg: Please change "would allow vat code to bypass" to "would let vat code bypass".

tyg (commented):
Other than the one minor suggestion for tightening a phrase, looks good. Assume I've approved it on my end once you finish addressing the SMEs' comments.

mhofman:
@tyg, I see some of the same content is pretty much word for word in the SES guide. Are there any plans to share snippets or should I just update both?

katelynsills:
I'll take myself off as a reviewer. Between @erights and @tyg this seems to be well covered. :)

main/guides/js-programming/agoric-js-overview.md
mhofman: @tyg done
@erights, I tried to address your concern while still highlighting the syntax aspect. PTAL

mhofman:
Copied the changes from JS Programming page onto SES Guide, except for one line which I copied in the other direction as it read more clearly IMO.

main/guides/js-programming/agoric-js-overview.md
```diff
@@ -198,7 +196,7 @@ sneaky = import
 // tricky comment to obscure function invocation
```
erights, line 196: This is an example where our `import` regexp successfully rejects code that should be rejected, but the text suggests the opposite. We need to make clear that for this prohibition (unlike direct eval) we *only* make mistakes of falsely rejecting code that was actually safe. We believe we never mistakenly accept code that we should have rejected. This example looks like it is an example of the second.

main/guides/js-programming/agoric-js-overview.md
```diff
@@ -198,7 +196,7 @@ sneaky = import
...
 (modulename);
 ```
-There are also problems with “import” being near a parenthesis inside a comment. 
+There are also problems when “import” is near a parenthesis inside a comment.
```
erights, line 199: Again, it is not clear what direction this problem is. It can only cause a false rejection.

main/guides/js-programming/ses/ses-guide.md
```diff
@@ -438,12 +438,13 @@ even if neither marks a comment.
...
+The "dynamic import" feature (`import('path')`) enables code to load dependencies at runtime.
+While it takes the form of an expression returning a Promise, it's actually not a normal
+function call, but is instead JavaScript syntax. As such it would let vat code bypass the
+shim's `Compartment`'s module map. For safety, the SES shim rejects code that looks like it
+uses the dynamic import syntax.
+The regular expression for this pattern can be confused into falsely rejecting legitimate 
+code. For example, the word “import” at the end of a line in a comment, such as:
 ```js
 //
 // This function calculates the import
```
erights, line 450: This example has the same problem as the one in agoric-js-overview.md

erights (changes requested):

mhofman:
PTAL @erights

erights (changes requested):
Somewhat clearer wording suggested. Also for where wording is duplicated of course.

main/guides/js-programming/agoric-js-overview.md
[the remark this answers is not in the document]
mhofman: I had just copy-pasted the code from another place, and technically there would be a difference if `fn` is declared as such:
```js
function fn(...args) {
  console.log(args.length);
}
```

That said, if you think it makes the example more clear, happy to change.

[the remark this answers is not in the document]
erights: Interesting example! Since the one argument would be a single `undefined` the shorter code is mostly-harmlessly different that the code it replaces. I think we should use the shorter code.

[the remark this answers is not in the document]
erights: Just saw your keybase chat note linking to https://jakearchibald.com/2021/function-callback-risks/ . Good point! Ok, the original is fine. No change suggested.

main/guides/js-programming/agoric-js-overview.md
```diff
@@ -176,11 +176,11 @@ containing either of the strings `<!--` or `-->`, even if neither marks a commen
 
 ### Dynamic import expressions
 
-The "dynamic import" feature (`import('path')`) enables code to load dependencies at runtime.
-While it takes the form of an expression returning a Promise, it's actually not a normal
-function call, but is instead JavaScript syntax. As such it would let vat code bypass the
-shim's `Compartment`'s module map. For safety, the SES shim rejects code that looks like it
-uses the dynamic import syntax.
+The "dynamic import expression" (`import('path')`) enables code to load dependencies at
+runtime. It returns a promise resolving in the module namespace object. While it takes
```
erights, line 180:
```suggestion
runtime. It returns a promise resolving to the module namespace object. While it takes
```

main/guides/js-programming/ses/ses-guide.md
```diff
@@ -438,11 +438,11 @@ even if neither marks a comment.
 
 ### Dynamic import expressions
 
-The "dynamic import" feature (`import('path')`) enables code to load dependencies at runtime.
-While it takes the form of an expression returning a Promise, it's actually not a normal
-function call, but is instead JavaScript syntax. As such it would let vat code bypass the
-shim's `Compartment`'s module map. For safety, the SES shim rejects code that looks like it
-uses the dynamic import syntax.
+The "dynamic import expression" (`import('path')`) enables code to load dependencies at
+runtime. It returns a promise resolving in the module namespace object. While it takes
```
erights, line 442:
```suggestion
runtime. It returns a promise resolving to the module namespace object. While it takes
```

erights (approved):
LGTM just one word remaining

PR #506, opened 2021-05-21
Document passable records

mhofman:
Clarify what are passable objects, add glossary entry for record and tuple.
Fix harden description to remove enumerable constraint.
Add a note on making remotable handles.

### DIFF
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -413,6 +413,10 @@ episode [The Gamesters of Triskelion](https://en.wikipedia.org/wiki/The_Gamester
 A transfer of [amounts](#amounts) between [seats](#seat) within Zoe; i.e. a change in their [allocations](#allocation). When a seat exits, it gets its
 current allocation as a [payout](#payout). 
 
+## Record and Tuple
+
+Records and tuples are immutable objects which can be passed by copy. They only contain values which can be passed by copy such as numbers, strings, or other records and tuples. 
+For more details, see [Passable objects documentation](/guides/js-programming/far.md#rules-for-creating-passable-objects).
 ## Seat
 
 Zoe uses a seat to represent an [offer](#offer) in progress, and has two seat [facets](#facet) representing
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -112,13 +112,9 @@ Destroy all digital assets in a payment. See [`issuer.burn(payment, optAmount)`]
 
 ## Comparable
 
-A *passable* is something that can be marshalled. A *comparable* is a
-passable whose leaves contain no promises. Two comparables can be
-synchronously compared for structural equivalence.
-
-A comparable is a JavaScript object containing no promises, and can
-thus be locally compared for equality with another object. If either object
-contains Promises, equality is indeterminable. If both are fulfilled down
+A *comparable* is a [passable](#passable) whose leaves contain no promises. Two 
+comparables can be synchronously compared for structural equivalence. If either passable
+object contains Promises, equality is indeterminable. If both are fulfilled down
 to Presences and local state, then either they're the same all the way
 down, or they represent different objects.
 
@@ -354,6 +350,18 @@ the user either gets what they said they wanted, or gets back (gets a refund) wh
 escrowed. One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
 can immediately cause the [seat](#seat) to exit, getting back the amount it offered.
 
+
+## Passable
+
+A *passable* is something that can be marshalled.
+
+There are three kinds of passables:
+   * Remotables, objects with methods that can be called remotely using `E()`, and their remote Presence.
+   * Pass-by-copy data, such as numbers or hardened records.
+   * Promises for passables.
+
+For more information, see the [Remotable and passable objects documentation](/guides/js-programming/far.md#remotable-and-passable-objects).
+
 ## Payment
 
 Payments hold assets created by [Mints](#mint). Specifically assets intended for transfer 
@@ -415,7 +423,7 @@ current allocation as a [payout](#payout).
 
 ## Record and Tuple
 
-Records and tuples are immutable objects which can be passed by copy. They only contain values which can be passed by copy such as numbers, strings, or other records and tuples. 
+Records and tuples are immutable objects which can be passed by copy. They only contain values which are [passables](#passable). 
 For more details, see [Passable objects documentation](/guides/js-programming/far.md#rules-for-creating-passable-objects).
 ## Seat
 
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -425,6 +425,22 @@ current allocation as a [payout](#payout).
 
 Records and tuples are immutable objects which can be passed by copy. They only contain values which are [passables](#passable). 
 For more details, see [Passable objects documentation](/guides/js-programming/far.md#rules-for-creating-passable-objects).
+
+## Remotable
+
+Remotables are objects intended to be used from other vats. Remote messages sent to remotables must only contain [passable](#passable)
+arguments and return passable results. 
+
+All of a remotable's property values must be functions, and cannot be accessors. Its methods are called by wrapping 
+the remotable object with [`E`](#e), such as `E(issuer).getBrand()`. As with all calls to `E()`, it results in a Promise.
+
+Every object returned from a smart contract, such a publicFacet or creatorFacet, must be passable. All objects used in 
+your contract's external API must be passable.
+
+ERTP objects, such as `Purses`, are automatically created as `Remotable`, as are `UserSeats` and `ZCFSeats`.
+
+For more information, see the [Remotable and passable objects documentation](/guides/js-programming/far.md#remotable-and-passable-objects).
+
 ## Seat
 
 Zoe uses a seat to represent an [offer](#offer) in progress, and has two seat [facets](#facet) representing
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
 `creatorFacet`, must be passable. All objects used in your contract's external API must
 be passable.
 
+### Rules for creating passable objects
+Record and tuples are hardened acyclical objects which can be passed-by-copy if they satisfy some constraints.
+
+#### Record
+A plain object where:
+- All properties must have pass-by-copy values, such as numbers, strings, or other records and tuples.
+  - Properties cannot be accessors.
+- All properties must be own and enumerable, with string names.
+  - Symbols are not allowed.
+  - The prototype must be `Object.prototype` or `null`.
+- Nested values cannot form a cycle.
+- It must be hardened with `harden()`.
+
+#### Record
+A plain array where:
+- All properties must have pass-by-copy values, such as numbers, strings, or other records and tuples.
+  - Properties cannot be accessors.
+- All properties must be own and enumerable, where names are the integers indexes or `length`.
+  - The array cannot have holes.
+  - The prototype must be `Array.prototype`.
+- Nested values cannot form a cycle.
+- It must be hardened with `harden()`.
+
 ### Rules for creating remotables
 - All property values must be functions. 
   - They cannot be accessors.
 - You must wrap the object with `Far()`.
 
 **Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`, as are
-`UserSeats` and `ZCFSeats`. 
+`UserSeats` and `ZCFSeats`.
+
+**Note**: A hardened empty object can be made remotable instead of passable by copy when
+wrapped with `Far()`. This allows making an unforgeable [Handle](/glossary/#handle).
 
 ### Using remotables
 - Call a remotable's method by first wrapping the remotable object with `E`, such as `E(issuer).getBrand();`
--- a/main/guides/js-programming/far.md
+++ b/main/guides/js-programming/far.md
@@ -33,22 +33,22 @@ Record and tuples are hardened acyclical objects which can be passed-by-copy if 
 
 #### Record
 A plain object where:
-- All properties must have pass-by-copy values, such as numbers, strings, or other records and tuples.
+- All property values must be passables.
   - Properties cannot be accessors.
 - All properties must be own and enumerable, with string names.
   - Symbols are not allowed.
   - The prototype must be `Object.prototype` or `null`.
-- Nested values cannot form a cycle.
+- Nested records and tuples cannot form a cycle.
 - It must be hardened with `harden()`.
 
-#### Record
+#### Tuple
 A plain array where:
-- All properties must have pass-by-copy values, such as numbers, strings, or other records and tuples.
+- All property values must be passables.
   - Properties cannot be accessors.
-- All properties must be own and enumerable, where names are the integers indexes or `length`.
+- All properties must be own, and be all the integers indexes or `length`.
   - The array cannot have holes.
   - The prototype must be `Array.prototype`.
-- Nested values cannot form a cycle.
+- Nested records and tuples cannot form a cycle.
 - It must be hardened with `harden()`.
 
 ### Rules for creating remotables
--- a/main/guides/js-programming/agoric-js-overview.md
+++ b/main/guides/js-programming/agoric-js-overview.md
@@ -69,7 +69,7 @@ The vat environment has four significant objects not part of standard JavaScript
   for debug information only. The console is not obliged to write to the POSIX 
   standard output.
 
-- `harden` is a global that freezes an object’s API surface (enumerable data properties). 
+- `harden` is a global that freezes an object’s API surface. 
   A hardened object’s properties cannot be changed, so the only way to interact 
   with a hardened object is through its methods. `harden()` is similar to `Object.freeze()` 
   but more powerful. For more details, 


main/glossary/README.md
```diff
@@ -413,6 +413,10 @@ episode [The Gamesters of Triskelion](https://en.wikipedia.org/wiki/The_Gamester
 A transfer of [amounts](#amounts) between [seats](#seat) within Zoe; i.e. a change in their [allocations](#allocation). When a seat exits, it gets its
 current allocation as a [payout](#payout). 
 
+## Record and Tuple
+
+Records and tuples are immutable objects which can be passed by copy. They only contain values which can be passed by copy such as numbers, strings, or other records and tuples.
```
katelynsills, line 418: Hmm, they can contain presences/remotables (what's the right term again?) as values, so we should probably mention that. Otherwise I think this description is misleading.

katelynsills (changes requested):
I think by record, we mean [CopyRecord, as defined in marshal](https://github.com/Agoric/agoric-sdk/blob/61c97d0e9c576d916931a2cefa758e7f08987a0b/packages/marshal/src/passStyleOf.js#L147), so the definition written in this PR is too restrictive. Can we change it to a description of a Passable CopyRecord? 

Additional types here: https://github.com/Agoric/agoric-sdk/blob/61c97d0e9c576d916931a2cefa758e7f08987a0b/packages/marshal/src/types.js#L65

main/glossary/README.md
mhofman: Oh I didn't realize that remotables could be nested inside pass by copy structures! And it is definitely not obvious from the documentation.
https://agoric.com/documentation/guides/js-programming/far.html

mhofman:
> I think by record, we mean [CopyRecord, as defined in marshal](https://github.com/Agoric/agoric-sdk/blob/61c97d0e9c576d916931a2cefa758e7f08987a0b/packages/marshal/src/passStyleOf.js#L147), so the definition written in this PR is too restrictive.

I believe these are the restrictions I followed. which ones are extraneous?

My understanding is that `isPassByCopyRecord` is a preliminary check, and that during actual marshalling `serialize`/`encode` does other recursive checks.
https://github.com/Agoric/agoric-sdk/blob/61c97d0e9c576d916931a2cefa758e7f08987a0b/packages/marshal/src/marshal.js#L265

> Additional types here: https://github.com/Agoric/agoric-sdk/blob/61c97d0e9c576d916931a2cefa758e7f08987a0b/packages/marshal/src/types.js#L65

Um I don't see any passable by copy in that type. I thought `Promise`s were allowed as passable if they resolve to a passable (remotable or pass by copy object). In that sense they're a "third kind" of passable.

mhofman:
I've fixed the passable in records issue.

Are there any other inaccuracies @katelynsills ?

main/glossary/README.md
```diff
@@ -354,6 +350,18 @@ the user either gets what they said they wanted, or gets back (gets a refund) wh
...
+## Passable
+
+A *passable* is something that can be marshalled.
+
+There are three kinds of passables:
+   * Remotables, objects with methods that can be called remotely using `E()`, and their remote Presence.
+   * Pass-by-copy data, such as numbers or hardened records.
+   * Promises for passables.
+
+For more information, see the [Remotable and passable objects documentation](/guides/js-programming/far.md#remotable-and-passable-objects).
```
erights, line 363: Link seems broken, at least when I try to follow it in github preview. In any case, I see the intended target below. So if this is the correct way to link to it, no change suggested.

erights (approved):
LGTM! Thanks so much for fixing this. I especially like your changes to far.md

katelynsills:
> > I think by record, we mean [CopyRecord, as defined in marshal](https://github.com/Agoric/agoric-sdk/blob/61c97d0e9c576d916931a2cefa758e7f08987a0b/packages/marshal/src/passStyleOf.js#L147), so the definition written in this PR is too restrictive.
> 
> I believe these are the restrictions I followed. which ones are extraneous?

Ah, I was referring the text before your latest commit, the text that restricted a record to only having string, number, etc values. It looks correct now 👍 

Would it be possible to make the explanation more accessible? I think we have a very different audience than our internal documentation. Our internal documentation values conciseness and often references itself. Here, we should value putting things in terms that the users are familiar with. So for instance, instead of the record definition referencing Passables and requiring the user to look that up, we might provide some examples of Passables first such the user only has to learn more if they really want to know.

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
 `creatorFacet`, must be passable. All objects used in your contract's external API must
 be passable.
 
+### Rules for creating passable objects
+Record and tuples are hardened acyclical objects which can be passed-by-copy if they satisfy some constraints.
```
Chris-Hibbert, line 32:
```suggestion
Record and tuples are hardened acyclic objects which can be passed-by-copy if they satisfy some constraints.
```

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
...
+  - Symbols are not allowed.
+  - The prototype must be `Object.prototype` or `null`.
+- Nested records and tuples cannot form a cycle.
+- It must be hardened with `harden()`.
+
+#### Tuple
+A plain array where:
+- All property values must be passables.
+  - Properties cannot be accessors.
+- All properties must be own, and be all the integers indexes or `length`.
```
Chris-Hibbert, line 48:
```suggestion
- All properties must be own, and contain all the integers indexes up to `length`.
```

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
...
+- It must be hardened with `harden()`.
+
 ### Rules for creating remotables
 - All property values must be functions. 
   - They cannot be accessors.
 - You must wrap the object with `Far()`.
 
 **Note**: ERTP objects, such as `Purses`, are automatically created as `Remotable`, as are
-`UserSeats` and `ZCFSeats`. 
+`UserSeats` and `ZCFSeats`.
```
Chris-Hibbert, line 60: UserSeat should link to https://agoric.com/documentation/zoe/api/zoe.html#userseat-object
ZCFSeat should link to https://agoric.com/documentation/zoe/api/zoe-contract-facet.html#zcfseat-object
Purse to https://agoric.com/documentation/ertp/api/purse.html

main/guides/js-programming/far.md
mhofman: Technically `length` is also an own property of the array, and the array cannot have non indexes properties (besides `length`). I know my formulation is a bit clunky, but I believe it is more correct, as "contain" doesn't preclude extra props. I would love to find better wording, but I think we should strive to keep it correct.

mhofman:
> Would it be possible to make the explanation more accessible? I think we have a very different audience than our internal documentation. Our internal documentation values conciseness and often references itself. Here, we should value putting things in terms that the users are familiar with. So for instance, instead of the record definition referencing Passables and requiring the user to look that up, we might provide some examples of Passables first such the user only has to learn more if they really want to know.

@katelynsills, is this in regard to the glossary or the js-programming/far page? IMO it's ok for a glossary to have self references, but maybe I can add some inline examples. The Far programming page lists somewhat exhaustively what passables are 2 paragraphs above the record and tuple definitions.

main/guides/js-programming/far.md
Chris-Hibbert: I couldn't parse "and be all the integers indexes or `length`."   Is the requirement that all integers up to `length - 1` must be present,  or only that nothing else is allowed, or something else?

erights: I like @Chris-Hibbert 's suggestion. I'd add "No other properties are allowed" (unless that is already implied by the surrounding text).

katelynsills:
> @katelynsills, is this in regard to the glossary or the js-programming/far page? IMO it's ok for a glossary to have self references, but maybe I can add some inline examples. The Far programming page lists somewhat exhaustively what passables are 2 paragraphs above the record and tuple definitions.

A few tiny inline examples in the glossary sound great, but feel free to merge without. Good point about the self-references there.

katelynsills (commented):
LGTM!

mhofman:
@Chris-Hibbert @erights, I've actually been considering flipping all this around and combining the records and tuples definitions. @tyg, any copy feedback also greatly appreciated :)

How's this: 

> ### Rules for creating passable data
> Complex data can be passed by copy. Besides primitive data such as strings and numbers, record and tuples are hardened acyclic objects which are also passed-by-copy if they satisfy the following constraints:
> 
> - A record is a plain object, containing only enumerable properties with string names.
>   - Symbol names are not allowed.
>   - The prototype must be `Object.prototype` or `null`.
> - A tuple is a plain array with no holes.
>   - Integer index properties up to `length` must exist and be enumerable.
>   - No other properties besides array indexes and `length` are allowed.
>   - The prototype must be `Array.prototype`.
> - All property values must be passables, including remotables or promises for passables.
>   - Properties cannot be accessors.
> - Nested records and tuples cannot form a cycle.
> - Objects (and arrays) must be hardened with `harden()`.

tyg:
At dinner, but will take a look in a bit. Was waiting for the dust to
settle from the definition/phrasing battles before going at the language. :
-)

Tom

erights:
Text at https://github.com/Agoric/documentation/pull/506#issuecomment-847459169 LGTM. Thanks!

main/glossary/README.md
tyg: Links get changed when Vuepress processes the Markdown into HTML for display, at the very least from .md to .html.

main/glossary/README.md
```diff
@@ -354,6 +350,18 @@ the user either gets what they said they wanted, or gets back (gets a refund) wh
 escrowed. One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
 can immediately cause the [seat](#seat) to exit, getting back the amount it offered.
 
+
+## Passable
+
+A *passable* is something that can be marshalled.
+
+There are three kinds of passables:
+   * Remotables, objects with methods that can be called remotely using `E()`, and their remote Presence.
```
tyg, line 359: Link E() to its glossary entry #e.

main/glossary/README.md
```diff
@@ -354,6 +350,18 @@ the user either gets what they said they wanted, or gets back (gets a refund) wh
...
+   * Pass-by-copy data, such as numbers or hardened records.
+   * Promises for passables.
```
tyg, line 361: This is a bit circular. How about "Promises for either remotables or pass-by-copy data." to be clearer?

main/glossary/README.md
```diff
@@ -112,13 +112,9 @@ Destroy all digital assets in a payment. See [`issuer.burn(payment, optAmount)`]
...
-A *passable* is something that can be marshalled. A *comparable* is a
-passable whose leaves contain no promises. Two comparables can be
-synchronously compared for structural equivalence.
-
-A comparable is a JavaScript object containing no promises, and can
-thus be locally compared for equality with another object. If either object
-contains Promises, equality is indeterminable. If both are fulfilled down
+A *comparable* is a [passable](#passable) whose leaves contain no promises. Two 
+comparables can be synchronously compared for structural equivalence. If either passable
+object contains Promises, equality is indeterminable. If both are fulfilled down
```
tyg, line 117: I think this can be made a bit clearer. How about 
"A *comparable* is a [passable](#passable) which does not contain any Promises, It can be locally and synchronously compared to another object for equality, including structural equivalence. But if either passable contains Promises, equality cannot be determined. If both objects are fulfilled down to Presences and local state, then either they're the same all the way down, or they represent different objects."

main/glossary/README.md
tyg: Added a "Remotable" entry.

main/glossary/README.md
erights: Saying "pass-by-copy data" would exclude remotables. But promises can be fulfilled to remotables, so that would not be correct. 

The truth is exactly as circular as it sounds. A promise can be fulfilled to any passable. To be fully accurate, a promise can be fulfilled to any non-promise passable.

main/glossary/README.md
```diff
@@ -413,6 +421,10 @@ episode [The Gamesters of Triskelion](https://en.wikipedia.org/wiki/The_Gamester
 A transfer of [amounts](#amounts) between [seats](#seat) within Zoe; i.e. a change in their [allocations](#allocation). When a seat exits, it gets its
 current allocation as a [payout](#payout). 
 
+## Record and Tuple
+
+Records and tuples are immutable objects which can be passed by copy. They only contain values which are [passables](#passable). 
+For more details, see [Passable objects documentation](/guides/js-programming/far.md#rules-for-creating-passable-objects).
```
tyg, line 427: Suggest adding "see the full..."

main/guides/js-programming/far.md
tyg: Suggest changing "if: to "when".

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
 `creatorFacet`, must be passable. All objects used in your contract's external API must
 be passable.
 
+### Rules for creating passable objects
+Record and tuples are hardened acyclical objects which can be passed-by-copy if they satisfy some constraints.
+
+#### Record
+A plain object where:
```
tyg, line 35: Change "where" to "for which"

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
...
+- All property values must be passables.
+  - Properties cannot be accessors.
+- All properties must be own and enumerable, with string names.
+  - Symbols are not allowed.
+  - The prototype must be `Object.prototype` or `null`.
+- Nested records and tuples cannot form a cycle.
+- It must be hardened with `harden()`.
+
+#### Tuple
+A plain array where:
```
tyg, line 45: Change "where" to "for which"

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
...
 be passable.
 
+### Rules for creating passable objects
+Record and tuples are hardened acyclical objects which can be passed-by-copy if they satisfy some constraints.
+
+#### Record
+A plain object where:
+- All property values must be passables.
+  - Properties cannot be accessors.
+- All properties must be own and enumerable, with string names.
```
tyg, line 38: Would change "own" to "own (defined directly on the object)"

main/guides/js-programming/far.md
tyg, line 48: Would change "own" to "own (defined directly on the object)"

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
...
+- Nested records and tuples cannot form a cycle.
+- It must be hardened with `harden()`.
+
+#### Tuple
+A plain array where:
+- All property values must be passables.
+  - Properties cannot be accessors.
+- All properties must be own, and be all the integers indexes or `length`.
+  - The array cannot have holes.
+  - The prototype must be `Array.prototype`.
```
tyg, line 50: Change "The prototype" to "Its prototype".

main/guides/js-programming/far.md
```diff
@@ -28,13 +28,39 @@ In particular, note that every object returned from a smart contract, such a `pu
...
+### Rules for creating passable objects
+Record and tuples are hardened acyclical objects which can be passed-by-copy if they satisfy some constraints.
+
+#### Record
+A plain object where:
+- All property values must be passables.
+  - Properties cannot be accessors.
+- All properties must be own and enumerable, with string names.
+  - Symbols are not allowed.
+  - The prototype must be `Object.prototype` or `null`.
```
tyg, line 40: Change "The prototype" to "Its prototype".

main/guides/js-programming/far.md
tyg: "present" how? As indices? If so, specify that.

tyg:
Re: Mathieu's proposed text. Good but still suggest minor tweaks I put in
comments;
change if they satisfy to "when they satisfy". "indexes" should be
"indices". "The prototype" -> "Its prototype".

Tom

main/glossary/README.md
```diff
@@ -425,6 +425,22 @@ current allocation as a [payout](#payout).
...
+## Remotable
+
+Remotables are objects intended to be used from other vats. Remote messages sent to remotables must only contain [passable](#passable)
+arguments and return passable results. 
+
+All of a remotable's property values must be functions, and cannot be accessors. Its methods are called by wrapping 
+the remotable object with [`E`](#e), such as `E(issuer).getBrand()`. As with all calls to `E()`, it results in a Promise.
+
+Every object returned from a smart contract, such a publicFacet or creatorFacet, must be passable. All objects used in 
+your contract's external API must be passable.
```
mhofman, line 438: I feel like there is a bit of repetition here, and slight confusion between remotable and presence.

Wondering if we could structure it the following way:
```suggestion
Remotables are objects containing only methods, intended to be used from other vats. The methods of remotables accept only [passable](#passable) arguments and return passable results. 

A remotable is used in other vats through its proxy called a [presence](#presence). A remotable's methods are called by wrapping the presence with [`E`](#e), such as `E(issuer).getBrand()`.
```

The last line feels more appropriate for the passable section.

@tyg, what do you think? Happy to include this proposed change when working on the other feedback.

erights: Since "Proxy" has a specific meaning in JS, and presences are not JS proxies, please avoid describing them using "proxy" terminology. "Representative" is good. (Our own internal use of "representative" for a different concept should not stop us. We can change our unexplained internal terminology if needed.)

mhofman: there are apparently other uses of the term proxy in relation to presence. I'll double check those.

tyg:
Looks fine modulot MarkM's points.

Tom

main/glossary/README.md
erights: Actually, both the old text and the new text are not quite right on the Remotable/presence terminology. See https://github.com/Agoric/agoric-sdk/pull/2909/files . Warning: Pedantic clarification ahead. Almost certainly too much detail to actually include here. But it should constrain the text we do include here to be correct.

The thing with methods is a PrimaryRemotable. The remote presences are Remotes. Both PrimaryRemotables and Remotes are Remotables. At this level of abstraction, we talk about Remotes, but all Remotes are presences. Not all presences are Remotes though. It is the eventual-send level of abstraction that defines presences. Captp uses presences only to implement Remotes. However, other code is free to use eventual-send to create presences that are not Remotes.

erights: Attn @michaelfig , with whom we worked out the precise terminology.

mhofman:
This PR completely dropped off my radar and is obviously very stale at this point. I'm not sure if some of these updates could be ported to our current documentation, but for now I'll be closing this. cc @erights @Tyrosine22

PR #514, opened 2021-06-04
chore: Patch RESM for emigration

kriskowal:
This will allow Agoric SDK packages to begin migrating away from `node
-r esm`.


katelynsills (approved):
LGTM!

PR #517, opened 2021-06-04
Added parameter and cleaned up other docs around ZCF.makeZCFMint()

tyg:


main/zoe/api/zoe-contract-facet.md
```diff
@@ -62,6 +65,22 @@ ERTP methods on a `ZCFMint` or vice versa.
 do have the same methods as their ERTP-derived counterparts. Assets created by a `zcfMint` are treated
 the same as ERTP `mint`-created assets by ERTP methods.
 
+The following demonstrates `zcf.makeZCFMint`:
+
+**Note**: The call to make the `ZCFMint` is asynchronous, but
+calls to the resulting `ZCFMint` are synchronous.
+```js
+const mySynchronousMint = await zcf.makeZCFMint('MySyncMint', 'set');
```
katelynsills, line 73: We don't want people the string `set` directly, but rather `AssetKind.SET`. Also, the first string will be used as the token name, so we should drop mention of a mint.

```suggestion
const mySynchronousMint = await zcf.makeZCFMint('MyToken', AssetKind.SET);
```

katelynsills (approved):

PR #512, opened 2021-06-01
chore: update packages, including resolving nested requirements to non-vulnerable versions

katelynsills:
Update packages, including resolving nested requirements to non-vulnerable versions

Related to https://github.com/Agoric/agoric-sdk/issues/3007

PR #518, opened 2021-06-05
chore: update packages generally

katelynsills:
Takes the good stuff from #512, without trying to resolve Vuepress V1's dependencies.


dckc (approved):
looks good: updating to current vuepress 1.x, plus updating devDependencies (eslint, ...)

PR #520, opened 2021-06-09
fix: Add language hint for Endo archive generator

kriskowal:
This adds a parsers declaration to package.json, which Endo's compartment mapper recognizes that a `.js` file is a JavaScript module and not a CommonJS module, but unlike `"type": "module"`, does not confuse `node -r esm`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "engines": {
     "node": ">=11.0"
   },
+  "parsers": {
+    "js": "mjs"
+  },
   "scripts": {
     "docs:dev": "vuepress dev main",
     "docs:build": "vuepress build main",


dckc:
I'm missing some context. What are we fixing?

Also, a pointer to a few paragraphs about this module transition stuff would be nice. I sorta have it in my head, but I'd like to confirm.

Maybe I'll try to write something down in the endojs/endo wiki or something.

kriskowal:
> I'm missing some context. What are we fixing?
> 
> Also, a pointer to a few paragraphs about this module transition stuff would be nice. I sorta have it in my head, but I'd like to confirm.
> 
> Maybe I'll try to write something down in the endojs/endo wiki or something.

The root issue is https://github.com/Agoric/agoric-sdk/issues/527. I might create a subtask to capture this part of the migration.

kriskowal:
The other root issue is Use Endo Zip https://github.com/Agoric/agoric-sdk/issues/2684

dckc:
ok, I'm mostly familiar with that context...

How about the symptoms? How do I reproduce the problem that we're fixing here?

dckc (approved):
Looks good, now that I remembered which end is up:

> In order to break the log jam, SESM will respect "type": "module" if present but will read "parsers": {"js": "mjs"}
https://github.com/Agoric/agoric-sdk/issues/2684

PR #522, opened 2021-06-12
docs: How to review a zipped up bundle

kriskowal:
With the imminent change to bundle contracts as base64 encoded Zip files, this prepares the documentation so folks can extract the contract source for review.


katelynsills (approved):
Looks great, thank you for adding this!

PR #523, opened 2021-06-15
refactor: Use import extensions

kriskowal:
This change changes all imports to use extensions in preparation for using ESM proper in Node.js.


katelynsills (approved):
LGTM! Thanks for doing this!

PR #530, opened 2021-06-21
fix: Move zoe to dependencies

kriskowal:
This ensures that bundleSource can find Zoe with the Endo Zip bundle format when that becomes the default.  For this case, I elected to keep the bundleSource example simple rather than thread the `dev` option.


katelynsills (approved):
LGTM!

PR #558, opened 2021-09-20
docs(timerServices): update for latest changes to agoric-sdk

michaelfig:
Refs Agoric/agoric-sdk#3853

As described.

### DIFF
--- a/main/repl/timerServices.md
+++ b/main/repl/timerServices.md
@@ -6,8 +6,8 @@ sidebar: auto
 ## TimerService objects
 
 `home` has two timer service objects: 
-- `localTimerService` is from the local vat.
-- `chainTimerService` is from the chain (if present). 
+- `localTimerService` is from the local vat **with a resolution of 1ms**.
+- `chainTimerService` is from the chain (if present), **with a resolution of 1s**. 
 
 In general, you should use the chain timer for contracts. But it's more efficient to 
 use the local timer for objects that don't need consensus or consistency. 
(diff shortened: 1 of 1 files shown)


main/repl/timerServices.md
```diff
@@ -58,41 +111,26 @@ A `TimerRepeater` has two methods, `schedule()` and `disable`:
...
 Returns the time, in the same format as the parameter,
 at which the call is scheduled to happen. 
 
 ```js
 command[3] handler = harden({ wake: now => { console.log(`woke up ${now}`); }})
 history[3] {"wake":[Function wake]}
-command[4] willWakeAt =  E(home.localTimerService).setWakeup(3, handler)
-history[4] 1342464583
+command[4] willWakeAt =  E(home.localTimerService).setWakeup(1342464580000n + 3_000n, handler)
+history[4] 1342464583000n
```
Chris-Hibbert, line 131: `1342464583000n` is in the range of timestamps I've seen recently from the localTimerService. Contrary to `history[2] 1608523723203n` above.   Why is the local timer producing time stamps that look like milliSeconds since **1979**?

Chris-Hibbert (approved):
The question of what kind of timestamp is being provided by localTimerService (looks like milliseconds since 1979) can be handled separately.

This is a great cleanup.  LGTM

main/repl/timerServices.md
michaelfig: > Why is the local timer producing time stamps that look like milliSeconds since **1979**?

Thanks for catching this.  I'll fix that in this documentation.

That was a miscalculation in the code that existed before #3853.

PR #561, opened 2021-09-25
chore: do not advise users to use Yarn 1

ctjlewis:
I had no issues installing deps, building, and linking the CLI with Yarn 2.


arirubinstein:
Thanks for the PR! - this appears to already have been fixed in main

PR #566, opened 2021-10-26
Documentation for the constantProduct AMM

Chris-Hibbert:
Accompanies https://github.com/Agoric/agoric-sdk/pull/4006.  See also https://github.com/Agoric/agoric-sdk/issues/3862

This documents the current state of the constantProduct AMM, and adds in some discussion of how to interact with governance (Which is being added by the PR above).

### DIFF
--- a/main/zoe/guide/contracts/constantProductAMM.md
+++ b/main/zoe/guide/contracts/constantProductAMM.md
@@ -3,11 +3,16 @@
 <Zoe-Version/>
 
 ##### [View the code on Github](https://github.com/Agoric/agoric-sdk/blob/master/packages/zoe/src/contracts/vpool-xyk-amm/multipoolMarketMaker.js) (Last updated: 2021-10-26)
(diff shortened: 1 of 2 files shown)


main/zoe/guide/contracts/constantProductAMM.md
```diff
@@ -0,0 +1,198 @@
...
+const atmRefund = await E(saraSeat).getPayout('Out');
+
+const BLDProceedsAmount = E(saraBLDPurse).deposit(BLDProceeds);
+E(saraAtmPurse).deposit(atmRefund);
+```
+
+###  Creating a new Pool
+
+When the contract is first instantiated, there won't be any pools ready for
+trading. `addPool()` adds a new currency, which can then be funded. When a pool is
```
dckc, line 108: Document that the issuer passed to `addPool` must be fungible? (use `NAT` MathKind) Or do we suppose the example suffices?

Also, note or refer to risks around trusting issuers?

dckc (approved):
I have a number of suggestions and a few questions, but I think this is good even if all those result in no change.

dckc (approved):

PR #571, opened 2021-11-06
Fix typos

rex4539:


katelynsills (approved):
Dang, this is great work! Thank you so much for doing this.

PR #573, opened 2021-11-08
Update test-readme.js snippet getId

zarutian:
The snippet `getId` does not make sense in the context of the ertp/guide/readme.md as it was.
Updated to be more sensible, plus it now hints at promise pipelining instead of having an unnescisarry `await` in it.
The sooner we get Agoric beginners to internalize promise pipelining the less `await`ing everything has to do.


katelynsills (approved):
Nice work! Thank you so much!

katelynsills:
Hi @zarutian, it looks like there's a test failure in the snippets. Would a change to the test be appropriate?
Screen Shot 2021-11-08 at 1 41 46 PM

PR #574, opened 2021-11-08
chore: bump

katelynsills:
Testing CI - no actual changes

#agoric-sdk-branch: eslint-root


katelynsills:
This PR has served its purpose, and we have identified and fixed the problem. Because the dapps were installing agoric-sdk as a subdirectory of the root, the agoric-sdk packages were going up the directory to the root and using that lint config instead. https://github.com/Agoric/agoric-sdk/pull/4045 fixes the issue.

broad-city-salute

PR #590, opened 2021-12-12
Fix the link to SES Compartment docs

turadg:
The backticks on Compartment were preventing the link from rendering correctly:
Screen Shot 2021-12-12 at 7 03 34 AM

While moving the backticks since the link text would work I moved the link to the use of the term in the sentence.


turadg:
Closing as obsolete

PR #592, opened 2021-12-16
NFTDropContractBasic contract documentation

tgrecojs:
Not sure if this should be opened up as a formal PR, but here's the my initial work as requested from @dtribble during our conversation this past tuesday.

Apologies in advance if I made any slip-ups regarding the proper way to contribute here 😬 🙂

### DIFF
--- a/main/zoe/guide/contracts/nft-drop.md
+++ b/main/zoe/guide/contracts/nft-drop.md
@@ -1,4 +1,4 @@
-∫# NFT Drop
+# NFT Drop
 
 <Zoe-Version/>
 
@@ -15,7 +15,7 @@ The purpose of the `NFTDropContractBasic.js` is to provide a point of reference 
   * Embedded additional metadata relating the project.
   * Whatever your heart desires!
 
-The code for the smart contract can be found below:
+The code for the smart contract is:
 ```js
 /** @type {ContractStartFn} */
 const start = async (zcf) => {
@@ -81,7 +81,7 @@ Once the `mint` has been created, the next step is to use it within the `buyNFTs
   };
 ```
 
-The final step is to create a [Remotable](../../../guides/js-programming/far.md#marshaling-by-copy-or-by-presence) for our public invitation using `zcf`s `makeInvitation` method. 
+The final step is to create a [Remotable](../../../guides/js-programming/far.md#marshaling-by-copy-or-by-presence) that provides the public with invitations on demand using `zcf`s `makeInvitation` method.
 
 ```js
 const publicFacet = Far('NFT Drop', {
@@ -90,7 +90,7 @@ const publicFacet = Far('NFT Drop', {
 
 return harden({ publicFacet });
 ```
-The contract is ready to be instantiated using the `E` as demonstrated in the test code below. In this snippet, the `nftName` for this collection is `PetRocks`, all of which are sold for the same fixed-price. 
+The contract is ready to be instantiated using the `E(zoe).startInstance` as demonstrated in the test code below.In this snippet, the `nftName` for this collection is `PetRocks`, all of which are sold for the same fixed-price.
 
 ```js
   const terms = harden({
@@ -101,7 +101,7 @@ The contract is ready to be instantiated using the `E` as demonstrated in the te
 
   const publicFacet = E(zoe).getPublicFacet(instance);
   // we now have access to the pubic invitation for the `buyNFTs` function.
-  const invitation = E(publicFacet).makeInvitation();
+  const invitation = E(publicFacet).makeBuyInvitation();
 ```
 
 Those who wish to understand more about this contract are encouraged to inspect these test code and run it on their own machine.
--- a/main/zoe/guide/contracts/nft-drop.md
+++ b/main/zoe/guide/contracts/nft-drop.md
@@ -1,0 +1,107 @@
+∫# NFT Drop
+
+<Zoe-Version/>
+
+##### [View the code on Github](https://github.com/Agoric/dapp-nft-drop/blob/main/contract/src/NFTDropContractBasic.js)
+##### [View all contracts on Github](https://github.com/Agoric/agoric-sdk/tree/master/packages/zoe/src/contracts)
+
+This contract provides users with an invitation that will mint a NFT in exchange for a fixed price. It demonstrates how one could use Zoe to facilitate the release, or "drop", of a set oF NFTs created for a particular project. The contract code results in the creation of an [invitation](../../../glossary/#invitation) which can be accessed via the public in order to purchase one of the NFTs in the collection.
+
+The purpose of the `NFTDropContractBasic.js` is to provide a point of reference for developers seeking to launch an NFT drop on the Agoric blockchain. While the implementation details are in complete control of the developer, some features that one may want to extend this contract with are:
+  * Fixing the total supply of NFTs available.
+  * Specifiying [the time](../../../repl/timerServices.md#timer-services):
+    * in which the NFT drop is open to the public.
+    * of an NFT drop deadline, after which users will no longer be able to mint and/or purchase.
+  * Embedded additional metadata relating the project.
+  * Whatever your heart desires!
+
+The code for the smart contract can be found below:
+```js
+/** @type {ContractStartFn} */
+const start = async (zcf) => {
+  const { pricePerNFT, nftName } = zcf.getTerms();
+
+  // Set up the mint
+  const mint = await zcf.makeZCFMint(nftName, AssetKind.SET);
+  const { brand: NFTBrand } = mint.getIssuerRecord();
+
+  let currentId = 1n;
+  const { zcfSeat: sellerSeat } = zcf.makeEmptySeatKit();
+
+  /** @type {OfferHandler} */
+  const buyNFTs = (buyerSeat) => {
+    // Mint the NFTs
+    const amount = AmountMath.make(NFTBrand, harden([currentId]));
+    mint.mintGains(harden({ NFTs: amount }), buyerSeat);
+    currentId += 1n;
+
+    // Take the money
+    sellerSeat.incrementBy(buyerSeat.decrementBy(harden({ Money: pricePerNFT })));
+
+    zcf.reallocate(buyerSeat, sellerSeat);
+    buyerSeat.exit();
+
+    return 'your offer was successful';
+  };
+
+  const publicFacet = Far('NFT Drop', {
+    makeInvitation: () => zcf.makeInvitation(buyNFTs, 'buyNFTs'),
+  });
+
+  return harden({ publicFacet });
+};
+
+```
+
+Taking a look inside of the contract code, you will notice that the minting process occurs **at the time of purchase**. This approach is in contrast to one in which a set of NFTs are minted prior to being made available to the public for purchase. 
+
+To accomplish this design, the contract first creates an internal `mint` variable. Note that the `mint` variable specifies the [`AssetKind`](../../../ertp/api/issuer.md#makeissuerkit-allegedname-assetkind-displayinfo) to be a `SET` which is necessary when creating NFTs with Zoe.
+
+```js
+const mint = await zcf.makeZCFMint(nftName, AssetKind.SET);
+```
+
+Once the `mint` has been created, the next step is to use it within the `buyNFTs` function which will be provided to the the public as an invitation. Additionally, we will see that each NFT has a unique identifier, `currentId`, that is embedded within the NFT at the time it is minted. Although it is intended to be a placeholder, the use of `currentId` is meant to demonstrate how one could add unique properties to an NFT.
+
+```JS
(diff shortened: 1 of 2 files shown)


main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
+∫# NFT Drop
```
dckc, line 1: curious... what is that `∫` thing?

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+
+The purpose of the `NFTDropContractBasic.js` is to provide a point of reference for developers seeking to launch an NFT drop on the Agoric blockchain. While the implementation details are in complete control of the developer, some features that one may want to extend this contract with are:
+  * Fixing the total supply of NFTs available.
+  * Specifiying [the time](../../../repl/timerServices.md#timer-services):
+    * in which the NFT drop is open to the public.
+    * of an NFT drop deadline, after which users will no longer be able to mint and/or purchase.
+  * Embedded additional metadata relating the project.
+  * Whatever your heart desires!
+
+The code for the smart contract can be found below:
```
dckc, line 18: editorial suggestion:

```suggestion
The code for the smart contract is:
```

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  const publicFacet = Far('NFT Drop', {
+    makeInvitation: () => zcf.makeInvitation(buyNFTs, 'buyNFTs'),
+  });
+
+  return harden({ publicFacet });
+};
+
+```
```
dckc, line 54: Hm... how do we keep this in sync with the dapp?

The docs repo has a snippets directory, but that's not appropriate for code in a separate repo.

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+
+Taking a look inside of the contract code, you will notice that the minting process occurs **at the time of purchase**. This approach is in contrast to one in which a set of NFTs are minted prior to being made available to the public for purchase. 
+
+To accomplish this design, the contract first creates an internal `mint` variable. Note that the `mint` variable specifies the [`AssetKind`](../../../ertp/api/issuer.md#makeissuerkit-allegedname-assetkind-displayinfo) to be a `SET` which is necessary when creating NFTs with Zoe.
```
dckc, line 58: `mint` is `const`, so calling it a variable seems odd.

And `AssetKind` is more of an ERTP level thing than Zoe. "with ERTP" is probably safe to just leave implicit.

```suggestion
To accomplish this design, the contract first creates an internal `mint` object. Note that the `mint` variable specifies the [`AssetKind`](../../../ertp/api/issuer.md#makeissuerkit-allegedname-assetkind-displayinfo) to be a `SET` which is necessary when creating NFTs.
```

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+
+```js
+const mint = await zcf.makeZCFMint(nftName, AssetKind.SET);
+```
+
+Once the `mint` has been created, the next step is to use it within the `buyNFTs` function which will be provided to the the public as an invitation. Additionally, we will see that each NFT has a unique identifier, `currentId`, that is embedded within the NFT at the time it is minted. Although it is intended to be a placeholder, the use of `currentId` is meant to demonstrate how one could add unique properties to an NFT.
```
dckc, line 64: the `buyNFTs` function is distinct from an invitation; it's an offer handler associated with an invitation.

```suggestion
Once the `mint` has been created, the next step is to use it within the `buyNFTs` function which will be provided to the the public via an invitation. Additionally, we will see that each NFT has a unique identifier, `currentId`, that is embedded within the NFT at the time it is minted. Although it is intended to be a placeholder, the use of `currentId` is meant to demonstrate how one could add unique properties to an NFT.
```

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+    sellerSeat.incrementBy(buyerSeat.decrementBy(harden({ Money: pricePerNFT })));
+
+    zcf.reallocate(buyerSeat, sellerSeat);
+    buyerSeat.exit();
+
+    return 'your offer was successful';
+  };
+```
+
+The final step is to create a [Remotable](../../../guides/js-programming/far.md#marshaling-by-copy-or-by-presence) for our public invitation using `zcf`s `makeInvitation` method.
```
dckc, line 84: This sounds a bit like `makeInvitation` creates a `Remotable`. Maybe something like:

```suggestion
The final step is to create a [Remotable](../../../guides/js-programming/far.md#marshaling-by-copy-or-by-presence) that provides the public with invitations on demand using `zcf`s `makeInvitation` method. 
```

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+The final step is to create a [Remotable](../../../guides/js-programming/far.md#marshaling-by-copy-or-by-presence) for our public invitation using `zcf`s `makeInvitation` method. 
+
+```js
+const publicFacet = Far('NFT Drop', {
+  makeBuyInvitation: () => zcf.makeInvitation(buyNFTs, 'buyNFTs'),
+});
+
+return harden({ publicFacet });
+```
+The contract is ready to be instantiated using the `E` as demonstrated in the test code below. In this snippet, the `nftName` for this collection is `PetRocks`, all of which are sold for the same fixed-price.
```
dckc, line 93: This suggests that `E` instantiates a contract. How about:

```suggestion
The contract is ready to be instantiated using the `E(zoe).startInstance` as demonstrated in the test code below. In this snippet, the `nftName` for this collection is `PetRocks`, all of which are sold for the same fixed-price. 
```

main/zoe/guide/contracts/nft-drop.md
```diff
@@ -0,0 +1,107 @@
...
+```js
+  const terms = harden({
+    pricePerNFT: AmountMath.make(moneyBrand, 10n),
+    nftName: 'PetRocks',
+  });
+  const { instance } = await E(zoe).startInstance(installation, issuers, terms);
+
+  const publicFacet = E(zoe).getPublicFacet(instance);
+  // we now have access to the pubic invitation for the `buyNFTs` function.
+  const invitation = E(publicFacet).makeInvitation();
```
dckc, line 104: This code would benefit from the testable snippets stuff; it's not in the other repo.

```suggestion
  const invitation = E(publicFacet).makeBuyInvitation();
```

dckc (dismissed):
Thanks for the contribution. I have a few suggestions.

At least one of them is critical: `makeInvitation` vs. `makeBuyInvitation`.

main/zoe/guide/contracts/nft-drop.md
tgrecojs: Accident 😬 will update

main/zoe/guide/contracts/nft-drop.md
tgrecojs: The `const` keyword is used to defined variables in JavaScript with the added behavior of variables defined using `const` can not be redeclared.

I understand it may seem odd, especially with the `var` keyword being the way in which variables were defined in JS for so long. The release of ES6 eliminated wide spread use of using `var` in exchange for `const` as well as `let`.

main/zoe/guide/contracts/nft-drop.md
tgrecojs: Yeah, i was teetering on whether to point out that `buyNFTs` function is a function but more specifically, it's an `offerHandler`. Will update though

main/zoe/guide/contracts/nft-drop.md
tgrecojs: Hmmm, are u saying that there's a currently examples of runnable code snippets in the docs? If so, then I'd gladly translate this code into that as i believe it would be greatly beneficial for developers.

tgrecojs:
> Thanks for the contribution. I have a few suggestions.
> 
> 
> 
> At least one of them is critical: `makeInvitation` vs. `makeBuyInvitation`.

Thanks for the review! I took the liberty of changing the invitation name to `makeBuyInvitation` in an effort to eliminate confusion that i thought could possibly stem from `makeInvitation` being used the `zcf.makeInvitation` method

main/zoe/guide/contracts/nft-drop.md
dckc: > The `const` keyword is used to defined variables in JavaScript ...

Really? I don't recall any JS docs that speak of `const` declarations declaring variables.

See, for example [const in MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) that distinguishes constants from variables:
> Constants are block-scoped, much like variables declared using the let keyword. --

main/zoe/guide/contracts/nft-drop.md
tgrecojs: Nevermind! Seeing the snippets info in the Contributing.md file now 🙂  thanks!

dckc: yes; see [details in CONTRIBUTING.md](https://github.com/Agoric/documentation/blob/main/CONTRIBUTING.md#importing-and-testing-code-snippets).

Setting up the context to test a contract is kinda involved, so it's perhaps better left to a separate PR.

tgrecojs: cool! i actually split it up so the commit i just pushed included all updates besides the snippet. 

reading each other's minds !

main/zoe/guide/contracts/nft-drop.md
tgrecojs: Hmmm... yeah, i never really gave it much thought but this insinuates `const` is not used for declaring variables. Taking a look at the [Declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#declarations) section of the MDN docs shows the following though:


> ### JavaScript has three kinds of variable declarations.
>
> **var**
>
>    Declares a variable, optionally initializing it to a value.
>
> **let**
>
>  Declares a block-scoped, local variable, optionally initializing it to a value.
>
> **const**
>
>    Declares a block-scoped, read-only named constant.

lol very odd right? i'm fine with changing the text, but again, I believe the majority of JS developers interpret `const` as a way to declare variables (at least the ones who have worked primarily in the last decade or so). I guess the question remains - what is the answer here?

tgrecojs: Initially i referenced info i got from glancing at w3schools on Google. 

https://www.w3schools.com/js/js_const.asp

I know Agoric has some representation on TC39. Maybe they could patch up the confusion for us 🤔🙂. 

For right now though i do feel as though the text won't be intepreted oddly by JS devs but again, happy to change it if we think it will cause confusion.

tgrecojs:
This comment addresses @dckc's questions about keeping this in sync with the rest of the docs.

> Hm... how do we keep this in sync with the dapp?
> 
> The docs repo has a snippets directory, but that's not appropriate for code in a separate repo.

I think there are some questions that need answering here before this writing sample makes its way into the live documentation. Does Agoric intend to have the [`NFTDropContractBasic.js`](https://github.com/Agoric/dapp-nft-drop/blob/main/contract/src/NFTDropContractBasic.js)  code  reside in the `@agoric/zoe/contracts`? 

Perhaps there's some fine-tuning needed within the contract code or additional test-cases that need adding (or both), but, FWIW I think that it could act as a solid starting point for developers looking to release an "nft drop" on agoric.

 
If the answer is "No", the process would likely stop here. 

If the answer is "Yes", then I'm assuming the next step would be to create a PR for this to be merged into the Zoe package. And then when that's done, work could resume on merging this text into the docs along with the code snippet discussed.

kbennett2000 (commented):
Coming up to speed on this... Is the issue whether or not to include this contract in @agoric/zoe/contracts? Just want to make sure there's not a bigger issue I'm missing here...

dckc:
> Is the issue whether or not to include this contract in @agoric/zoe/contracts?

That question was asked above, yes. But I think we can treat that as separate from this PR.

This PR is a proposal to document the contract is it stands in the https://github.com/Agoric/dapp-nft-drop repo: https://github.com/Agoric/dapp-nft-drop/blob/main/contract/src/NFTDropContractBasic.js

So as a reviewer, the questions to you are things like:
  - is this something our audience wants to know? There's a cost to maintaining this documentation; is it worth it?
  - Does it make sense to you?
  - If you follow the directions, does it work?
  - Do you see anything technically wrong with it?
  - Are there important editorial issues to fix?

dckc:
This contract is remarkably close to dapp-offer-up. So this PR is sort of overtaken by events.

PR #605, opened 2022-01-10
chore: Fix typos

gibson042:


main/guides/js-programming/notifiers.md
```diff
@@ -152,7 +152,7 @@ Zoe provides updates on the state of seats within a contract. The updates
 from Zoe indicate changes to the allocation of a seat and seats exiting. 
 These are available from `E(userSeat).getNotifier()` and `zcfSeat.getNotifier()`, 
 which provide long-lived notifier objects associated with a particular 
-seat. `zcfSeats` are available within contracts while `userSeats` are accessible 
+seat. `zcfSeat`s are available within contracts while `userSeat`s are accessible
```
gibson042, line 155: These still don't seem to align with https://agoric.com/documentation/zoe/api/zoe-contract-facet.html , perhaps the following would be better?
```suggestion
seat. `ZCFSeat`s are available within contracts while `UserSeat`s are accessible 
```

dckc:
I hope reviewing the notifiers stuff will be quick for @erights ; it will take careful study, for me; that's not feasible today.

main/guides/js-programming/notifiers.md
```diff
@@ -130,16 +129,17 @@ The `updater` and `publication` both have the same three methods:
     signalling that the monitored object hit an error condition.
 
 The `subscription` and `publication` both have this method:
```
erights, line 131: Neither do. This method is only on notifier.

main/guides/js-programming/notifiers.md
```diff
@@ -255,15 +255,15 @@ non-final values due to `NotifierKit()`'s lossy nature.
...
+Either make `NotifierKit()` or `makeSubscriptionKit()` can be used in a multicast
 manner with good distributed systems properties, where there is only one
 producing site but any number of consuming sites. The producer is not vulnerable
 to the consumers; they cannot cause the kit to malfunction or prevent the code
 producing values from making progress. The consumers cannot cause each other to
 hang or miss values. 
 
-For distributed operation, all the iteration values---non-final values,
-successful completion value, failure reason---must be `Passable`; values that
+For distributed operation, all the iteration values—non-final values,
```
erights, line 265: I think there's some way to express an m-dash. If not, triple dash is still a better approximation than single dash.

main/guides/js-programming/notifiers.md
gibson042: So notifiers and subscriptions can both be consumed with `for await`, but only notifiers can be consumed with `.getUpdateSince`?
```suggestion
The `notifier` has an additional method that the `subscription` does not:
```

main/guides/js-programming/notifiers.md
gibson042: This is an em dash, which is hard to see in the GitHub source view but easier in the rendered Markdown: https://github.com/Agoric/documentation/blob/95d1f32a02d75ef5defffdb49d7aedcd6c391d8e/main/guides/js-programming/notifiers.md#distributed-operation

Alternatively, it could be made even more clear in the source at the loss of readability:
```suggestion
For distributed operation, all the iteration values&mdash;non-final values,
```

main/guides/js-programming/notifiers.md
gibson042: @erights ?

erights: The upper case clearly refers to the types, which stands for all the instances of that kind. The lower case is how you'd refer to an instance, which in context clearly also stands for all instances of that kind. I'm ok either way, so whichever you prefer. The main thing is consistency throughout this and related documents.

main/guides/js-programming/notifiers.md
erights: yes

main/guides/js-programming/notifiers.md
erights: I prefer the `&mdash;` in the source. In general, I prefer sources in ascii, so I'm more likely to guess right when interpreting what I'm looking at when reading source. When the thing in the source is clearly non-ascii from how it appears in the source view, that can be ok. I really want to avoid source that looks like ascii but isn't. In fact, it would be nice to have a lint-rule to alert us to those cases.

main/guides/js-programming/notifiers.md
erights: They can also both be consumed with `observeIteration`. We should move towards considering `getUpdateSince` to be an implementation detail, and recommend that consumption happen either by `for await` or `observeIteration`.

main/guides/js-programming/notifiers.md
erights: Yes, that doc clearly establishes that we lean towards the upper case type-name form. I like that.

erights (approved):
LGTM module removing non-ascii characters from source.

Thanks!

Issue #610, opened 2022-01-19
why re2 in yarn build in documentation?


dckc:
```
documentation % yarn why re2
yarn why v1.22.5
[1/4] 🤔  Why do we have the module "re2"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "re2@1.16.0"
info Reasons this module exists
   - "@agoric#zoe#@agoric#swingset-vat" depends on it
   - Hoisted from "@agoric#zoe#@agoric#swingset-vat#re2"
   - Hoisted from "@agoric#zoe#@agoric#transform-metering#re2"
info Disk size without dependencies: "17.02MB"
info Disk size with unique dependencies: "20.59MB"
info Disk size with transitive dependencies: "26.8MB"
info Number of shared dependencies: 59
✨  Done in 0.28s.
```

cc @turadg @warner @michaelfig

michaelfig:
Please `agoric install beta` on the documentation repo, it looks really stale.

turadg:
> Please agoric install beta on the documentation repo, it looks really stale.

Should repo `main` be using beta? https://github.com/Agoric/documentation/commit/1c1b20cb1b119def155ce00173f044c31d99e7fe

michaelfig:
> Should repo `main` be using beta? [1c1b20c](https://github.com/Agoric/documentation/commit/1c1b20cb1b119def155ce00173f044c31d99e7fe)

Yes, the `main` documentation site documents the Agoric SDK `beta` version.

Tyrosine22:
We no longer reference the beta version of the Agoric SDK. Does that mean this bug is now stale & irrelevant?

dckc:
Yes, it's ready to close, because it's fixed:

```
~/projects/documentation$ yarn why re2
yarn why v1.22.5
[1/4] Why do we have the module "re2"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
error We couldn't find a match!
Done in 1.36s.
```

This is about the build system for the documentation: why did it depend on the re2 package, which involves a C++ extension? Evidently not for any good reason. re2 is no longer a dependency.

PR #611, opened 2022-01-20
chore: source deps from agoric beta

turadg:
As @michaelfig [pointed out](https://github.com/Agoric/documentation/issues/610#issuecomment-1016961708), the documentation site documents the beta SDK.


michaelfig (approved):
LGTM!  Thanks for doing this.

PR #632, opened 2022-03-08
fix: Correct and clarify agoric cli documentation

gibson042:
* Remove a page replaced by HTTP redirect
* Document correct default `agoric init` dapp-template
* Improve the explanation of `agoric start` arguments
* Document the need for `cd` or similar between `agoric init` and `agoric install`
* Rename "required argument" to "positional argument" and "optional argument" to "option"


JimLarson (approved):
New content looks good! I know the removed file was not being served, but there might be some good content in it that should be rescued. Before merging, please make a pass and rescue any content that isn't already covered elsewhere - or just reply that you've made such a scan and found nothing.

gibson042:
I confirmed that no content is being lost.

PR #635, opened 2022-03-10
style: Italicize in-page links

gibson042:
I think this makes the [Glossary](https://agoric.com/documentation/glossary/) more readable, and on other pages has a negligible effect.

Before | After
-- | --
glossary (before) | glossary (after)
Treasury contract (before) | Treasury contract (after)

### DIFF
--- a/main/.vuepress/config.js
+++ b/main/.vuepress/config.js
@@ -15,6 +15,9 @@ module.exports = {
     }
     .two-col-table table {
         table-layout: fixed;
+    }
+    a[href^='#'] {
+        font-style: italic;
     }`],
     ['script', { src: 'https://www.googletagmanager.com/gtag/js?id=UA-118217811-1' }],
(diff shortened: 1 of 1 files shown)


kriskowal:
I’m +½. Italicizing terms in a glossary seems idiomatic to me, though @turadg’s point stands.

gibson042:
All links are already bold and red, which I'm not proposing to change (although I'm also not opposed if the resulting pattern is coherent). But I _am_ trying to differentiate in-page links from cross-page links, and reached for italics as a style that is a) available, and b) conventional inside a glossary-like document (although I'm open to other options as well).

> I defer to whomever is responsible for documentation UX

As do I. Do we know who that is?

Chris-Hibbert:
I don't think anyone currently has ownership.

turadg:
>> I defer to whomever is responsible for documentation UX
> As do I. Do we know who that is?

@tatyana0219 said she'll take a look

tatyana0219:
LGTM.

tatyana0219 (approved):
LGTM

PR #637, opened 2022-03-10
refactor: Remove now-superfluous babel-standalone

kriskowal:
- refactor: Remove the now-superfluous babel-standalone
- chore: Update yarn.lock

I have run `yarn-deduplicate` here since there was certainly some nonsensical duplication of compatible semver range packages, but there remain some duplicates because of skew between the most recent Endo and Agoric SDK releases, that I assume will settle when an Agoric SDK release goes up with the latest Endo dependencies.

Refs https://github.com/Agoric/agoric-sdk/issues/4785

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,19 +36,18 @@
   "homepage": "https://github.com/Agoric/documentation#readme",
   "dependencies": {
     "@agoric/zoe": "beta",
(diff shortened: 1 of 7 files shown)


turadg (approved):
👍

PR #638, opened 2022-03-11
fix: Clean up the glossary

gibson042:
Mostly editorial changes to increase readability and internal consistency and addition of hyperlinks, but there are a few fixes of e.g. broken link targets and invalid code.

This is probably best reviewed using the rich diff, at least for a first pass.

### DIFF
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -40,9 +40,9 @@ five tickets is performed by combining arrays rather than by arithmetic.
 `AmountMath` has a single set of polymorphic methods that deal with two different asset kinds:
 - `AssetKind.NAT`: Used with [fungible](#fungible) assets. Each amount value is a natural number (non-negative integer). This is the default `AssetKind`.
 - `AssetKind.SET`: Used with [non-fungible](#non-fungible) assets. Each amount value is an
-  array of comparable values (strings, numbers, objects, etc.).
-  Values should never include promises (they aren't comparable), or
-  data that should not be shared freely such as payments and purses.
+  array of [comparable](#comparable) values (strings, numbers, objects, etc.).
+  Values cannot include promises (they aren't comparable), and should not
+  include privileged objects such as payments and purses.
 
 For more information, see the [ERTP Guide's AmountMath section](/ertp/guide/amount-math.md)
 and the [ERTP API's AmountMath section](/ertp/api/amount-math.md).
@@ -121,13 +121,12 @@ Destroy all digital assets in a payment. See [`issuer.burn(payment, optAmount)`]
 
 ## Comparable
 
-A *passable* is something that can be marshalled. A *comparable* is a
-passable whose leaves contain no promises. Two comparables can be
-synchronously compared for structural equivalence.
+A *passable* is something that can be marshalled (see the
+[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence)).
 
-A comparable is a JavaScript object containing no promises, and can
-thus be locally compared for equality with another object. If either object
-contains Promises, equality is indeterminable. If both are fulfilled down
+A *comparable* is a passable containing no promises or errors, and can
+thus be synchronously compared for structural equivalence with another object. If either object
+contains promises and/or errors, equality is indeterminable. If both are fulfilled down
 to Presences and local state, then either they're the same all the way
 down, or they represent different objects.
 
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -36,8 +36,8 @@ disappear from the total allocation.
 ## AmountMath
 
 The AmountMath library executes the logic of how [amounts](#amount) are changed when digital assets are merged, separated,
-or otherwise manipulated. For example, a deposit of 2 [Quatloos](#quatloos) into a purse that already has 3 Quatloos
-gives a new balance of 5 Quatloos. But a deposit of a non-fungible theater ticket into a purse that already holds
+or otherwise manipulated. For example, a deposit of 3 [Quatloos](#quatloos) into a purse that already has 7 Quatloos
+updates its balance to 10 Quatloos. But a deposit of a non-fungible theater ticket into a purse that already holds
 five tickets is performed by set union rather than by arithmetic.
 
 `AmountMath` has a single set of polymorphic methods that deal with different asset kinds:
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -382,8 +382,8 @@ See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord, offerArgs)`](/zoe
 ## Offer Safety
 
 Zoe guarantees offer safety. When a user makes an [offer](#offer) and its payments are [escrowed](#escrow) with Zoe, Zoe guarantees that
-the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and
-escrowed. One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
+the user either gets what they said they wanted, or gets back what they originally offered and
+escrowed (i.e., a refund). One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
 can immediately cause the [seat](#seat) to exit, getting back the amount it offered.
 
 ## Payment
@@ -425,8 +425,8 @@ const myProposal = harden({
   exit: { onDemand: null }
 })
 ```
-`give` and `want` use [keywords](#keywords) defined by the contract. Each specifies via an [amount](#amounts), a description of what
-asset they are willing to give/want to get, and how much of it.
+`give` and `want` use [keywords](#keywords) defined by the contract. Each specifies via an [amount](#amounts) a description of what
+they are willing to give or want to get.
 
 ## Purse
 A purse holds [amounts](#amounts) of assets issued by a particular [mint](#mint) that are all of the same [brand](#brand) and generally _stationary_.
@@ -459,7 +459,7 @@ the [UserSeat documentation](/zoe/api/zoe.md#userseat-object).
 
 ## SES
 
-We have renamed SES to [Hardened JavaScript](#hardened-javascript-ses).
+Secure ECMAScript has been renamed to [Hardened JavaScript](#hardened-javascript-ses).
 
 ## Simoleons
 An imaginary currency Agoric documentation uses in examples.
@@ -476,11 +476,11 @@ it set at $10. They can specify the instance's minimum bid amount in its terms.
 
 Values are the part of an [amount](#amounts) that describe the value of something
 that can be owned or shared: How much, how many, or a description of a unique asset, such as
-Pixel(3,2), $3, or 'Seat J12 for the show September 27th at 9:00pm'.
+Pixel(3,2), $3, or “Seat J12 for the show September 27th at 9:00pm”.
 [Fungible](#fungible) values are usually
 represented by natural numbers. Other values may be represented as strings naming a particular
 rights, or an array of arbitrary objects representing the rights at issue. The latter two examples
-are usually [non-fungible](#non-fungible) assets. Values must be [Comparable](#comparable).
+are usually [non-fungible](#non-fungible) assets. Values must be [comparable](#comparable).
 
 For more information, see the [ERTP Guide's Value section](/ertp/guide/amounts.md#values).
 
@@ -494,7 +494,7 @@ For more information, see the [Vat section in the Distributed JS Programming Gui
 ## Wallet
 
 The overall place a party keeps their assets of all brands. For example, your wallet might contain 5 Quatloos
-[purses](#purse), 8 Moola purses, and 2 Simoleons purses. A wallet can distinguish between [Issuers](#issuer).
+[purses](#purse), 8 Moola purses, and 2 Simoleons purses. A wallet can distinguish between [issuers](#issuer).
 Dapps can propose [offers](#offer) to a wallet. If a user accepts the offer proposal,
 the wallet makes an offer on the user's behalf and deposits the [payout](#payout) in the user's [purses](#purse).
 See the [Wallet Guide and API](/guides/wallet/).
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -120,14 +120,11 @@ Destroy all digital assets in a payment, for example as part of consuming it in 
 
 ## Comparable
 
-A *passable* is something that can be marshalled (see the
-[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence)).
-
-A *comparable* is a passable containing no promises or errors, and can
-thus be synchronously compared for structural equivalence with another object. If either object
-contains promises and/or errors, equality is indeterminable. If both are fulfilled down
-to Presences and local state, then either they're the same all the way
-down, or they represent different objects.
+A *comparable* is a [passable](#passable) containing no promises or errors, and can
+thus be synchronously compared for structural equivalence with another piece of data.
+If either side of the comparison contains promises and/or errors, equality is indeterminable.
+If both are fulfilled down to [presences](#presence) and local state, then either they're the
+same all the way down, or they represent different objects.
 
 ## Contract and Contract Instance
 In Agoric documentation, *contract* usually refers to a contract's source code that
@@ -384,6 +381,11 @@ Zoe guarantees offer safety. When a user makes an [offer](#offer) and its paymen
 the user either gets what they said they wanted, or gets back what they originally offered and
 escrowed (a refund). One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
 can immediately cause the [seat](#seat) to exit, getting back the amount it offered.
+
+## Passable
+
+A *passable* is something that can be marshalled (see the
+[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence)).
 
 ## Payment
 
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -160,7 +160,7 @@ See [here](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/do
 ## E()
 
 (Also referred to as *eventual send*) `E()` is a local "bridge" function that invokes methods on remote objects, for example
-in another vat, machine, or blockchain. It takes a local representative (a proxy) for a remote object as an argument and
+in another vat, machine, or blockchain. It takes a local representative (a [presence](#presence)) for a remote object as an argument and
 sends messages to it using normal message-sending syntax. The local proxy forwards all messages to the remote object to deal with.
 All `E()` calls return a promise for the eventual returned value. For more detail, see
 the [`E()` section in the Distributed JavaScript page](/guides/js-programming/eventual-send.md).
@@ -272,7 +272,7 @@ the invitation is a [`Payment`](#payment), and so is associated with a specific 
 
 To participate in a contract instance by making an [offer](#offer), an invitation to that instance must accompany the offer.
 
-An `invitation`'s amount includes the following properties:
+An invitation's [amount](#amounts) includes the following properties:
 - The contract's installation in Zoe, including access to its source code.
 - The contract instance this invitation is for.
 - A handle used to refer to this invitation.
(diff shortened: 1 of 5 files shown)


main/glossary/README.md
```diff
@@ -109,27 +107,26 @@ The installation operation returns
 an `installation`, which is an object with one method; `getBundle()`. You can access an installed contract's source
 code via `const { source } = await E(installation).getBundle();`.
 In many cases, the bundled source is a single reviewable string.
-In others, the bundle contains to base 64 encoded zip file that you can
+In others, the bundle contains a base64-encoded zip file that you can
```
kriskowal, line 110: It might be worthwhile to shift the emphasis here. Base64-encoded Zip files are the norm now, not the exception.

main/glossary/README.md
```diff
@@ -109,27 +107,26 @@ The installation operation returns
 an `installation`, which is an object with one method; `getBundle()`. You can access an installed contract's source
 code via `const { source } = await E(installation).getBundle();`.
```
kriskowal, line 108: This probably needs to change to `const { endoZipBase64 } = await E(installation).getBundle(); console.log(endoZipBase64)` and down below `echo "$endoZipBase64" | base64 -d > bundle.zip` for clarity about how this transits from JS to shell.

main/glossary/README.md
```diff
@@ -109,27 +107,26 @@ The installation operation returns
...
+```sh
+echo "$bundle_source" | base64 -d > bundle.zip
 unzip bundle.zip
 ```
 
 
 ## Burn
 
-Destroy all digital assets in a payment. See [`issuer.burn(payment, optAmount)`](/ertp/api/issuer.md#issuer-burn-payment-optamount).
+Destroy all digital assets in a payment, e.g. as part of consuming it in an exchange. See [`issuer.burn(payment, optAmount)`](/ertp/api/issuer.md#issuer-burn-payment-optamount).
```
kriskowal, line 120:
```suggestion
Destroy all digital assets in a payment, for example as part of consuming it in an exchange. See [`issuer.burn(payment, optAmount)`](/ertp/api/issuer.md#issuer-burn-payment-optamount).
```

main/glossary/README.md
```diff
@@ -389,13 +382,13 @@ See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord, offerArgs)`](/zoe
 ## Offer Safety
 
 Zoe guarantees offer safety. When a user makes an [offer](#offer) and its payments are [escrowed](#escrow) with Zoe, Zoe guarantees that
-the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and
-escrowed. One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
+the user either gets what they said they wanted, or gets back what they originally offered and
+escrowed (i.e., a refund). One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
```
kriskowal, line 386:
```suggestion
escrowed (a refund). One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
```

kriskowal (approved):
Please consider comments. We should in general avoid Latin where English will suffice. English is burden enough.

gibson042:
@kriskowal Thanks; suggestions applied.

main/glossary/README.md
```diff
@@ -33,19 +33,16 @@ disappear from the total allocation.
...
-methods of two different asset kinds to deal with [fungible](#fungible) assets (values are natural numbers) and
-[non-fungible](#non-fungible) assets (values are an array or object). The two `AssetKinds` are
-- `AssetKind.NAT`: Used with fungible assets. Amount values are natural numbers (non-negative integers). Default value.
-- `AssetKind.SET`: Used with non-fungible assets. Amount values are
-  arrays of strings, numbers, objects, or other comparable values.
-  Values should never include promises (they aren't comparable), or
-  payments, purses, and anything else that can't be shared freely.
+or otherwise manipulated. For example, a deposit of 2 [Quatloos](#quatloos) into a purse that already has 3 Quatloos
+gives a new balance of 5 Quatloos. But a deposit of a non-fungible theater ticket into a purse that already holds
+five tickets is performed by combining arrays rather than by arithmetic.
```
erights, line 38:
```suggestion
five tickets is performed by set union rather than by arithmetic.
```

main/glossary/README.md
```diff
@@ -33,19 +33,16 @@ disappear from the total allocation.
...
+
+`AmountMath` has a single set of polymorphic methods that deal with two different asset kinds:
```
erights, line 40:
```suggestion
`AmountMath` has a single set of polymorphic methods that deal with different asset kinds:
```
Add sections below on AssetKind.COPY_SET and AssetKind.COPY_BAG . Mention that AssetKind.SET is recently deprecated in favor of AssetKind.COPY_SET but still in wide use. AssetKind.COPY_BAG is for semi-fungible.

main/glossary/README.md
```diff
@@ -87,51 +84,47 @@ or `BigInt(123)`.
...
 unzip bundle.zip
 ```
 
 
 ## Burn
 
-Destroy all digital assets in a payment. See [`issuer.burn(payment, optAmount)`](/ertp/api/issuer.md#issuer-burn-payment-optamount).
+Destroy all digital assets in a payment, for example as part of consuming it in an exchange. See [`issuer.burn(payment, optAmount)`](/ertp/api/issuer.md#issuer-burn-payment-optamount).
 
 ## Comparable
```
erights, line 121:
```suggestion
## Key
```
The `Comparable` type is now just a deprecated name for `Key`. All mentions of `Comparable` should be changed to `Key`.

main/glossary/README.md
```diff
@@ -87,51 +84,47 @@ or `BigInt(123)`.
...
-synchronously compared for structural equivalence.
-
-A comparable is a JavaScript object containing no promises, and can
-thus be locally compared for equality with another object. If either object
-contains Promises, equality is indeterminable. If both are fulfilled down
-to Presences and local state, then either they're the same all the way
-down, or they represent different objects.
+A *comparable* is a [passable](#passable) containing no promises or errors, and can
+thus be synchronously compared for structural equivalence with another piece of data.
+If either side of the comparison contains promises and/or errors, equality is indeterminable.
```
erights, line 125: Do you want to mention that we're talking about comparisons at the `compareKeys` level of abstraction, or leave that implicit since that's the only one normal users should be exposed to?

main/glossary/README.md
```diff
@@ -87,51 +84,47 @@ or `BigInt(123)`.
...
+If both are fulfilled down to [presences](#presence) and local state, then either they're the
```
erights, line 126:
```suggestion
If both are fulfilled down to [remotables](#remotables) and local state, then either they're the
```
In making this suggestion, I didn't check if `#remotables` links to anything.

main/glossary/README.md
```diff
@@ -87,51 +84,47 @@ or `BigInt(123)`.
...
+same all the way down, or they represent different objects.
 
 ## Contract and Contract Instance
```
erights, line 129:
```suggestion
## Contract Installation and Contract Instance
```
and use that terminology consistently below. I prefer to avoid synecdoche when speaking precisely.

main/glossary/README.md
```diff
@@ -146,7 +139,7 @@ If they have ten houses for sale, they have ten different contract instances.
 
 ## CreatorInvitation
```
erights, line 140: This identifier, with the leading and midcap, does not appear anywhere else. I suggest just two words, so it's clear it is prose rather than a code identifier.
```suggestion
## Creator Invitation
```

main/glossary/README.md
```diff
@@ -163,15 +156,15 @@ See [here](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/do
 ## E()
 
 (Also referred to as *eventual send*) `E()` is a local "bridge" function that invokes methods on remote objects, for example
```
erights, line 158: We use `E` to send to local and remote objects. The key features are that the designated object *can be* remote, and that the invocation is always asynchronous, to be handled in its own later turn. The expression returns a promise for the invocation's result.

erights (changes requested):

main/glossary/README.md
gibson042: The page doesn't have a Remotable heading, but aren't non-presence remotables already covered here by "and local state"?

main/glossary/README.md
```diff
@@ -33,19 +33,27 @@ disappear from the total allocation.
...
+  subject to the same constraints as those of `AssetKind.COPY_SET`.
+- `AssetKind.COPY_SET`: Used with [non-fungible](#non-fungible) assets.
+  Each amount value is a set of [key](#key) values
+  (strings, numbers, objects, etc.).
+  Values cannot include promises (they aren't keys), and should not
+  include privileged objects such as payments and purses.
+- `AssetKind.COPY_BAG`: Used with semi-fungible assets.
+  Each amount value is a [multiset](https://en.wikipedia.org/wiki/Multiset)
+  of [key](#key) values subject to the same constraints as
+  those of `AssetKind.COPY_SET` but allowed to be present more than once.
```
gibson042, line 56: @erights I've added the text for COPY_SET and COPY_BAG. Please (re-)review, and if possible suggest some text for "semi-fungible" (which is currently absent, a glaring omission here where we link to "fungible" and "non-fungible").

erights: ## Semi-fungible
A semi-fungible asset is one where there are distinct forms of the asset, which are not interchangeable with each other, but in which there may be multiple instances of each form which are interchangeable with other instances of the same form. As a variation on the [non-fungible](#non-fungible) theater ticket example, perhaps the tickets only specify a section, where the holder may sit in any seat of that section. For each section, the number of theater tickets minted should not exceed the number of seats in that section.

erights: Waddaya think?

main/glossary/README.md
```diff
@@ -87,53 +95,45 @@ or `BigInt(123)`.
 The Board is a shared, on-chain location where users can post a value and make it
 accessible to others. When a user posts a value, they receive a unique ID for the value.
 Others can get the value just by knowing the ID. You can make an ID known by any
-communication method; private email, a DM or other private message, a phone call/voicemail,
-an email blast to a mailing list or many individuals, listing it on a website, etc.
+communication method&mdash;a DM or email or other private message, a phone call/voicemail,
+an email blast to a mailing list, publishing it on a website, etc.
```
erights, line 99: Note that these IDs are publicly visible, so publishing an object in the board makes it publicly accessible. If this is not appropriate for the object you have in mind, do not use the board to communicate access to that object.

main/glossary/README.md
```diff
@@ -87,53 +95,45 @@ or `BigInt(123)`.
...
 ```
 The installation operation returns
 an `installation`, which is an object with one method; `getBundle()`. You can access an installed contract's source
-code via `const { source } = await E(installation).getBundle();`.
-In many cases, the bundled source is a single reviewable string.
-In others, the bundle contains to base 64 encoded zip file that you can
-extract for review.
-```
-jq -r .endoZipBase64 bundle.json | base64 -d > bundle.zip
+code via `const { endoZipBase64 } = await E(installation).getBundle();`.
```
erights, line 119: Suggest putting this code in its own block
```js
  const { endoZipBase64 } = await E(installation).getBundle();
```

main/glossary/README.md
```diff
@@ -162,16 +162,18 @@ See [here](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/do
 
 ## E()
 
-(Also referred to as *eventual send*) `E()` is a local "bridge" function that invokes methods on remote objects, for example
-in another vat, machine, or blockchain. It takes a local representative (a proxy) for a remote object as an argument and
-sends messages to it using normal message-sending syntax. The local proxy forwards all messages to the remote object to deal with.
-All `E()` calls return a promise for the eventual returned value. For more detail, see
+(Also referred to as *eventual send*) `E()` is a local "bridge" function that
+asynchronously invokes methods on local or remote objects, for example those
+in another vat, machine, or blockchain. It takes a local object or a [presence](#presence)
```
erights, line 167: or a promise for a local or remote object

main/glossary/README.md
```diff
@@ -219,7 +221,7 @@ See [`E()`](#e) above.
 Part of an [offer](#offer) specifying how the offer can be cancelled/exited. There are three values:
 - `onDemand: null`: (Default) The offering party can cancel on demand.
 - `waived: null`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
-- `afterDeadline`: The offer is automatically cancelled after a deadline, as determined by its timer and deadline properties.
+- `afterDeadline: {…}`: The offer is automatically cancelled after a deadline, as determined by its `deadline` and `timer` properties.
```
erights, line 224: Maybe expand the nested record to show the timer and deadline?

main/glossary/README.md
```diff
@@ -266,16 +268,12 @@ The Inter-Blockchain Communication protocol, used by blockchains to communicate
...
-return a creator invitation on their instantiation, in case the contract instantiator wants
+A [payment](#payment) whose amount represents (and is required for) participation in a contract instance.
+Contracts often return a creator invitation on their instantiation, in case the contract instantiator wants
 to immediately participate. Otherwise, the contract instance must create any additional invitations.
-These, or any invitation held by a party, are distributed by sending it to someone's wallet. When you receive
-an invitation, your wallet will validate it via the [InvitationIssuer](#invitationissuer). Note that
-the invitation is a [`Payment`](#payment), and so is associated with a specific [`Issuer`](#issuer).
-
-To participate in a contract instance by making an [offer](#offer), an invitation to that instance must accompany the offer.
+Every [offer](#offer) to participate in a contract instance must include an invitation to that instance in its payment, and any wallet receiving one will validate it via the [InvitationIssuer](#invitationissuer).
```
erights, line 274:
```suggestion
Every [offer](#offer) to participate in a contract instance must include an invitation to that instance as the first `offer` argument, and any wallet receiving one will validate it via the [InvitationIssuer](#invitationissuer).
```

main/glossary/README.md
```diff
@@ -309,6 +307,14 @@ its asset type is valid.
 For more information, see the [ERTP Guide's Issuer section](/ertp/guide/issuers-and-mints.md)
 and the [ERTP API's Issuer section](/ertp/api/issuer.md).
 
+## Key
+
+A *key* is a [passable](#passable) containing no promises or errors, and can
+thus be synchronously compared for structural equivalence with another piece of data.
+If either side of the comparison contains promises and/or errors, equality is indeterminable.
+If both are fulfilled down to [presences](#presence) and local state, then either they're the
+same all the way down, or they represent different objects.
```
erights, line 316: Keys can be used as elements of CopySets and CopyBags, and as keys of CopyMaps. Amounts can only contain Keys.

main/glossary/README.md
```diff
@@ -309,6 +307,14 @@ its asset type is valid.
...
+
 ## Keywords
 
 Keywords are unique identifiers per contract. They tie together the [proposal](#proposal), [payments](#payment)
```
erights, line 320: This explanation is just wrong. Maybe something like

Keywords are property names strings which are valid identifiers, and which start with an upper case letter. A KeywordRecord is a CopyRecord in which all the properties are named by Keywords.

Please rephrase to taste.

main/glossary/README.md
```diff
@@ -318,34 +324,34 @@ records with values of [amounts](#amounts), [issuers](#issuer), etc.
...
 - ERTP mints create digital assets and are the only ERTP objects with the authority to do so.
   Access to an ERTP mint gives you the power to create more digital assets of its type at will. Mints
   can only create one type of asset and cannot change to create a different type.
 
-  ERTP mints are [issuer's](#issuer) admin [facets](#facet), and there is a one-to-one relationship between an issuer and
-  its mint. ERTP mints are also in a one-to-one relationship with that issuer's associated [brand](#brand).
+  There is a one-to-one relationship between an [issuer](#issuer) and its mint, and a mint is an administrative [facet](#facet) of its issuer.
+  ERTP mints are also in a one-to-one relationship with that issuer's associated [brand](#brand).
 
 - ZCFMints give contract code a simpler interface to interact with an ERTP mint. Because ZCFMints encapsulate
```
erights, line 337:
```suggestion
- ZCFMints give contract code a simpler interface to interact with an ERTP mint from inside a contract. Because ZCFMints encapsulate
```

main/glossary/README.md
```diff
@@ -318,34 +324,34 @@ records with values of [amounts](#amounts), [issuers](#issuer), etc.
...
   an internal ERTP mint, they have the same one-to-one relationships
   with an issuer and its associated brand. A ZCFMint can mint assets and assign them to a seat without having to escrow
-  payments, and burn assets that used to be associated with a seat without having to payout assets.
+  payments, and burn seat-associated assets without corresponding payout.
 
 ZCFMints and ERTP mints do **not** have the same methods. Do not try to use ERTP methods on a ZCFMint or vice versa.
```
erights, line 342: I can't decide whether it is helpful to point out that a ZCFMint actually wraps an ERTP Mint. Do whatever seems best.

main/glossary/README.md
```diff
@@ -389,13 +395,18 @@ See [`E(Zoe).offer(invitation, proposal, paymentKeywordRecord, offerArgs)`](/zoe
...
 Zoe guarantees offer safety. When a user makes an [offer](#offer) and its payments are [escrowed](#escrow) with Zoe, Zoe guarantees that
-the user either gets what they said they wanted, or gets back (gets a refund) what they originally offered and
-escrowed. One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
+the user either gets what they said they wanted, or gets back what they originally offered and
+escrowed (a refund). One reason this is possible is if a [proposal](#proposal) doesn't match what the contract expects to do, it
 can immediately cause the [seat](#seat) to exit, getting back the amount it offered.
 
+## Passable
+
+A *passable* is something that can be marshalled (see the
```
erights, line 404: and sent as arguments to remote objects.

main/glossary/README.md
```diff
@@ -425,20 +436,19 @@ For more information, see the [JavaScript Distributed Programming Guide](/guides
 Proposals are records with `give`, `want`, and `exit` keys. [Offers](#offer) must include a proposal, which states
```
erights, line 436:
```suggestion
Proposals are records with `give`, `want`, and `exit` properties. [Offers](#offer) must include a proposal, which states
```

main/glossary/README.md
```diff
@@ -425,20 +436,19 @@ For more information, see the [JavaScript Distributed Programming Guide](/guides
...
-`give` and `want` use [keywords](#keywords) defined by the contract. Each specifies via an [amount](#amounts), a description of what
-asset they are willing to give/want to get, and how much of it.
+`give` and `want` use [keywords](#keywords) defined by the contract. Each specifies via an [amount](#amounts) a description of what
+they are willing to give or want to get.
 
 ## Purse
-Purses hold [amounts](#amounts) of a certain [mint](#mint) issued assets. Specifically amounts that are _stationary_.
-Purses can transfer part of their held balance to a [payment](#payment), which is usually used to transfer value.
-A purse's contents are all of the same [brand](#brand).
+A purse holds [amounts](#amounts) of assets issued by a particular [mint](#mint) that are all of the same [brand](#brand) and generally _stationary_.
```
erights, line 450: I don't understand what "_stationary_" is supposed to mean here.

main/glossary/README.md
```diff
@@ -484,11 +494,11 @@ it set at $10. They can specify the instance's minimum bid amount in its terms.
 
 Values are the part of an [amount](#amounts) that describe the value of something
 that can be owned or shared: How much, how many, or a description of a unique asset, such as
-Pixel(3,2), $3, or 'Seat J12 for the show September 27th at 9:00pm'.
+Pixel(3,2), $3, or “Seat J12 for the show September 27th at 9:00pm”.
 [Fungible](#fungible) values are usually
 represented by natural numbers. Other values may be represented as strings naming a particular
```
erights, line 499: Please rephrase in terms of CopySets.

erights (approved):
Already made review comments individually. All are minor and local, so LGTM. Thanks!

erights (approved):

main/glossary/README.md
gibson042: I opted instead for a link to https://agoric.com/documentation/zoe/api/zoe.html#proposals-and-payments .

main/glossary/README.md
gibson042: I think it's intended as a contrast to [usually short-lived] payments. Reworded to "…often for arbitrarily long periods of time".

gibson042:
@erights Updated, some suggestions taken directly and others tweaked slightly. Would you like to re-review?

PR #640, opened 2022-03-12
Use Netlify for deployment

michaelfig:
Deploy this repo to a separate documentation site.

This reduces coupling between https://agoric.com and https://agoric.com/documentation/


_redirects
```diff
@@ -0,0 +1,2 @@
+# Display a 404 error page for things we don't recognize.
+/documentation/* https://static-agoric-com.netlify.app/404.html 200
```
dckc, line 2: Why 200?

dckc (commented):
Why 200 for the 404 thing?

dckc (approved):
Ok. Looks good.

PR #641, opened 2022-03-12
fix: forward to main site for errors

michaelfig:
Break the forwarding loop with a temporary redirect.


dckc (approved):
What a pain. Whatever works.

PR #643, opened 2022-03-14
style: Use Markdown code fences for all shells in "Starting A Project"

gibson042:
We shouldn't give up syntax highlighting so easily. This PR uses a different [Prism syntax highlighting theme](https://github.com/PrismJS/prism-themes) for secondary shells, but we could alternatively use different styling, e.g. colored borders.

Before | After
-- | --
before | after


main/.vuepress/styles/code-secondary.styl
```diff
@@ -0,0 +1,325 @@
+div[class*="language-"].secondary {
+  background-color: transparent;
+  code {
+    color: #111b27;
+  }
+
+  /* https://github.com/PrismJS/prism-themes/blob/master/themes/prism-coldark-cold.css */
```
dckc, line 7: It's hard to tell how much novel CSS code this proposes.

How about moving the stuff that came from `prism-coldark-cold.css` to its own file?

dckc (commented):
I'm not inclined to take on maintenance of this much CSS. Perhaps other are? Or perhaps we can make it more clear that we're "vendoring" most of this CSS code?

main/.vuepress/styles/code-secondary.styl
gibson042: This is the "its own file"; I couldn't find a way to use it completely unmodified because the styles must be scoped to just the secondary blocks (the wrapping `div[class*="language-"].secondary`). Everything from this line down is copied directly, but I'd welcome a better pattern if anyone has sufficient SASS expertise to suggest it.

dckc: > I couldn't find a way to use it completely unmodified ...

one approach I sometimes use is to check it in unmodified and then commit the change necessary to make it work in our context. I'm not sure how well that would work here, but maybe it's worth a try?

tatyana0219 (approved):
LGTM

PR #650, opened 2022-03-26
fix: Cleanup the "JavaScript Framework" pages

gibson042:
* Remove meaningless spaces (and therefore best viewed with [whitespace changes ignored](650/files?w=1) or as a rich diff)
* Fix typos
* Improve outbound links
* Improve notifier/subscription explanations
* Clarify the PassStyle classifications (please pay particular attention to the removal of a redundant `"promise"`, which is currently in both "special cases" and "so-called `PassableCap` leafs")


main/guides/js-programming/far.md
```diff
@@ -54,16 +54,12 @@ data leads to behavior across vats that is straightforward to reason about.
...
+It is classified as one of the following `PassStyle` values:
+   * atomic pass-by-copy primitives (`"undefined" | "null" |
      "boolean" | "number" | "bigint" | "string" | "symbol"`),
-   * the pass-by-copy containers (`"copyArray" | "copyRecord"`) that
-     contain other Passables,
-   * and the special cases (`"error" | "promise"`), which
-     also contain other Passables.
+   * pass-by-copy containers that contain other Passables (`"copyArray" | "copyRecord"`),
+   * special cases, which also contain other Passables (`"error"`).
    * so-called `PassableCap` leafs (`"remotable" | "promise"`).
```
gibson042, line 63: I took a guess that "promise" belongs in this category rather than "special cases" because it doesn't actually contain other Passables, but please suggest otherwise if the guess was incorrect.

main/repl/timerServices.md
```diff
@@ -77,8 +77,8 @@ history[32] 1632170321864n // about 3 seconds later, in milliseconds since epoch
 - Returns: `{ Notifier<Timestamp> }`
 
 Creates and returns a `Notifier` object. It repeatedly delivers updates at times
-that are a multiple of the passed-in `interval` value, with the first update happening
-the value of `delay` after the notifier is created.
+that are a multiple of the provided `interval` value, with the first update happening
+after the provided `delay` value.
```
gibson042, line 81: This fails to define or link to the semantics of RelativeTime (units, shape, etc.), and therefore of `delay` and `interval`. But it should do so or else the documentation is of little value. So, what _is_ a relative time and how is it expressed?

main/guides/js-programming/notifiers.md
```diff
@@ -47,8 +47,7 @@ The key difference between the two is
 - `subscriptions` are lossless.
    - The consumer will see every value in the sequence.
 
-If your consumers only care about more recent states, use a `NotifierKit`.  For
-consumers that need to see all the values, use a `SubscriptionKit`. Subscriptions
+If your consumers only care about more recent states, use `makeNotifierKit()`. Notifiers
```
dckc, line 50: These notifier / subscription changes look like more than typos / editorial. They look like (good) substantive fixes. A more clear commit message would be nice.

main/guides/js-programming/far.md
dckc: I'm pretty sure I copied this from JSDoc that @erights wrote, so I'm not in a position to approve a change.

dckc (commented):
I defer to @erights on the PassStyle change.

The rest looks good.

gibson042:
Thanks @dckc. I updated main/repl/timerServices.md to:
* clarify the nature of Timestamp/RelativeTime
* correct the documented function signatures
* improve consistency across descriptions
* correct outdated REPL examples

Please (re-)review that file.

main/repl/timerServices.md
```diff
@@ -133,71 +128,62 @@ history[4] 1632170402207n
...
-history[22]  50
-command[23]  h1.getCalls()
-history[23]  53
-command[24]  h1.getCalls()
-history[24]  54
-command[25]  E(tl).getCurrentTimestamp()
-history[25]  1571783375000n
-command[26]  E(tc).getCurrentTimestamp()
-history[26]  1571783384n
+command[22]  Promise.resolve(E(tl).getCurrentTimestamp()).then(t => [t, c1.getCount()])
```
dckc, line 185: Why add `Promise.resolve()`?

gibson042: In case it was a mistake for `getCurrentTimestamp` to return a promise rather than a bigint. If we are committed to the former, then I can drop this `Promise.resolve`.

main/repl/timerServices.md
```diff
@@ -21,54 +21,48 @@ Both timer services let you:
...
+has elapsed.
 
-```
-command[32] E(home.localTimerService).delay(3_000n)
-history[32] unresolved Promise
-history[32] 1632170321864n // about 3 seconds later, in milliseconds since epoch
+```js
+command[3] E(home.localTimerService).delay(3_000n)
+history[3] unresolved Promise
+history[3] 1632170321864n // about 3 seconds later, in milliseconds since epoch
```
dckc, line 65: any particular reason to change these numbers?

main/repl/timerServices.md
```diff
@@ -93,36 +87,37 @@ These are more complex objects that are only used for the following advanced met
 
 ### `TimerWaker` object
 
-A `TimerWaker` has one method, `wake`, which takes a `Timestamp` argument and returns `void`. 
-The passed `timestamp` is the time when the call to `wake()` is scheduled to occur.
+A `TimerWaker` has one method, `wake()`
+- `wake(timestamp)`
+  - `timestamp` `{ Timestamp }`
+  - Returns `undefined`
```
dckc: `undefined` is more specific than [void](https://www.typescriptlang.org/docs/handbook/2/functions.html#void). This is more than an editorial change. Is that on purpose?

main/repl/timerServices.md
gibson042: Which I see on reinspection is necessarily the case because that's how `E()` works. Fixing now.

dckc (approved):
Reviewing code that doesn't go through ci feels a little funny, but I don't see any serious problems.

main/repl/timerServices.md
gibson042: No big reason, just for continuity with the previous block and consistency with sibling documents.

main/repl/timerServices.md
gibson042: No, and I realized at about the same time that you posted this so it's already fixed.

PR #651, opened 2022-03-26
fix: Cleanup ERTP Introduction

gibson042:
Extends #638, so is probably not reviewable until that merges.


main/getting-started/intro-zoe.md
```diff
@@ -158,16 +157,14 @@ proposal:
 
 Proposals must use Keywords, which are capitalized ASCII keys. Here,
```
gibson042, line 158: @erights This differs from the definition we settled on in [#638](https://github.com/Agoric/documentation/pull/638):
> A keyword is a property name string that starts with an upper case letter and is a valid
[identifier](https://developer.mozilla.org/en-US/docs/Glossary/Identifier).

Is this ASCII constraint accurate?

erights: My mistake. I just looked, and the ascii constraint is accurate.

gibson042: No worries, I just added the fix to this PR: https://github.com/Agoric/documentation/pull/651/commits/77a6067a43e6299e1c4e6530df080862ffb808ff

main/getting-started/ertp-introduction.md
```diff
@@ -72,29 +72,29 @@ We refer to
...
 An `issuer`'s special admin facet is a `Mint`, and that `Mint` and `Issuer`
 have a one-to-one relationship. With a reference to an `Issuer`, you can
 check the validity of a `payment` in that `issuer`'s assets;
 i.e. If you have a reference to the Quatloos `issuer`, you can validate
-any `payment` made in Quatloos. You can also claim the `payment` either 
+any `payment` made in Quatloos. You can also claim the `payment` either
 as a new `payment` to yourself or a `purse` you control.
 
-`Issuers` should be gotten from a trusted source
+`Issuers` should be retrieved from a trusted source
```
erights, line 88:
```suggestion
`Issuers` should be obtained from a trusted source
```
No objective best answer at this level of vagueness, but how does "obtained" sound? The term "retrieved" sounds like it was the response to some kind of retrieval request initiated by this party, which is too specific.

main/getting-started/intro-zoe.md
```diff
@@ -94,8 +93,8 @@ E() is part of the Agoric platform and is used to [call methods on
 remote objects and receive a promise for the
 result](/guides/js-programming/eventual-send.md).
 Code on the Agoric platform is put in separate environments, called
-[vats](/glossary/#vat), for security. Zoe is in a different vat,
-making it a remote object, so we must use E().
+[vats](/glossary/#vat), for security. Zoe is a remote object in its own vat,
```
erights, line 96: Nice!

main/getting-started/intro-zoe.md
```diff
@@ -257,27 +254,24 @@ be compared to other installations, and you can call
...
 she uses the keyword `Price` to label the `simoleanIssuer`. When you
-create a new instance of the contract ([see permalink to line
-58](https://github.com/Agoric/agoric-sdk/blob/23c3f9df56940230e21a16b4861f40197192fdea/packages/zoe/src/contracts/atomicSwap.js#L58))
-by calling `startInstance()`, it returns a `creatorInvitation`. Alice
-uses this to make her offer; even the instance's creator needs to have
-an invitation to it to participate in it.
+create a new instance of the contract
+by calling `startInstance()`, it returns a `creatorInvitation`
+(even the instance's creator needs to have
+an invitation to participate in it).
```
erights, line 263:
```suggestion
by calling `startInstance()`, it returns a `creatorInvitation`.
(Even the instance's creator needs to have
an invitation to participate in it.)
```
In this case, the parenthetical is a complete sentence, so I'd end the previous sentence and make it a proper sentence. I think. Don't follow my advice if it violates writing norms.

erights (changes requested):
All your changes LGTM (modulo only minor suggestions). The only significant issue is an old error I just now spotted: The use of "burn" when talking about destroying objects rather than assets/erights.

Thanks for doing this!

erights:
> "end the lifecycle of a payment, making it unavailable for later use and either destroying or transferring its assets".

Sorry, but no. I think we can often speak about "burning a payment" as shorthand for "burning the assets in the payment and using up / destroying the payment object." When that fits, the greater detail is not needed. But if the assets are transferred rather than destroyed, no burning happened.

If you want to split the overall changes into two PRs, fine. But the first incomplete PR should insert visible TODOs in the documentation explaining that "burning" is wrong and a further revision is coming.

gibson042:
I opened #654 and will start working on it now. In the meantime, approving this PR will allow me to merge it and minimize the risk of conflicts.

erights (approved):
LGTM

PR #656, opened 2022-04-04
chore: Synchronize public networking docs with agoric-sdk

gibson042:
Synchronize with https://github.com/Agoric/agoric-sdk/blob/HEAD/packages/SwingSet/docs/networking.md (after https://github.com/Agoric/agoric-sdk/pull/4949 )


michaelfig (approved):
LGTM!

PR #697, opened 2022-08-15
fix: patch last remaining text

arirubinstein:
One text blob was missing the new disclaimer test - this fixes it.


jessysaurusrex (approved):
LGTM

PR #718, opened 2022-09-20
shell consolidate

cboydstun:
brought together some steps for terminal 1 and put some things in order a little better

### DIFF
--- a/main/getting-started/start-a-project.md
+++ b/main/getting-started/start-a-project.md
@@ -4,13 +4,10 @@
 The Agoric platform is at the beta stage. It is in the process of being formally evaluated for vulnerabilities, and is undergoing security treatment and hardening to support mainnet-1 launch. Do not use for production purposes. 
 :::
 
-Now that you have [installed the Agoric SDK](/getting-started/before-using-agoric.md),
-let's try out your first Agoric _Dapp_ (decentralized application).
+Do not proceed unless you've [installed the Agoric SDK](/getting-started/before-using-agoric.md).
+If `agoric --version` is available, then you're ready for our first Agoric _Dapp_ (decentralized application)!
 
-We'll be running **three terminal windows**, and
-[dapp template initialization](#initialize-demo-from-dapp-template) in the first
-window will create their shared working directory.
-
+We'll be running **three terminal windows**. See below: 
 
  1. ```sh
     # Terminal 1: simulated blockchain and "solo" client
@@ -35,28 +32,19 @@ This presentation includes starting a project, but note an outdated detail:
 Use the [Agoric CLI](/guides/agoric-cli/commands.md) to fetch from a Dapp template
 and put it in a `demo` directory _not located in your `agoric-sdk` clone_:
 
-
 ```sh
 # Terminal 1
-# Don't use your agoric-sdk clone as the parent of the demo directory.
-# It doesn't have to be your home directory; any other directory will do.
+# Don't use your agoric-sdk as the parent of the demo directory.
 cd $HOME
-agoric init demo
-```
-
-The name `demo` is an arbitrary suggestion; in general,
-use `agoric init $DIRNAME` with any name you like.
-
-The default template is the [Fungible Faucet Dapp](https://github.com/Agoric/dapp-fungible-faucet).
-Learn more about the [available dapp templates](/dapps/dapp-templates.md).
-
-### Install the Agoric SDK in the Dapp
-
-```sh
-# Terminal 1
+agoric init demo # use `agoric init $DIRNAME` with any name you like
 cd demo
 agoric install
+agoric start --verbose # `agoric start --reset` to start over
 ```
+
+The name `demo` is an arbitrary suggestion. The default template is the [Fungible Faucet Dapp](https://github.com/Agoric/dapp-fungible-faucet).
+
+Learn more about the [available dapp templates](/dapps/dapp-templates.md).
 
 It may take a minute or so to install all the dependencies.
 
@@ -65,40 +53,15 @@ On a Mac, you must first install
 [Xcode](https://apps.apple.com/us/app/xcode/id497799835)
 :::
 
-## Start the Agoric Solo Client and Simulated Blockchain
-
-```sh
-# Terminal 1
-cd demo # if not already there
-agoric start
-
-# if the process was interrupted or crashed
-# rerun the command above and it should resume
-
-# to reset and start over
-agoric start --reset
-```
-
 Leave this process and its logs running in its own terminal window.
-## Deploy the Contract and API
 
-In a separate terminal, deploy the contract to the simulated blockchain
-and the API to the solo client.
 
-```sh secondary style2
-# Terminal 2
-cd demo # if not already there
-agoric deploy ./contract/deploy.js ./api/deploy.js
-```
-
-We'll cover [deploying smart contracts](/getting-started/deploying.md)
-in detail later.
 
 ## Open the Agoric Wallet and REPL
 
 ```sh secondary style2
 # Terminal 2
-cd demo # if not already there
+cd demo # new terminal window
 agoric open --repl
 ```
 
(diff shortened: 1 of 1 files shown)


dckc:
any idea why the cloudflare bot isn't doing a preview deploy?

main/getting-started/start-a-project.md
```diff
@@ -4,13 +4,10 @@
 The Agoric platform is at the beta stage. It is in the process of being formally evaluated for vulnerabilities, and is undergoing security treatment and hardening to support mainnet-1 launch. Do not use for production purposes. 
 :::
 
-Now that you have [installed the Agoric SDK](/getting-started/before-using-agoric.md),
-let's try out your first Agoric _Dapp_ (decentralized application).
-
-We'll be running **three terminal windows**, and
-[dapp template initialization](#initialize-demo-from-dapp-template) in the first
-window will create their shared working directory.
+Do not proceed unless you've [installed the Agoric SDK](/getting-started/before-using-agoric.md).
```
dckc, line 7: Starting the page with "Do not" seems sort of jarring. Any particular reason for that change?

main/getting-started/start-a-project.md
```diff
@@ -35,70 +32,36 @@ This presentation includes starting a project, but note an outdated detail:
 Use the [Agoric CLI](/guides/agoric-cli/commands.md) to fetch from a Dapp template
 and put it in a `demo` directory _not located in your `agoric-sdk` clone_:
 
-
 ```sh
 # Terminal 1
-# Don't use your agoric-sdk clone as the parent of the demo directory.
-# It doesn't have to be your home directory; any other directory will do.
```
dckc, line 42: why drop this line?

main/getting-started/start-a-project.md
```diff
@@ -35,70 +32,36 @@ This presentation includes starting a project, but note an outdated detail:
...
-agoric init demo
+agoric init demo # use `agoric init $DIRNAME` with any name you like
+cd demo
+agoric install
+agoric start --verbose # `agoric start --reset` to start over
 ```
 
-The name `demo` is an arbitrary suggestion; in general,
-use `agoric init $DIRNAME` with any name you like.
+The name `demo` is an arbitrary suggestion. The default template is the [Fungible Faucet Dapp](https://github.com/Agoric/dapp-fungible-faucet).
```
dckc, line 45: Seems like a non-sequitor. The fact that other names can be used in place of `demo` is unrelated to the choice of dapp.

main/getting-started/start-a-project.md
```diff
@@ -35,70 +32,36 @@ This presentation includes starting a project, but note an outdated detail:
...
 
-### Install the Agoric SDK in the Dapp
-
-```sh
-# Terminal 1
-cd demo
-agoric install
-```
-
 It may take a minute or so to install all the dependencies.
```
dckc, line 49: I wonder about moving "It may take a minute or so..." up as a shell comment.

main/getting-started/start-a-project.md
```diff
@@ -35,70 +32,36 @@ This presentation includes starting a project, but note an outdated detail:
...
 ```
 
-The name `demo` is an arbitrary suggestion; in general,
-use `agoric init $DIRNAME` with any name you like.
+The name `demo` is an arbitrary suggestion. The default template is the [Fungible Faucet Dapp](https://github.com/Agoric/dapp-fungible-faucet).
 
-The default template is the [Fungible Faucet Dapp](https://github.com/Agoric/dapp-fungible-faucet).
 Learn more about the [available dapp templates](/dapps/dapp-templates.md).
 
-### Install the Agoric SDK in the Dapp
```
dckc, line 53: This heading seems important to help the reader build a mental model of the parts of the system; in particular, that a dapp depends on the SDK but is a separate thing. Do you expect to explain that somewhere else? Or do you think we don't need it?

main/getting-started/start-a-project.md
```diff
@@ -35,70 +32,36 @@ This presentation includes starting a project, but note an outdated detail:
...
-```
-
 It may take a minute or so to install all the dependencies.
 
 ::: tip Mac Dev Tools
 On a Mac, you must first install
 [Xcode](https://apps.apple.com/us/app/xcode/id497799835)
 :::
 
-## Start the Agoric Solo Client and Simulated Blockchain
```
dckc, line 68: Likewise, this heading seems important for building a mental model. It seems like a PR that deletes this heading is incomplete without adding some explanation elsewhere.

main/getting-started/start-a-project.md
```diff
@@ -35,70 +32,36 @@ This presentation includes starting a project, but note an outdated detail:
...
 
-We'll cover [deploying smart contracts](/getting-started/deploying.md)
-in detail later.
 
 ## Open the Agoric Wallet and REPL
 
 ```sh secondary style2
 # Terminal 2
-cd demo # if not already there
+cd demo # new terminal window
```
dckc, line 64: "new terminal window" seems redundant with "Terminal 2" and the color clues. Does it seem worthwhile to have that much repetition?

main/getting-started/start-a-project.md
```diff
@@ -108,14 +71,31 @@ To begin using the wallet, click the "Connect Solo Wallet" button.
...
+
 ## Start the Dapp User Interface
 
 The web user interface communicates with the API in the solo client as well as the wallet.
 
 ```sh secondary style3
 # Terminal 3
 cd demo # if not already there
-cd ui && yarn start
+cd ui && yarn start
```
dckc, line 98: stray space?

dckc (commented):
In general, shorter is better, but I hesitate to skip over the distinctions in the headings without seeing a change somewhere else to make those distinctions.

cboydstun:
> any idea why the cloudflare bot isn't doing a preview deploy?

I see the Cloudflare preview [here](https://github.com/Agoric/documentation/pull/718/checks?check_run_id=8441013128)

main/getting-started/start-a-project.md
cboydstun: I actually wanted sort of jarring... this entire page is a non-starter unless they've got `agoric` available from their terminal and that seems to be a point of friction for many, but of course we can soften the language.

main/getting-started/start-a-project.md
cboydstun: it's still there just consolidated and bumped up to line 37 - also i do not believe we need two more lines inside of the code block for it immediately after mentioning it above the code block as well...

main/getting-started/start-a-project.md
cboydstun: broke that up a little in next push

main/getting-started/start-a-project.md
cboydstun: i would argue that we do not need it as it is covered under the consolidated "Initialized `demo` from Dapp Template` section.

dckc: The only mention I see in the revised version of that section is "Don't use your agoric-sdk as the parent of the demo directory." I don't see how that establishes in the readers mind that the SDK and the dapp are separate things, where one gets installed as a dependency of the other.

main/getting-started/start-a-project.md
cboydstun: I added a short section below the sub heading "Initialize `demo`..." that summarizes these two lost headings.

dckc: ah. good.

main/getting-started/start-a-project.md
dckc: But if they follow the docs in the normal order, they'll have `agoric` installed, yes?

In any case, I much prefer that we phrase it positively; for example: "Use `agoric --version` to double-check that you are done [installing the Agoric SDK](...)". If people are still tripping over it, I suggest we use a call-out box rather than starting with "Do not ...".

main/getting-started/start-a-project.md
cboydstun: i've amended the introduction into three sentences at the top of "start a project" to clarify the distinction... maybe that helps?

main/getting-started/start-a-project.md
cboydstun: more positive it is now! <(°.°)>

PR #724, opened 2022-11-08
Adding a needed link redirect

Tyrosine22:


cboydstun (approved):
looks good to me... I found the [redirect](https://agoric.com/documentation/getting-started/before-using-agoric.html) I was looking for =)

PR #730, opened 2022-11-17
add smart contract definition

cboydstun:
adding the definition of a smart contract re: @dckc from our TDD sesh from the [contract we did yesterday](https://gist.github.com/dckc/0988dfe516c24a1e925a1d68b3a69217)


dckc (approved):
yup.

PR #741, opened 2022-12-23
chore: Miscellaneous improvements to ERTP API pages

gibson042:
* Alphabetize ERTP Data Types sections
* Use consistent phrasing for "Both **Amount** arguments must have the same **Brand**" and "an error is thrown"
* Use consistent `// <Capitalized word> …` code comment style
* Format code with `…`
* Remove implications that AmountMath methods ever return their actual arguments (rather than equivalent but distinct values)
* Remove empty lines at end of file

### DIFF
--- a/main/reference/ertp-api/amount-math.md
+++ b/main/reference/ertp-api/amount-math.md
@@ -25,8 +25,7 @@ not equal, an error is thrown and no changes are made.
 Creates an **Amount** from a given **Brand** and **AmountValue**.
 
 ```js
-// amount837 = { brand: quatloosBrand, value: 837n }
-const amount837 = AmountMath.make(quatloosBrand, 837n);
+const bid = AmountMath.make(quatloosBrand, 300n);
 ```
 
 ## AmountMath.coerce(brand, allegedAmount)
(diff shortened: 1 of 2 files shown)


main/reference/ertp-api/amount-math.md
```diff
@@ -183,21 +176,21 @@ AmountMath.isEqual(empty, quatloos10);
...
 If the optional *brand* argument doesn't match the **Amount**s' **Brand**, an error is thrown.
 
 For fungible **Amounts** this means adding their **[Values](./ertp-data-types.md#value)**. For non-fungible
 **Amounts**, it usually means including all of the elements from *leftAmount*
 and *rightAmount*.
 
-If either *leftAmount* or *rightAmount* is empty, this method returns the non-empty 
-**Amount** argument. If both are empty, this method returns an empty **Amount**.
+If either argument is empty, this method returns an **Amount** equivalent to the other argument.
+If both are empty, this method returns an empty **Amount**.
```
Tyrosine22, line 190: There are actually 3 arguments, so my wording in this instance is clearer.

main/reference/ertp-api/amount-math.md
```diff
@@ -213,11 +206,11 @@ const combinedList = AmountMath.add(listAmountA, listAmountB);
 
 Returns a new **Amount** that is the *leftAmount* minus the *rightAmount* (i.e., 
 everything in the *leftAmount* that is not in the *rightAmount*). If *leftAmount* 
-doesn't include *rightAmount* (i.e., subtraction results in a negative), this method throws an 
-error. Because *leftAmount* must include *rightAmount*, this is **not** 
+doesn't include *rightAmount* (e.g., subtraction results in a negative), an error is thrown.
+Because *leftAmount* must include *rightAmount*, this is **not** 
 equivalent to set subtraction.
```
Tyrosine22, line 211: Why e.g.? If the leftAmount doesn't include rightAmount, won't that always means that subtracgtion will result in a negative?

Tyrosine22 (changes requested):

main/reference/ertp-api/amount-math.md
gibson042: No, because Amounts are not limited to numbers. For example, subtracting the [set](https://docs.agoric.com/reference/ertp-api/ertp-data-types.html#assetkind) {1, 2} from the set {0, 1} does not result in a negative but still throws an error.

main/reference/ertp-api/amount-math.md
Tyrosine22: Yup. Makes sense.

main/reference/ertp-api/amount-math.md
Tyrosine22: K.

Tyrosine22 (approved):

ivanlei (approved):
🛳️

ivanlei:
closing & reopening to kick CI

gibson042:
Replaced by #916

PR #764, opened 2023-01-12
fix: add syncing information

arirubinstein:
Migrates syncing up docs into docs - where they should be!

### DIFF
--- a/main/guides/getting-started/syncing-up.md
+++ b/main/guides/getting-started/syncing-up.md
@@ -1,0 +1,86 @@
+# Syncing Up on Mainnet
+
+To run a node on the `agoric-3` network from scratch, please follow the 
+steps below.
+
+If you wish to restore from a recent community snapshot, please follow the instructions from the heading [Current Community Snapshot](#current-community-snapshot).
+
+| Current |     Upgrade Name     | Upgrade Height |                  Git Hash                |    Repo    |          Docker Tag        |
+| ------- |--------------------- | -------------- | ---------------------------------------- |    ag0     |             N/A            |
+|         | agoric-3.1 (genesis) |     2,115,669  | a2ebcdb21418bb157f9c747a042b2a859b2a5986 |    ag0     |             N/A            |
+|         | agoric-upgrade-5     |     3,565,000  | a2ebcdb21418bb157f9c747a042b2a859b2a5986 |    ag0     |             N/A            |
+|         | agoric-upgrade-6     |     5,901,622  | 31c78ba3aa872b54c4de448763c5b8044b8f950c |    ag0     |             N/A            |
+|         | agoric-upgrade-7-2   |     6,263,783  | f4759c9f15b869c453f847a63ba734cacb9a991a |    ag0     |             N/A            |
+|         | agoric-upgrade-8     |     7,179,262  | 2c812d22161cd297587979b262eab6e2cc76e23d |    agd     |    agoric/agoric-sdk:29    |
+| Current | agoric-upgrade-8-1   |     7,179,262  | 08ca9d4fd8413da59b73d53e12851fe00583ddc1 |    agd     |    agoric/agoric-sdk:30    |
+
+
+### agoric-3.1
+
+1. `git clone https://github.com/Agoric/ag0.git && cd ag0`
+1. `git checkout tags/agoric-3.1`
+1. `make install`
+1. `ag0 init follow --chain-id agoric-3`
+1. `curl https://main.rpc.agoric.net/genesis | jq .result.genesis > $HOME/.agoric/config/genesis.json`
+1. `ag0 start`
+
+   chain will halt at 3,565,000
+
+### agoric-upgrade-5
+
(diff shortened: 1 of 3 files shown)


main/guides/getting-started/syncing-up.md
```diff
@@ -0,0 +1,86 @@
...
+After more than 10 blocks have successfully synced, stop the `agd` process 
+and update to `agoric-upgrade-8-1`, a soft upgrade.  This soft upgrade includes 
+logic to handle a memory condition that may lead to your node falling out of consensus. 
+It is recommended to upgrade to `agoric-upgrade-8-1` as soon as you have successfully 
+produced a handful of blocks with the `agoric-upgrade-8` upgrade.
+
+### agoric-upgrade-8-1
+#### Current Community Snapshot
+
+1. `git clone https://github.com/agoric/agoric-sdk.git;  agoric-sdk && git checkout tags/agoric-upgrade-8-1`
```
dckc, line 81: I was going to ask if we're comfortable giving instructions on using a community snapshot without "use at your own risk" warnings... but this doesn't seem to include instructions about getting a community snapshot.

dckc (commented):
@Tyrosine22 this proposal expands the scope of these docs from just smart contract development to also operating a node, so I suppose I should check with you before approving.

Likewise, @michaelfig this looks like a proposal to start moving validator docs from https://github.com/Agoric/agoric-sdk/wiki/Validator-Guide to this docs repo, so I suppose I should check that you agree that it's time to do so.

main/guides/getting-started/syncing-up.md
```diff
@@ -0,0 +1,86 @@
+# Syncing Up on Mainnet
+
+To run a node on the `agoric-3` network from scratch, please follow the 
+steps below.
+
+If you wish to restore from a recent community snapshot, please follow the instructions from the heading [Current Community Snapshot](#current-community-snapshot).
```
dckc, line 6: We should probably let folks know that if they choose not to use a community snapshot, they should expect to spend a few days catching up.

main/guides/getting-started/syncing-up.md
```diff
@@ -0,0 +1,86 @@
...
+|         | agoric-upgrade-8     |     7,179,262  | 2c812d22161cd297587979b262eab6e2cc76e23d |    agd     |    agoric/agoric-sdk:29    |
+| Current | agoric-upgrade-8-1   |     7,179,262  | 08ca9d4fd8413da59b73d53e12851fe00583ddc1 |    agd     |    agoric/agoric-sdk:30    |
+
+
+### agoric-3.1
+
+1. `git clone https://github.com/Agoric/ag0.git && cd ag0`
+1. `git checkout tags/agoric-3.1`
+1. `make install`
+1. `ag0 init follow --chain-id agoric-3`
```
dckc, line 23: I take it `init follow` means there's no precious validator key, so we don't need to give loud warnings about backing up key material, right?

main/guides/getting-started/syncing-up.md
```diff
@@ -0,0 +1,86 @@
+# Syncing Up on Mainnet
+
+To run a node on the `agoric-3` network from scratch, please follow the
```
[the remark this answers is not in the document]
arirubinstein: This is an attempt to start documenting something that is needed today.  We should continue to expand this documentation over time and talk more about mainnet.  Right now, there is no documentation about mainnet.

main/guides/getting-started/syncing-up.md
arirubinstein: We have no alternatives to offer users besides our GitHub as a source repo, so how would a user make an informed risk decision worth highlighting here?

main/guides/getting-started/syncing-up.md
arirubinstein: Correct

main/guides/getting-started/syncing-up.md
dckc: I guess I don't understand what this section is for. I thought it was about downloading a snapshot and resuming from there. If somebody follows the directions in this section, I don't understand what they should expect to get as a result.

We could point them at https://github.com/Agoric/validator-profiles/wiki/Snapshots , which has a **Use at your own risk.** disclaimer at the top. As noted there, discussion of this topic goes back to Feb in https://github.com/Agoric/testnet-notes/issues/42 . Or we could migrate that info to community.agoric.com ... I gather that in the cosmos community, sharing validator profiles and "look what I did for the community" is more traditionally done in discourse forums than github wikis. We could start a thread where we don't give any snapshot pointers, but we invite snapshot providers to give pointers.

arirubinstein: This is "how to run the most recent version of the code with the latest upgrade" if you bring along current state from a snapshot

Tyrosine22 (approved):
There are several aspects of this PR I don't care for, most importantly the fact that we mention "Mainnet" in the title with no context or explanation. But until I think of a better way to handle the problematic elements (e.g., how best to introduce & discuss Agoric versioning) I'll approve this PR as-is.

PR #775, opened 2023-02-16
Docs: Add usage of the Agoric VS Code extension

danBamikiya:
@cboydstun please take a look

cc: @dckc

### DIFF
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -8,7 +8,7 @@ After installing the Agoric SDK, you can proceed to [starting a project](./start
 
 ### Automated Install
 
-If you're using VS Code as your code editor, install the official [Agoric Extension](https://marketplace.visualstudio.com/items?itemName=danbmky.agoric-vscode) to automate the install process:
+If you're using VS Code as your code editor, install the official [Agoric Extension](https://marketplace.visualstudio.com/items?itemName=danbmky.agoric-vscode) to automate the Agoric SDK install process:
 
 - Open the **Extensions** sidebar in VS Code. `View → Extensions`
 - Search for `Agoric`, choose "Agoric" by **danbmky**
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -6,6 +6,16 @@ After installing the Agoric SDK, you can proceed to [starting a project](./start
 
 ## Quick Start
 
+### Automated Install
+
+If you're using VS Code as your code editor, install the official [Agoric Extension](https://marketplace.visualstudio.com/items?itemName=danbmky.agoric-vscode) to automate the install process:
+
+- Open the **Extensions** sidebar in VS Code. `View → Extensions`
+- Search for `Agoric`, choose "Agoric" by **danbmky**
+- Click **Install** to install it
+
+### Manual Install
+
 If you're familar with JavaScript development tools such as `node`, `yarn`, and `git`:
 
 ```shell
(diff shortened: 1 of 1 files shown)


dckc:
For context, note https://github.com/Agoric/agoric-sdk/issues/5156 . I wonder why that's still open.

main/guides/getting-started/README.md
```diff
@@ -6,6 +6,16 @@ After installing the Agoric SDK, you can proceed to [starting a project](./start
 
 ## Quick Start
 
+### Automated Install
+
+If you're using VS Code as your code editor, install the official [Agoric Extension](https://marketplace.visualstudio.com/items?itemName=danbmky.agoric-vscode) to automate the Agoric SDK install process:
```
dckc, line 11: I wonder if "official" is how we should describe it. That might suggest that it's supported by the Agoric engineering team, and I'm not sure we're there yet.

In particular, https://github.com/Agoric/agoric-sdk/issues/6419 seems to be outstanding and is out for a bounty rather than being in the engineering queue.

dckc (commented):
I'd like input from some other folks.

main/guides/getting-started/README.md
danBamikiya: @dckc thanks for the review. [Agoric/agoric#5156](https://github.com/Agoric/agoric-sdk/issues/5156) is already completed and should be closed I think (cc: @jeetraut).

also, [Agoric/agoric-sdk#6419](https://github.com/Agoric/agoric-sdk/issues/6419) has long been completed and the bounty has been paid too, so I guess it should be closed too (cc: @cboydstun).

I used the word "official" to describe it because there's already another extension on the VS Code marketplace (and registry) with the description "agoric sdk install" and it has some downloads already. So, to not confuse users on which to download, I thought "official" would better describe it.

Screenshot 2023-02-16 at 10 24 02 PM

and "official" also suggests that the extension is maintained, which I do, I published the version 1.0.7 of the Agoric Extension today to both the VS Code Registry and the Open VSX Registry, this recent version converts the extension into a "Workspace Extension" from a "UI Extension" allowing the extension to also be downloadable and run on both Gitpod and GitHub Codespaces platform, supporting more users! 🎉

---
Agoric Extension running on Gitpod:
Screenshot 2023-02-16 at 10 44 03 PM

---
Agoric Extension running on GitHub Codespaces:
Screenshot 2023-02-16 at 10 48 38 PM


*Side Note:*
Agoric Extension as seen on the Open VSX Registry that allows it to be downloaded on Gitpod:
image


we could go for another description though.

dckc: wow! very nice work!

danBamikiya: thanks! 🥳

arirubinstein:
For this to be "official", we would need to own the extension and the Agoric namespace in the VSCode store, so in the mean time I'm ok with referring to it as a community resource.

danBamikiya:
@dckc I have removed the 'official' description as adviced by @arirubinstein

ivanlei:
While the extension is quite cool, these docs target an developer audience familiar working with node & installing dependencies. This extension is most helpful for a different audience.

I think for developers whose workflow includes project specific VS Code plugins, this is a nice solve, and can be found searching `agoric` on the VC code extension marketplace - https://marketplace.visualstudio.com/search?term=agoric&target=VSCode&category=All%20categories&sortBy=Relevance

Issue #776, opened 2023-02-21
`yarn docs:build` fails with `Error: error:0308010C:digital envelope routines::unsupported`


ivanlei:
**Summary**
Local builds and CI checks are failing `yarn docs:build` with `Error: error:0308010C:digital envelope routines::unsupported`

**Description**
See https://github.com/Agoric/documentation/pull/774 for an example of [failing checks](https://github.com/Agoric/documentation/actions/runs/4211520015/jobs/7309922099). The same errors reproduce in my local build.
```
➜  documentation git:(temp-new-content) yarn docs:build
yarn run v1.22.5
$ vuepress build main
wait Extracting site metadata...
tip Apply theme @vuepress/theme-default ...
tip Apply plugin container (i.e. "vuepress-plugin-container") ...
tip Apply plugin @vuepress/register-components (i.e. "@vuepress/plugin-register-components") ...
tip Apply plugin @vuepress/active-header-links (i.e. "@vuepress/plugin-active-header-links") ...
tip Apply plugin @vuepress/search (i.e. "@vuepress/plugin-search") ...
tip Apply plugin @vuepress/nprogress (i.e. "@vuepress/plugin-nprogress") ...
tip Apply plugin check-md (i.e. "vuepress-plugin-check-md") ...
tip Apply plugin @vuepress/last-updated (i.e. "@vuepress/plugin-last-updated") ...

● Client █████████████████████████ building (10%) 1/2 modules 1 active
 cache-loader › babel-loader › node_modules/@vuepress/core/lib/client/clientEntry.js

● Server █████████████████████████ building (10%) 1/2 modules 1 active
 cache-loader › babel-loader › node_modules/@vuepress/core/lib/client/serverEntry.js

Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating
/Users/ivanleichtling/source/documentation/node_modules/loader-runner/lib/LoaderRunner.js:114
			throw e;
			^

Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
    at module.exports (/Users/ivanleichtling/source/documentation/node_modules/webpack/lib/util/createHash.js:135:53)
    at NormalModule._initBuildHash (/Users/ivanleichtling/source/documentation/node_modules/webpack/lib/NormalModule.js:417:16)
[...]

PR #780, opened 2023-02-21
fix: require lts 16 due to openssl changes and webpack4 dep via vuepress

arirubinstein:
Fixes the node pin on 16, in both yarn and .node-version. Vuepress depends on webpack4, which is incompatible with stock node 18 without arg modifications. Also performed a driveby update of deprecated github actions versions.

closes: https://github.com/Agoric/documentation/issues/776


ivanlei (approved):
Verified manually -- https://gist.github.com/ivanlei/147561b7051aef2c995429644506ca14

I did need to:
```shell
$ nvm install v16.19.1
$ npm install -g yarn
```

PR #782, opened 2023-02-24
plural of IST is awkward

dckc:
> minter of ISTs

That doesn't seem like conventional usage.

### DIFF
--- a/main/guides/zoe/actual-contracts/PSM.md
+++ b/main/guides/zoe/actual-contracts/PSM.md
@@ -12,7 +12,7 @@ token for the interchain ecosystem on the Agoric chain. IST is designed to maint
 with the US dollar (USD) for broad accessibility and provides Agoric and the interchain
 ecosystem users an asset with minimum price volatility. In addition, IST is the native 
 fee token for the Agoric platform and provides some of the core functionality and stability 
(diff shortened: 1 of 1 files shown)


Tyrosine22 (changes requested):
To make acronyms plural, you add an "s" to the end. Since "IST" = "Inter Stable Token", if we want to talk about multiple tokens (or Inter Stable Tokens in general), the plural is ISTs.

I did notice that earlier in the paragraph, I talk about singular IST, which is incorrect. (e.g., The previous sentence should be "In addition,, ISTs are the native etc.") Please close this PR, and I'
ll submit a PR to you with the instances of "IST" made plural.

dckc:
I have never seen anyone write "I paid 10 USDs." Always "I paid 10 USD." Likewise IST.

In https://en.wikipedia.org/wiki/United_States_dollar we see "100 USD" not "100 USDs".

IST is effectively a unit of measure. 10sec , not 10secs. 40m, not 40ms, which would change the meaning completely from "40 meteres" to "40 microseconds"

Tyrosine22 (approved):
Well I'll be.... I didn't know about the different rules for pluralizing monetary symbols. 

You're correct again, sir.

PR #783, opened 2023-02-24
reword multiple PSMs re software component

dckc:
Phrase it in the ever-present tense to document the software component independent of how it may be deployed in a particular blockchain.


Tyrosine22 (approved):
Looks good.

PR #785, opened 2023-02-25
UserSeat: avoid "trusted parties"

dckc:
Better would be to give examples of when sharing a UserSeat is mutually beneficial vs. when it leads to unwanted vulnerability.


Tyrosine22 (approved):
Looks good.

ivanlei (approved):
🚀

PR #789, opened 2023-03-28
fix: removed unnecessary character

tgrecojs:
While referencing the documentation I noticed an unnecessary character, "a", preceding "ZCFSeat". I did not seek out any other instances of this in the documentation - this PR only addresses what I came across a moment ago. 

It's not extensive, but I felt it was worth opening a PR to resolve this quickly.

### Link to live issue
[ZCFSeat - docs.agoric.com](https://docs.agoric.com/reference/zoe-api/zcfseat.html)

### Screengrab of issue
https://user-images.githubusercontent.com/6646552/228098212-76c40f81-f9f4-47df-ba2f-3ce6c3fa8904.mp4

### DIFF
--- a/main/reference/zoe-api/zcfmint.md
+++ b/main/reference/zoe-api/zcfmint.md
@@ -6,13 +6,13 @@ methods.
 
 **ZCFMints** are created and returned by **Zoe Contract Facet's** **[zcf.makeZCFMint()](./zoe-contract-facet.md#zcf-makezcfmint-keyword-assetkind-displayinfo)** method.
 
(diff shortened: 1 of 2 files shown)


ivanlei (approved):
🚀 Thank you

Tyrosine22 (changes requested):
All those characters are intentional; we settled on a naming convention where all the object names in method declarations have an indefinite article in front, in order to make it clear that they're not static methods.

Tyrosine22:
Closing because the fixed "error" is there by design.

PR #794, opened 2023-04-27
fix: bump node version

### LINKED ISSUE #793: Support Node18
Pages now supports 18, so this ticket is to track the pin upgrade

arirubinstein:
fixes #793


package.json
```diff
@@ -4,15 +4,15 @@
   "description": "Documentation of Agoric products",
   "type": "module",
   "engines": {
-    "node": "^16"
+    "node": "^18"
```
mhofman, line 7: This must not be more specific than `agoric-sdk`:
https://github.com/Agoric/agoric-sdk/blob/4ad26d25fa13936a1f9c2de38baee9e05d89932b/package.json#L36

PR #797, opened 2023-05-04
Documentation for the CoreEval feature

Tyrosine22:
Here's the initial pass at CoreEval documentation. I want to add additional detail about the permissions soon-ish, but this PR should be good enough to get us started.

Also... git is claiming that I completely rewrote the first topic in Getting Started, but as you can see.... I didn't. I'm not sure why git is making that claim; pls ignore that "change", I suppose.


Tyrosine22:
Shutting down this PR in favor of one that only includes the relevant files.

PR #809, opened 2023-06-07
fix: counter example heading/description

dckc:
The counter example is not a smart contract. A smart contract in our framework is a module that exports a `start` or `prepare` function.

And "x is secure" is a claim that we typically don't make. Security is rarely an absolute property but more often a matter of degree.


ivanlei (approved):
🚀

PR #799, opened 2023-05-08
Documentation for the CoreEval feature

Tyrosine22:
Here's the initial pass at CoreEval documentation. 

BTW, some of the changes which originally appeared in this PR have been moved to the revisedgettingstarted PR (https://github.com/Agoric/documentation/pull/811).

The gist with the output from the initial set of console commands is located here: https://gist.github.com/Tyrosine22/6da2367ae8897736953c2f1286164eed

The gist with the output from the final set of console commands is located here:
https://gist.github.com/Tyrosine22/12ba3222f15926f25eee30ce86be33f2

PR #811, opened 2023-06-12
Revisedgettingstarted

Tyrosine22:
Contains new setup info required by CoreEval, as well as a version update required by everything.

### DIFF
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -48,7 +48,7 @@ The Agoric SDK is supported on <a href="https://en.wikipedia.org/wiki/Linux">Lin
 - To launch a terminal on Linux, use the **Terminal** application.
 - To access WSL from Windows, visit the [WSL documentation](https://docs.microsoft.com/en-us/windows/wsl/).
 
-## Install Go Version 1.20.3 or Higher
+## Install Go
 
 Download Go from [go.dev/doc/install](https://go.dev/doc/install) and follow the instructions for your platform.
 
@@ -65,7 +65,7 @@ jq --version
 ```
 
 
-## Install Node.js LTS Version
+## Install Node.js
 
 Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform. We recommend installing the LTS version of node 18.
 
@@ -122,8 +122,7 @@ yarn build
 ## Build the Cosmic Swingset Package
 
 ```shell
-cd packages/cosmic-swingset
-make
+(cd packages/cosmic-swingset && make)
 ```
 
 ## Install Agoric CLI
@@ -131,14 +130,12 @@ make
 Install the Agoric CLI (Command Line Interface) in a convenient place in your `$PATH` such as:
 
 ```shell
-cd ../..
 yarn link-cli ~/bin/agoric
 ```
 
 or:
 
 ```shell
-cd ../..
 sudo yarn link-cli /usr/local/bin/agoric
 ```
 
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -11,7 +11,7 @@ If you're familar with JavaScript development tools such as `node`, `yarn`, and 
 ```shell
 go version
 jq --version
-node --version # 18.16.0 or higher
+node --version # LTS version such as 18.16.0
 npm install --global yarn # Install yarn for package management
 git clone --branch community-dev https://github.com/Agoric/agoric-sdk # Clone the "community-dev" branch
 cd agoric-sdk
@@ -32,7 +32,7 @@ If `agoric` is not found, then you may need to add the parent directory of where
 This presentation is a good overview of the Agoric SDK setup process,
 though a few details are out of date:
 
-- node version: 12.x is too old; use 18.16.0 or higher
+- node version: 12.x is too old; use the LTS version 18.16.0 or a later LTS version
 - skip `git checkout hackathon-2020-11`; use the `community-dev` branch
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/w0By22jYhJA" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
@@ -65,13 +65,15 @@ jq --version
 ```
 
 
-## Install Node.js 18.16.0 or Higher
+## Install Node.js LTS Version
 
 Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform. We recommend installing the LTS version of node 18.
 
 ```shell
-node --version # 18.16.0 or higher
+node --version # LTS version such as 18.16.0
 ```
+
+**Note:** Agoric will support all even numbered versions of Node.js. 
 
 ## Install the Yarn Package Manager
 
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -10,7 +10,6 @@ If you're familar with JavaScript development tools such as `node`, `yarn`, and 
 
 ```shell
 go version # Version 1.20.3 or higher
-jq --version
 node --version # LTS version such as 18.16.0
 npm install --global yarn # Install yarn for package management
 git clone --branch community-dev https://github.com/Agoric/agoric-sdk # Clone the "community-dev" branch
@@ -55,15 +54,6 @@ Download Go from [go.dev/doc/install](https://go.dev/doc/install) and follow the
 ```shell
 go version # Version 1.20.3 or higher
 ```
-
-## Install jq
-
-Download jq from [jqlang.github.io/jq/](https://jqlang.github.io/jq/) and follow the instructions for your platform.
-
-```shell
-jq --version
-```
-
 
 ## Install Node.js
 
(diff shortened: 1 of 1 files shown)


main/guides/getting-started/README.md
```diff
@@ -9,12 +9,17 @@ After installing the Agoric SDK, you can proceed to [starting a project](./start
 If you're familar with JavaScript development tools such as `node`, `yarn`, and `git`:
 
 ```shell
-node --version # 16.19.1 or higher
+go version
+jq --version
+node --version # 18.16.0 or higher
```
dckc, line 14: **important**:

```suggestion
node --version # LTS version such as 18.16.0
```

"or higher" suggests we will support version 19, 21, etc. but only even numbered node major version are stable. I don't think we'll support version 19, 21, etc.
I think our policy is to support LTS (long term support) releases... though I'm not sure if we've gotten as far as making that a documented policy.

main/guides/getting-started/README.md
```diff
@@ -43,12 +48,29 @@ The Agoric SDK is supported on <a href="https://en.wikipedia.org/wiki/Linux">Lin
 - To launch a terminal on Linux, use the **Terminal** application.
 - To access WSL from Windows, visit the [WSL documentation](https://docs.microsoft.com/en-us/windows/wsl/).
 
-## Install Node.js 16.19.1 or Higher
+## Install Go
```
dckc, line 51:
```suggestion
## Install Go version 1.20.3 or higher
```

ref:

https://github.com/Agoric/agoric-sdk/blob/b8d6697074b899630909440f0ea3caa519cdad44/repoconfig.sh#LL4C1-L4C22

main/guides/getting-started/README.md
```diff
@@ -43,12 +48,29 @@ The Agoric SDK is supported on <a href="https://en.wikipedia.org/wiki/Linux">Lin
...
+## Install jq
+
+Download jq from [jqlang.github.io/jq/](https://jqlang.github.io/jq/) and follow the instructions for your platform.
+
+```shell
+jq --version
+```
+
+
+## Install Node.js 18.16.0 or Higher
```
dckc, line 68:
```suggestion
## Install Node.js LTS version such as 18.16.0
```

main/guides/getting-started/README.md
```diff
@@ -43,12 +48,29 @@ The Agoric SDK is supported on <a href="https://en.wikipedia.org/wiki/Linux">Lin
...
+
+Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform. We recommend installing the LTS version of node 18.
+
+```shell
+node --version # 18.16.0 or higher
```
dckc, line 73:
```suggestion
node --version # LTS version such as 18.16.0
```

main/guides/getting-started/README.md
```diff
@@ -85,7 +107,7 @@ Cloning and installing the Agoric SDK can take a while. Please be patient.
 yarn install
 ```
 
-**Note:** If you run into errors during install or build, make sure you have build-essential installed. `gcc --version`.
+**Note:** If you run into errors during install or build, make sure you have build-essential components installed. `gcc --version`.
```
dckc, line 110: **important**:

"build-essential" here is a noun, not an adjective. It's the name of a debian / Ubuntu package. So the note should perhaps say:

```suggestion
**Note:** If you run into errors during install or build, make sure you have the relevant devloper tools installed. For example, on debian or Ubuntu linux: `sudo apt-get install build-essential`.
```

By way of example, DigitalOcean, which has really good docs, discusses it this way:

[How To Compile and Install Packages From Source Using Make on a VPS \| DigitalOcean](https://www.digitalocean.com/community/tutorials/how-to-compile-and-install-packages-from-source-using-make-on-a-vps)

> To compile sources on Linux, you will need the package called "build-essential" on Debian-based systems and "Development Tools" on CentOS, as it contains the gcc/g++ compilers and libraries required to compile packages. To install this on Debian and Ubuntu run:
> 
> ```
> apt-get install build-essential
> ```
> And on CentOS run:
>
> ```
> yum groupinstall "Development Tools"
> ```

but they don't support Mac nor WSL.

main/guides/getting-started/README.md
```diff
@@ -95,17 +117,26 @@ yarn build
 
 **Note:** MacOS, Linux, and WSL are currently supported. There is currently **no support** for native Windows 10 or native Windows 11.
 
+## Build the Cosmic Swingset Package
+
+```shell
+cd packages/cosmic-swingset
+make
```
dckc, line 124: suggestion:

```suggestion
(cd packages/cosmic-swingset && make)
```

Doing the `cd` (and `make`) in a sub-shell means that the `cd ../..` below aren't necessary.

dckc (changes requested):
a couple **important** comments; the others are suggestions

main/guides/getting-started/README.md
```diff
@@ -43,12 +48,29 @@ The Agoric SDK is supported on <a href="https://en.wikipedia.org/wiki/Linux">Lin
...
 
-Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform. We recommend installing the LTS version of node 16.
+Download Go from [go.dev/doc/install](https://go.dev/doc/install) and follow the instructions for your platform.
 
 ```shell
-node --version # 16.19.1 or higher
+go version
+```
+
+## Install jq
```
ivanlei, line 59: `jq` should not be a required piece of getting started. what requires it?

main/guides/getting-started/README.md
ivanlei: _comment_

We need to centralize documentation on required versions in a single spot so there's only one spot to update. Adding versions into titles, headers, comments, multiple docs, etc. leads to chronically out of date documentation.

ivanlei (changes requested):

main/guides/getting-started/README.md
dckc: Where do you suggest? I can't think of a better place. For this audience, anyway. For the validator audience, we seem to put the required version info in each release. For the dapp / smart contract side of the SDK, we don't seem to do releases.

Perhaps putting it in the heading is unnecessary.

main/guides/getting-started/README.md
dckc: I agree that it _should_ not be required, but the proposed core eval docs (#799) use `make scenario2-setup` in `packages/cosmic-swingset` and that requires `jq`. As noted there, I'd like us to use `agoric start ...` instead of `make` but then we would have to more or less copy many of the commands out of the Makefile into the documentation. `scenario2-setup` does at least run in ci.

main/guides/getting-started/README.md
Tyrosine22: I've updated the various :Node version 18.16.0 or later" strings to reflect this LTS versioning instead.

main/guides/getting-started/README.md
Tyrosine22: Yeah, this page seems like a logical place to put all the versioning info.

I've updated the Go versioning info within the code blocks, and removed all versioning info from any headers.

main/guides/getting-started/README.md
Tyrosine22: This was changed to simply "Install Node.js". (The body of the section explains LTS versioning for Node.js)

main/guides/getting-started/README.md
Tyrosine22: Changed.

main/guides/getting-started/README.md
Tyrosine22: Pretty slick. Change made.

main/guides/getting-started/README.md
Tyrosine22: Since the CoreEval docs now use the various agoric XXX commands rather than make scenario2-setup, I deleted all references to installing jq, since it's no longer required.

main/guides/getting-started/README.md
dckc: fwiw, https://github.com/Agoric/agoric-sdk/blob/master/repoconfig.sh is a candidate single-source-of-truth. But importing from there into the docs repo seems kinda tricky. It might be worth having a unit test to check that they're in sync.

main/guides/getting-started/README.md
```diff
@@ -43,14 +47,24 @@ The Agoric SDK is supported on <a href="https://en.wikipedia.org/wiki/Linux">Lin
...
 
+## Install Node.js
+
+Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform. We recommend installing the LTS version of node 18.
+
+```shell
+node --version # LTS version such as 18.16.0
+```
+
+**Note:** Agoric will support all even numbered versions of Node.js.
```
dckc, line 66: **important**:

```suggestion
**Note:** Agoric will support all long-term support (LTS) versions of Node.js. 
```

not all even numbered versions are always LTS versions. first node 18.1.0 comes out, then folks notice problems, then 18.2.0... by around 18.5.0, problem reports slow down and they put the LTS stamp on it.

dckc (commented):
getting there. 2 remaining **important** comments

main/guides/getting-started/README.md
Tyrosine22: Ah. So odd numbered versions don't get the LTS stamp and only certain iterations of even numbered versions get the LTS stamp? Gotcha.

Tyrosine22: Oh... and the text was changed to match your suggestion.

main/guides/getting-started/README.md
Tyrosine22: The text has been changed to match your suggestion.

ivanlei (approved):

PR #812, opened 2023-06-13
fix(solo): remove stale `ag-cosmos-helper` and `local.agoric.app` refs

michaelfig:
These references are stale, they are now `agd` and https://wallet.agoric.app/locator/


dckc (commented):
The `agd` part looks right. I'm less familiar with the `/locator/` stuff. @samsiegart ?

ivanlei (approved):

PR #825, opened 2023-07-06
correct zoe.getConfiguration()

Chris-Hibbert:
This returns info that isn't specific to any zoe contract. It describes the feeIssuer, which is the issuer of **IST**.


ivanlei (approved):
🚀

Issue #824, opened 2023-07-06
UserSeat description should describe timing of promise resolution


Chris-Hibbert:
See  https://github.com/Agoric/agoric-sdk/issues/8016

In my response there, I wrote

> We will add something to the [userSeat documentation](https://docs.agoric.com/reference/zoe-api/user-seat.html#e-userseat-getofferresult) for getPayouts() and getPayout() explaining the timing of these promises. It's not obvious from those documents that .getOfferResult() is intended to resolve quickly, .getProposal() and .hasExited() should resolve immediately, while .getPayout() and .getPayouts() like .numWantsSatisfied() and .getFinalAllocation() don't resolve until the seat exits.

PR #826, opened 2023-07-07
Updating the zoe-api per Issue #824

Tyrosine22:


Chris-Hibbert (changes requested):
As I described in #824, there are three distinct cases. 
 1 .`.getOfferResult()` is expected to resolve quickly, but is at the discretion of the contract
 2 `.getProposal()` and `.hasExited()` resolve immediately, 
 3 `.getPayout()`, `.getPayouts()`, `.numWantsSatisfied()` and `.getFinalAllocation()` resolve when the seat exits.

Let's see what's left.  `.getExitSubscriber()` is in case 2.   `.tryExit()` requires some work from Zoe, but it's effectively case 2 as well.

We should explain the difference and annotate all these methods.

Chris-Hibbert:
closed via #834

PR #829, opened 2023-07-13
dapp not loading note

erondpowell:
A tiny little thing, but the initial "link dapp to wallet" modal message would not load for me. I tried reinstalling Agoric SDK as well as the dapp, restarted the terminals, re-opened the localhost link.... but in the end, it needed a simple page refresh.

It looks like multiple people bumped into this issue, so this PR adds a gentle tip for newcomers.


erondpowell:
**To The Reviewer:** 
I got a minor question.
Why is this in the package.json scripts? 
`--openssl-legacy-provider`

This makes all the package.json commands fail for me.
I simply removed it to run `docs:dev`
(I used node 16.14, not 16.13, but that should be OK I think)

dckc (approved):
this looks helpful. Thanks!

dckc:
> **To The Reviewer:** I got a minor question. Why is this in the package.json scripts? `--openssl-legacy-provider`

hm. I don't know. Maybe something about building in cloudflare? @arirubinstein , any ideas?

> This makes all the package.json scripts fail for me.

Do us a favor? Open a new issue, detailing the steps to reproduce and the symptoms?

Issue #798, opened 2023-05-04
prepare for migration to zcf atomicTransfer()


Chris-Hibbert:
@erights came up with [a new approach to managing reallocation in contracts (6577)](https://github.com/Agoric/agoric-sdk/pull/6577) that we have been very happy with. We have been planning to [migrate the temporary helper into ZCF (6679)](https://github.com/Agoric/agoric-sdk/pull/6679) and remove the old approach, but we neglected to warn the developers that this major change was coming.

We should, ASAP, prepare documentation on use of the `atomicTransfer` helper, and explain that it will be moved into ZCF shortly, but more importantly, we should add *deprecation* annotations to the existing `incrementBy` and `decrementBy` ZCF functions so our developers know what to expect.

Luckily, all our contracts are now written in the interim style, so anyone reading our code will see examples of the new approach.

Chris-Hibbert:
`zcfSeat.getStagedAllocation()` and `zcfSeat.hasStagedAllocation()` should also be deprecated. They'll go away when we remove the old approach, and aren't needed while using the interim approach.

PR #833, opened 2023-07-24
doc: final update for atomicRearrange once zcf version is released

Chris-Hibbert:
refs: #798

Add documentation for the version of `atomicRearrange()` internal to Zoe. Deprecate the helper (though there's no need to actually get rid of it.)

This change should go out shortly after [the upgrade to mainnet](https://github.com/Agoric/agoric-sdk/issues/7982).


turadg:
Dropped to draft since ZCF on Mainnet doesn't yet have `atomicRearrange`.

PR #834, opened 2023-07-26
Add timing details for userSeat promise returns

Chris-Hibbert:
see: https://github.com/Agoric/agoric-sdk/issues/8016
closes: #824

All the methods on userSeat are described as returning Promises, but some resolve immediately, while others can wait varying periods. This adds some clarity on what to expect.

### DIFF
--- a/main/reference/zoe-api/user-seat.md
+++ b/main/reference/zoe-api/user-seat.md
@@ -14,6 +14,13 @@ and, if it's allowed for this seat, call **tryExit()**.
 
 Take care when sharing a **UserSeat**, since it includes authority to (attempt to) exit the seat.
 
+These methods are all documented as returning Promises, since they are always called remotely,
+but that obscures the fact that they are expected to resolve at semantically different times.
+Some return a value promptly, and others return a promise that won't be fulfilled until the
+seat is exited. **getOfferResult()** resolves at the discretion of the contract. The
+descriptions below will call out explicitly the cases in which the value is not guaranteed to
+be resolved promptly.
+
 **UserSeat** includes queries for the associated offer's current state
 and an operation to request that the offer exit, as follows:
 
@@ -54,6 +61,8 @@ const paymentKeywordRecord = {
 };
 ```
 
+This promise will be resolved when the seat exits.
+
 ## E(UserSeat).getPayout(keyword)
 - **keyword**: **[Keyword](./zoe-data-types.md#keyword)**
 - Returns: **Promise&lt;[Payment](/reference/ertp-api/payment.md)>**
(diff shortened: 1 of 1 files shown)


main/reference/zoe-api/user-seat.md
```diff
@@ -14,6 +14,13 @@ and, if it's allowed for this seat, call **tryExit()**.
 
 Take care when sharing a **UserSeat**, since it includes authority to (attempt to) exit the seat.
 
+These methods are all documented as returning Promises, since they are always called remotely,
+but that obscures the fact that they are expected to resolve at semantically different times.
+Some return a value immediately, some do a small amount of computation before returning, and
```
erights, line 19:
```suggestion
Some return a value promptly, some do a small amount of computation before returning, and
```

You use "promptly" below, and for this purpose it is much clearer than immediately. Do we define either term anywhere? I like Norm's definition of "prompt" or "promptly", that it will happen without any further actions needed by any other party. Also, I would say, implies only computational delays such as round trips, rather than scheduled timing delays.

Thus, unclear if you meant "do a small amount of computation before returning" to be a separate category. From the pov of a remote caller, I'd expect these to also be prompt and to not be in a different category from your "immediately".

I would like to reserve terms like "immediate" and "immediately" for things that happen synchronously in the same turn, which is never true for remote calls.

main/reference/zoe-api/user-seat.md
```diff
@@ -62,21 +71,35 @@ A **Payout** is a **Payment** that goes to a party in a successful transaction,
...
 
+Since the contract can return whatever it wants as an offerResult, there is no guarantee that the
+promise will resolve promptly.
+
 ## E(UserSeat).hasExited()
   - Returns: **Promise&lt;Boolean>**
 
-Returns **true** if the seat has exited, **false** if it's still active.
+Returns **true** if the seat has exited, **false** if it's still active. The value is returned
+immediately.
```
erights, line 93:
```suggestion
promptly.
```

main/reference/zoe-api/user-seat.md
```diff
@@ -62,21 +71,35 @@ A **Payout** is a **Payment** that goes to a party in a successful transaction,
...
+
+If you want to take some action when the seat does exit, use **getExitSubscriber()**
+and use
+```
+const subscriber = E(seat).getExitSubscriber();
+E.when(E(subscriber).getUpdateSince(), () => takeAction());
+```
```
erights, line 100: The difference being that subscribing to a remote contract remains valid across upgrade of the contract, whereas a remote promise to be resolved by a contract will instead be broken by an intervening contract upgrade.

main/reference/zoe-api/user-seat.md
```diff
@@ -62,21 +71,35 @@ A **Payout** is a **Payment** that goes to a party in a successful transaction,
...
+
+await
```
erights, line 102: Dangling `await`?

main/reference/zoe-api/user-seat.md
```diff
@@ -99,6 +121,7 @@ Returns a **Promise** for a number which indicates the result of the exited **Pr
 | 0 | The user didn't get what they wanted from the **Proposal**, so their offer was refunded. |
 | 1 | The user got what they wanted from the **Proposal**, so their offer is spent & gone. |
 
+This promise will be resolved when the seat exits.
```
erights, line 124:
```suggestion
This promise will be resolved promptly once the seat exits.
```

Is technically more precise, but the original text is fine if this precision is unneeded.

main/reference/zoe-api/user-seat.md
```diff
@@ -110,5 +133,4 @@ Returns a **Promise** for the **Subscriber** for the seat.
 - Returns: **Promise&lt;[Allocation](./zoe-data-types.md#allocation)>**
 
 Returns a **Promise** for the **Allocation** when the **UserSeat** exits the **proposal**.
-
-
+This promise will be resolved when the seat exits.
```
erights, line 136: Likewise

erights (approved):
Good clarifications!

Issue #813, opened 2023-06-16
links to parts of pages get their fragments stripped off


dckc:
For example, following this link:

https://docs.agoric.com/reference/zoe-api/zoe.html#e-zoe-installbundleid-bundleid

ends up at

https://docs.agoric.com/reference/zoe-api/zoe.html

and following

https://docs.agoric.com/guides/getting-started/deploying.html#contract-deployment

ends up at

https://docs.agoric.com/guides/getting-started/deploying.html

cc @Tyrosine22 @Chris-Hibbert

PR #840, opened 2023-08-10
Support directly jumping to document section URLs

michaelfig:
closes: #813

Before, if somebody clicked on a URL like http://docs.agoric.com/guides/js-programming/hardened-js.html#object-capabilities-ocaps it would strip off the `#` fragment, and just render the top of the page.

This PR fixes that stripping by patching the VuePress router, so now an unnecessary redirect is avoided.

Moreover, the target of the fragment doesn't exist until VuePress renders the page, so keep trying every 500ms until we find the app's main tag has rendered.  It would be nice to find a non-polling way to do this, but it works for the old `logoUrlChanger`, so I just generalized that approach.


dckc (approved):
woot!

fixes: #813

at least it works in one case:
https://mfig-jump-to-fragment.documentation-7tp.pages.dev/reference/zoe-api/zoe.html#e-zoe-getpublicfacet-instance

PR #841, opened 2023-08-11
feat: Explain, DRY out, and isolate page fixups

gibson042:
Ref #840

### DIFF
--- a/main/.vuepress/config.js
+++ b/main/.vuepress/config.js
@@ -36,39 +36,39 @@ module.exports = {
       'script',
       {},
       `
-    const withSelector = (selector, callback) => {
-      const el = document.querySelector(selector);
-      if (el) {
-        callback(el);
+    /** @type {Map<[...anySelectors: string[]], (elems: (Element | null)[]) => unknown>} */
+    const fixups = new Map();
+
+    // Update the "home" link to target agoric.com while intercepting clicks
+    // such that those outside of its image continue routing to the root of
+    // the documentation site.
+    fixups.set(['.home-link', '.logo'], ([homeEl, logoEl]) => {
+      if (homeEl) {
+        homeEl.setAttribute('href', 'https://agoric.com');
+        homeEl.setAttribute('onclick', 'return false;');
       }
-      return el;
-    };
-    const fixups = new Map([
-      // Update the "home" link to target agoric.com while intercepting clicks
-      // such that those outside of its image continue routing to the root of
-      // the documentation site.
-      ['homeLink', () => withSelector(".home-link", homeEl => {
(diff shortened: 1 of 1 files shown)


dckc (commented):
looks plausible, but I don't think I grok this code deeply enough to approve without tests. I suppose the live preview is a test, but I'm not sure what it's testing.

main/.vuepress/config.js
```diff
@@ -36,42 +36,34 @@ module.exports = {
       'script',
       {},
       `
+    const withSelector = (selector, callback) => {
+      const el = document.querySelector(selector);
+      if (el) {
+        callback(el);
```
michaelfig, line 42: If you're not using the return value of the callback, and we still want to be robust against callback failure, can I suggest something like:
```suggestion
        Promise.resolve(el).then(callback);
```

michaelfig (approved):
LGTM!  Nice cleanup.

Only one nit that I found that could cause multiple reexecution of a callback if it keeps failing.  Would you be able to execute them in a new turn?

michaelfig:
Manually tested, and I saw that the fixups were being properly applied.

main/.vuepress/config.js
gibson042: Yeah, I noticed that as well. Updated to adopt that pattern in the `setInterval` callback and ended up removing `withSelector` altogether, and also restored handling of the home-link logo and text in a single callback to match pre-#840 behavior. The changes are substantial enough to warrant re-review; PTAL.

main/.vuepress/config.js
```diff
@@ -36,47 +36,39 @@ module.exports = {
...
+    // the documentation site.
+    fixups.set(['.home-link', '.logo'], ([homeEl, logoEl]) => {
+      if (homeEl) {
+        homeEl.setAttribute('href', 'https://agoric.com');
+        homeEl.setAttribute('onclick', 'return false;');
+      }
+      if (logoEl) {
+        logoEl.setAttribute('onclick', "document.location='https://agoric.com';return false;");
+      }
+    });
```
michaelfig, line 53: I was pretty sure this coupling of the `.home-link` fixup with the `.logo` fixup was brittle, not a feature.  Do you know anything that says otherwise?

michaelfig (approved):
LGTM.  My comment is just a question, and your implementation is still definitely an improvement on the pre-#840 one.

main/.vuepress/config.js
gibson042: I think it's just robustness against future structural changes in Vuepress... if one selector matches but the other does not, we don't want the interval to continue firing and possibly matching the wrong content later on.

PR #845, opened 2023-08-23
rationale for upper-case constraint on Zoe keywords

dckc:


erights (commented):
Also mention that it must not be the identifiers `NaN` or `Infinity` ?

erights (approved):
LGTM

PR #846, opened 2023-08-24
Update "deployed contracts" for vaults release

dckc:
I came up with this list by starting with [test-boot-config.js](https://github.com/Agoric/agoric-sdk/blob/mainnet1B/packages/boot/test/test-boot-config.js), which leads to [decentral-main-vaults-config.json](https://github.com/Agoric/agoric-sdk/blob/mainnet1B/packages/vats/decentral-main-vaults-config.json), where we find a few such as `mintHolder` plus proposals to install more; looking thru the proposals is a little tedious.

Ideally `test-boot-config.js` would have a static enumeration of all these contracts.


ivanlei (approved):
LGTM

PR #848, opened 2023-09-05
test(snippets): permit Date "secure mode" exceptions

michaelfig:
Newer SES `Date.now()` and `new Date()` throw exceptions starting with `secure mode`.

This updates the snippets tests to tolerate both the new behaviour, and for compatibility, the old behaviour of `Date.now()` returning `NaN`.


dckc (commented):
This test doesn't seem to intersect with any snippets. But ah... there is corresponding text:

> With the default Compartment options, the non-deterministic Math.random is not available and Date.now() always returns NaN.
-- guides/js-programming/hardened-js.html

Seems like that should get fixed in this PR too.

kriskowal (approved):

PR #855, opened 2023-09-20
chore(getting-started): Improve `yarn link-cli` documentation

gibson042:
cc @sufyaankhan


main/guides/getting-started/README.md
```diff
@@ -131,7 +131,7 @@ or:
 sudo yarn link-cli /usr/local/bin/agoric
 ```
 
-**Note:** Run `echo $PATH` to see your current `$PATH`. Here you'll see which directory you should use for `yarn link-cli`.
+**Note:** Run `echo $PATH | tr : '\n'` to see directories in your current `$PATH`, which are good candidates for where to have `yarn link-cli` place the executable.
```
ivanlei, line 134: _comment_:

IMO this is a level of complexity unnecessary for most users. It introduces using `tr` to split `PATH` which might confuse with the same magnitude it improves things.

gibson042: Updated to simplify.

dckc (approved):
looks helpful

PR #865, opened 2023-10-01
refer to ERTP as a protocol rather than standard

dckc:
"ERTP standard" seems a little redundant.


erights (approved):
I agree. LGTM, thanks.

PR #871, opened 2023-10-31
Update README.md

kbennett2000:
As per Dan C (regarding Sufyaan's issues), added the following commands after `agoric install`: `cd contract`
`yarn test`

### DIFF
--- a/main/guides/dapps/README.md
+++ b/main/guides/dapps/README.md
@@ -24,6 +24,10 @@ If you have installed the [Agoric CLI](../getting-started/) and you want to try 
 	cd my-fungible-faucet
 	# Install the project dependencies
 	agoric install
+	# Move to the contract directory
+ 	cd contract
+ 	# Run the yarn test command
+ 	yarn test
 	# Start the Agoric VM
 	agoric start --reset
 	```


main/guides/dapps/README.md
```diff
@@ -24,6 +24,10 @@ If you have installed the [Agoric CLI](../getting-started/) and you want to try
 	cd my-fungible-faucet
 	# Install the project dependencies
 	agoric install
+	# Move to the contract directory
+ 	cd contract
```
dckc, line 28: Does the `agoric start --reset` command below work with cwd = contract?

main/guides/dapps/README.md
```diff
@@ -24,6 +24,10 @@ If you have installed the [Agoric CLI](../getting-started/) and you want to try
...
+ 	# Run the yarn test command
+ 	yarn test
```
dckc, line 30: That's perhaps worth adding on its own, but it would be better to actually document how to use our test framework. That is: it would be better to add a short version of https://github.com/Agoric/agoric-sdk/wiki/agoric-sdk-unit-testing

dckc:
@kbennett2000 it's worth some effort to go beyond the default "Update README.md" commit message. We look at commit headlines _a lot_.

dckc (commented):
I'm not sure the `agoric start --reset` command works when the current directory is changed

Issue #627, opened 2022-02-02
document node version policy: LTS (and note v14 vs M1)


kennyrowe:
## Describe the bug
When building the SDK with a mac M1 the final build step fails

## To Reproduce
1. Using Node 14.15
2. Followed instructions for building with apple silicon (ie, install canvas)
3. `yarn  install` 
4. `yarn build`
5.  error
```
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@endo/init'
```

## Platform Environment
 - agoric-sdk version cc79ff6
 - Node version 14.15
 - Macbook pro M1 max chip

## Additional context
Tried adding package with `yarn add @endo/init` and got

```
yarn add v1.22.15
error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

tgrecojs:
> I'm not sure I follow. [Agoric/agoric-sdk#4460](https://github.com/Agoric/agoric-sdk/pull/4460) was because [Agoric/agoric-sdk#4420](https://github.com/Agoric/agoric-sdk/pull/4420), which was merged after this issue was opened, introduced a dev dependency that has a higher node version requirement, so we bumped up our CI image.

Ah, sorry for misinterpreting. My knowledge of ARM vx. x86 arch only begun when getting this M1 so apologies for being so green. 


> I'm not sure which arch or from where you installed node.js, but the official package for Node 14 does not support the M1 architecture, only the x86 emulation. That is nothing specific to this repo.

FWIW, I installed node via NVM. Running the `node -p process.arch` prints arm64. Below are some relevant contents i got from inspecting `process` within my terminals node repl.

```sh
  arch: 'arm64',
  platform: 'darwin',
  release: {
    name: 'node',
    lts: 'Fermium',
    sourceUrl: 'https://nodejs.org/download/release/v14.17.6/node-v14.17.6.tar.gz',
    headersUrl: 'https://nodejs.org/download/release/v14.17.6/node-v14.17.6-headers.tar.gz'
  },
 ```
 
Thanks for all the info here! @mhofman

Tartuffo:
Devs on M1 need to use V16.

dtribble:
[...]

PR #874, opened 2023-11-11
refine node LTS support wording

dckc:
It currently says "We recommend installing the LTS version of node 18" as if there are non-LTS versions of node 18 and as if we don't recommend v20 once it becomes LTS.

refs: #627

### DIFF
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -68,7 +68,7 @@ go version # Version 1.20.3 or higher
 ## Install Node.js
 
 Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform.
-We support Long Term Support (LTS) versions of node; currently, version 18.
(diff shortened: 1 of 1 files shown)


main/guides/getting-started/README.md
```diff
@@ -68,7 +68,7 @@ go version # Version 1.20.3 or higher
 ## Install Node.js
 
 Download Node.js from [Node.js](https://nodejs.org/) and follow the instructions for your platform.
-We recommend installing the LTS version of node 18.
+We support Long Term Support (LTS) versions of node; currently, version 18.
```
turadg, line 71: while you're here, "Node" should be capitalized. I think we should link to the Node docs on releases. And we can drop the version number from the copy here as it's in the shell comment right below. The "Current" release is likely to work but is not officially supported.

```suggestion
We support the Long Term Support (LTS) releases of Node, both [Active and Maintenance](https://nodejs.org/en/about/previous-releases).
```

kbennett2000 (approved):

turadg (approved):

dckc:
overtaken by rewrite of getting-started

PR #876, opened 2023-11-21
fix: Improve offer documentation

gibson042:
...particularly to document `offerArgs`.


main/glossary/README.md
```diff
@@ -424,9 +421,14 @@ can immediately cause the [seat](#seat) to exit, getting back the amount it offe
...
 
-A *passable* is something that can be marshalled (see the
-[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence))
-and sent to and from remote objects.
+A *passable* is something that can be sent to and from remote objects.
+Passables include pass-by-copy primitive values such as numbers and strings and
+pass-by-reference values such as Remotables and Promises.
+Passables also include hardened acyclic pass-by-copy containers that recursively terminate
+in non-container passables, such as CopyArrays like `harden(['foo', 'bar'])` and
+CopyRecords like `harden({ keys: [0, 1], values: ['foo', 'bar'] })`.
```
FUDCo, line 429: Might be worth a mention in passing of some things that aren't passable, just to underscore that it's a meaningful concept.

main/guides/zoe/proposal.md
```diff
@@ -4,69 +4,85 @@
...
+of 15 units of our imaginary Moola currency. (The "n" appended after each number indicates that
+it is represented as a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)
+rather than a [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number))
+
+::: warning Note
+It's important to understand that Amounts are just descriptions of assets with no
+intrinsic value. [Payments](/reference/ertp-api/payment.md) hold actual digital assets.
+:::
+
+**exit** specifies how the offer can be can cancelled. It must conform to one of three shapes:
```
Chris-Hibbert, line 53:
```suggestion
**exit** specifies how the offer can be cancelled. It must conform to one of three shapes:
```

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -63,9 +35,8 @@ You can get information about the contract instance via these methods:
 
 ## Invitation
 
-An **Invitation** is a kind of **[Payment](/reference/ertp-api/payment.md)**. It's a non-fungible eright
-that can be held in **Payments** or **[Purses](/reference/ertp-api/purse.md)**, just like any other
-eright. An **Invitation** **Payment** is a **Payment** holding an **Invitation**.
+An **Invitation** is a **[Payment](/reference/ertp-api/payment.md)** holding a non-fungible eright created by an **[InvitationIssuer](#invitationissuer)**.
```
Chris-Hibbert, line 38: There's only one InvitationIssuer per Zoe, so in the usual case it's an eright created by **the** invitationIssuer. Do you mean to hint that it's conceivable that there can be more than one invitationIssuer? I think that's not worth it.

main/glossary/README.md
```diff
@@ -424,9 +421,14 @@ can immediately cause the [seat](#seat) to exit, getting back the amount it offe
...
+See the
+[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence).
```
FUDCo, line 431: This appears to be a dangling link

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -83,11 +54,34 @@ is exclusively yours and no one else has access to it.
 
 ## Keyword
 
-An ASCII identifier string that must begin with an upper case letter
-in order to avoid collisions with JavaScript properties such as `toString`.
+A **Keyword** is a unique identifier string within a contract for tying together its
+**proposals**, **payments**, and **payouts**.
```
Chris-Hibbert, line 58:
```suggestion
A **Keyword** is a unique identifier string within a contract for tying together the 
**issuers** in its **proposals**, **payments**, and **payouts**.
```

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -83,11 +54,34 @@ is exclusively yours and no one else has access to it.
...
+It must be a valid ASCII [identifier](https://developer.mozilla.org/en-US/docs/Glossary/Identifier)
+and start with an upper case letter in order to avoid collisions with JavaScript properties
+such as `toString` when used as a property name in a record.
 (For more detail, see [Why do Zoe keywords have to start with a capital letter? #8241](https://github.com/Agoric/agoric-sdk/discussions/8241).)
 `NaN` and `Infinity` are also not allowed as keywords.
 
+<a name="amountkeywordrecord"></a>
```
Chris-Hibbert, line 65: Doesn't that HTML tag have to surround something to have an effect? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a says to use `id` instead.

stackoverflow thinks it's not okay for it to be empty.

FUDCo (approved):
A couple of minor nits, but this seems like a strict improvement.

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,63 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
-
-- **onDemand: null**: (Default) The offering party can cancel on demand.
-- **waived: null**: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
-- **afterDeadline**: The offer is automatically cancelled after a deadline, as determined 
-  by its **timer** and **deadline** properties.
-  **timer** must be a timer, and **deadline** must be a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) value interpreted with respect to the timer.
-  Some example timers use Unix epoch time, while others count block height.
+**proposal** must be either `undefined` or a record with **give**, **want**, and/or **exit** properties
+respectively expressing conditions regarding what assets will be given, what must be
+received in exchange, and an exit rule defining how/when the offer can be canceled.
```
Chris-Hibbert, line 303: ... what is being given, what must be received to satisfy offer safety, and an exit rule ...

I'd say "is being given" because zoe requires that the proposal match the payments, and the contract won't see the offer if they don't match. (and the invitation will be burned.

"what must be received to satisfy offer safety" because the user can want any arbitrary thing (subject to `want patterns` provided by the contract), and it's a valid offer. If the contract doesn't match the want, then the offer gets its give back.

Is there a reasonable place to describe `want patterns`? Do we talk about `makeInvitation` from the contract author's perspective?

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,63 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
 
- **offer()** returns a promise for a **UserSeat**.
- 
-```js
-const paymentKeywordRecord = {
-  Asset : quatloosPayment,
-  Price : moolaPayment
-};
-```
+**exit** specifies how the offer can be can cancelled. It must conform to one of three shapes:
```
Chris-Hibbert, line 317: It's worth mentioning that the contract can restrict the allowed exit clauses.

Chris-Hibbert (commented):
This is great! Lots of helpful changes here. I have a few suggestions, but mostly this is very good as it stands.

I'll bring up a local copy and verify links and visual formatting before approving.

main/reference/zoe-api/zcfmint.md
```diff
@@ -13,7 +13,7 @@ Returns an **IssuerRecord** containing the **[Issuer](/reference/ertp-api/issuer
 **[Brand](/reference/ertp-api/brand.md)** associated with the **zcfMint**.
 
 ## aZCFMint.mintGains(gains, zcfSeat?)
-  - **gains**: **[AmountKeywordRecord](./zoe-data-types.md#amountkeywordrecord)**
+  - **gains**: **[AmountKeywordRecord](./zoe-data-types.md#keywordrecord)**
```
dckc, line 16: Why this change?

I don't see any `#keywordrecord` anchor in https://docs.agoric.com/reference/zoe-api/zoe-data-types.html

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -83,11 +54,34 @@ is exclusively yours and no one else has access to it.
...
+## KeywordRecord
+
+A **KeywordRecord** is a [CopyRecord](/glossary/#passable) in which every property name
+is a **[Keyword](#keyword)**, such as `harden({ Currency: quatloosBrand, CurrencyUnit: 100n })`.
+Subtypes further constrain property values (for example, an
+**AmountKeywordRecord** is a **KeywordRecord** in which every value is an
+**[Amount](/reference/ertp-api/ertp-data-types.md#amount)** and a
+**PaymentKeywordRecord** is a **KeywordRecord** in which every value is a
+**[Payment](/reference/ertp-api/payment.md)**).
```
dckc, line 74: I know `PaymentPKeywordRecord` is in our API; that is: `Record<Keyword, Promise<Payment>>` that's returned from `getPayouts()`. I don't know of any use of `Record<Keyword, Payment>`.

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -83,11 +54,34 @@ is exclusively yours and no one else has access to it.
...
+and start with an upper case letter in order to avoid collisions with JavaScript properties
+such as `toString` when used as a property name in a record.
 (For more detail, see [Why do Zoe keywords have to start with a capital letter? #8241](https://github.com/Agoric/agoric-sdk/discussions/8241).)
 `NaN` and `Infinity` are also not allowed as keywords.
 
+<a name="amountkeywordrecord"></a>
+## KeywordRecord
+
+A **KeywordRecord** is a [CopyRecord](/glossary/#passable) in which every property name
+is a **[Keyword](#keyword)**, such as `harden({ Currency: quatloosBrand, CurrencyUnit: 100n })`.
```
dckc, line 69: That example seems weird. I don't believe anything in our `Record<Keyword, Brand | bigint>`. Did you mean to exhibit a `Record<Keyword, Amount<'nat'>>`? aka a `Record<Keyword, { brand: Brand<'nat'>, value: NatValue }>`

I wonder if it's useful to generalize to KeywordRecord at all. If so, we might as well lean into TypeScript notation: `Record<Keyword, T>`

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,63 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
-## E(Zoe).offer(invitation, proposal?, paymentKeywordRecord?, offerArgs)
-- **invitation**: **[Invitation](./zoe-data-types.md#invitation) | Promise&lt;Invitation>**
-- **proposal**: **Proposal** - Optional.
-- **paymentKeywordRecord**: **PaymentKeywordRecord** - Optional.
-- **offerArgs**: **Object**
+## E(Zoe).offer(invitation, proposal?, paymentKeywordRecord?, offerArgs?)
+- **invitation**: **[Invitation](./zoe-data-types.md#invitation) | Promise&lt;[Invitation](./zoe-data-types.md#invitation)>**
+- **proposal**: **[Proposal](/glossary/#proposal)** - Optional.
+- **paymentKeywordRecord**: **[PaymentKeywordRecord](./zoe-data-types.md#keywordrecord)** - Optional.
+- **offerArgs**: **[CopyRecord](/glossary/#passable)** - Optional.
```
dckc, line 293: comment: not having these kept in sync with agoric-sdk by machine is just maddening.

 - https://github.com/Agoric/agoric-sdk/issues/4291

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,63 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
-  Asset : quatloosPayment,
-  Price : moolaPayment
-};
-```
+**exit** specifies how the offer can be can cancelled. It must conform to one of three shapes:
+- `{ onDemand: null }`: (Default) The offering party can cancel on demand.
+- `{ waived: null }`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
+- `{ afterDeadline: deadlineDetails }`: The offer is automatically cancelled after a deadline,
+  as determined by its **timer** and **deadline** properties.
+  **timer** must be a timer, and **deadline** must be a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) value interpreted with respect to the timer.
```
dckc, line 322: deadline must be a bigint? I have recent experience using a `{ timerBrand, absValue }` record.

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,63 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
+  Some timers use Unix epoch time, while others count block height.
+  For more details, see [Timer Services](/reference/repl/timerServices.md).
+
+### Payments
 
-### OfferArgs
+**paymentKeywordRecord** must be either `undefined` or a **[PaymentKeywordRecord](./zoe-data-types.md#keywordrecord)**
```
dckc, line 328:
```suggestion
**paymentKeywordRecord** must be either `undefined` or a **[PaymentPKeywordRecord](./zoe-data-types.md#keywordrecord)**
```

https://github.com/Agoric/agoric-sdk/blob/0160e2fcb277340db2e3d25fca4c79af59b7e87c/packages/zoe/src/zoeService/types.js#L154

dckc (changes requested):
The keyword record stuff doesn't look right.

main/glossary/README.md
```diff
@@ -159,7 +159,7 @@ If they have ten houses for sale, they have ten different contract instances.
 
 ## Creator Invitation
 
-An [invitation](#invitation) optionally returned by [`startInstance()`](/reference/zoe-api/zoe.md#e-zoe-startinstance-installation-issuerkeywordrecord-terms-privateargs) that the contract instance
+An [invitation](#invitation) optionally returned by [`E(zoe).startInstance(...)`](/reference/zoe-api/zoe.md#e-zoe-startinstance-installation-issuerkeywordrecord-terms-privateargs) that the contract instance
```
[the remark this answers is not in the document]
Chris-Hibbert: It's a hope, with no concrete plan behind it. Not worth covering in the docs at this point.

main/guides/zoe/proposal.md
```diff
@@ -4,69 +4,85 @@
...
 };
 ```
-## Returned Value
 
-`offer()` returns a `UserSeat` object. Its name comes from the concept of "having a seat at the table" 
-for the contract's execution. 
+## Offer Arguments
+
+To pass additional arguments to the **offerHandler** contract code associated with the
+invitation, send them in an **offerArgs** record.
```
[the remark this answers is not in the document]
Chris-Hibbert: I thought that was official JavaScript terminology.

main/reference/zoe-api/zoe-data-types.md
Chris-Hibbert: The payments in `E(zoe).offer()` are a `PaymentKeywordRecord`.

main/reference/zoe-api/zoe.md
Chris-Hibbert: [deadline must be a bigInt](https://docs.agoric.com/reference/zoe-api/zoe.html#e-zoe-offer-invitation-proposal-paymentkeywordrecord-offerargs)

[TimerService also uses bigInts](https://docs.agoric.com/reference/repl/timerServices.html#timerservice-objects)

main/reference/zoe-api/zoe.md
Chris-Hibbert: Looks like you're right. I consider [the typeGuards](https://github.com/Agoric/agoric-sdk/blob/0160e2fcb277340db2e3d25fca4c79af59b7e87c/packages/zoe/src/typeGuards.js#L323) definitive. Thanks for the correction.

main/glossary/README.md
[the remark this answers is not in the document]
kriskowal: It might also be worth noting that `startInstance` is subsumed into `startUpgradable`. (It is, right?)

main/glossary/README.md
```diff
@@ -337,10 +335,9 @@ Keys can be used as elements of CopySets and CopyBags and as keys of CopyMaps (s
 
 ## Keyword
 
-A keyword is a property name string that is a valid
-[identifier](https://developer.mozilla.org/en-US/docs/Glossary/Identifier),
-starts with an upper case letter, and contains no non-ASCII characters.
-A KeywordRecord is a CopyRecord in which every property name is a keyword.
+A *Keyword* is a string that is a valid ASCII identifier
```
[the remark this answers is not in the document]
kriskowal: Aye. Do you mean _printable_ ASCII and _JavaScript_ identifier?

main/glossary/README.md
```diff
@@ -424,9 +421,14 @@ can immediately cause the [seat](#seat) to exit, getting back the amount it offe
 
 ## Passable
 
-A *passable* is something that can be marshalled (see the
-[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence))
-and sent to and from remote objects.
+A *passable* is something that can be sent to and from remote objects.
```
kriskowal, line 424:
```suggestion
A *passable* value is a JavaScript value that can be sent to and from remote objects.
```

main/glossary/README.md
kriskowal, line 429: I found this framing a little opaque so I’m suggesting a break-down:
```suggestion
Values that are pass-by-copy include:
* Primitive values including `string`, `number`, `bigint`, `boolean`, `null`, and `undefined`,
  * Agoric _tentatively_ supports registered symbols, as in `Symbol.for(name)`, though we advise against using them,
  * Agoric _even more tentatively_ supports well-known-symbols like `Symbol.asyncIterator`, though we strongly advise against using them,
* `CopyArray`: [hardened](#harden, acyclic, arrays like `harden(['foo', 'bar'])`,
* `CopyRecord`: hardened, acyclic objects like `harden({ keys: [0, 1], values: ['foo', 'bar'] })`,
* `CopyTagged`: [tagged](#tagged) pass-by-copy values like `{ [Symbol.for('passStyle')]: 'tagged', tag: 'copySet', payload: [1, 2, 3] }`.
```

I have not verified `#tagged` exists.

main/guides/zoe/proposal.md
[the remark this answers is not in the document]
kriskowal: I believe @erights coined `record` and in his interpretation, record means const-object-as-struct. The TC-39 Records and Tuples proposal uses the word to mean new-primitive-as-immutable-struct. TypeScript’s `Record` type suggests object-as-mutable-dictionary. I believe `offerArgs` is a const-object-as-dictionary. I am distinguishing “const” from “hardened” to mean: the informal contract is that this method will not mutate it and it consequently _may_ be hardened.

It might be best to just say `Object` here.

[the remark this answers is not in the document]
kriskowal: @gibson042 This is great. We saw places where offerArgs was typed `unknown` and others where it was typed `Object`. I think we need to also converge on the latter in the code.

main/reference/zoe-api/zoe-data-types.md
kriskowal: I’ve found that this works, but the browser is not obliged to scroll farther than necessary to put the top of this marker above the fold. Capturing content obliges the browser to scroll to show as much of the content as fits.

kriskowal (approved):
Thank you, suggestions only.

main/glossary/README.md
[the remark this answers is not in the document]
Chris-Hibbert: > It might also be worth noting that startInstance is subsumed into startUpgradable. (It is, right?)

That doesn't sound familiar. I think you're misremembering something, though I don't know what.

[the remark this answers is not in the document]
kriskowal: I dug into this. The Agoric chain home object can provide `startUpgradable` as a thin wrapper for `E(zoe).startInstance`. It’s provided by `vats/src/core/basic-behaviors.js`. This is why we were unable to find `startInstance` in the core proposal in the `dapp-game-places` template during the hackathon. We eventually found `startUpgradable` and from there were able to figure out how to thread `offerArgs` thru the level of indirection.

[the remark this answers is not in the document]
dckc: `startUpgradable` is a power from the bootstrap environment; it's available to core-evals, but it's not part of the Zoe API.

The purpose of `startUpgradable` is somewhat relevant, though: it's to make sure we don't lose admin facets needed for upgrade.

Do we tell folks in the Zoe docs that startInstance() also returns an admin facet, and if you lose it, you will NEVER BE ABLE TO UPGRADE?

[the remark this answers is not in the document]
dckc: `startUpgradable` should be covered in [docs for bootstrap powers](https://docs.agoric.com/guides/coreeval/permissions.html#declaring-required-capabilities). Those are pretty thin. That's probably worth a separate issue...

main/reference/zoe-api/zoe.md
dckc: > deadline must be a bigInt

Well, those docs are wrong (which is somewhat inevitable until we sync them by machine https://github.com/Agoric/agoric-sdk/issues/4291). As I say, I have recent experience to the contrary. Plus, the code says otherwise:

```js
      afterDeadline: {
        timer: M.eref(TimerShape),
        deadline: TimestampShape,
      },
```
https://github.com/Agoric/agoric-sdk/blob/6d9ae67c57499d7139dfce019b8c76f9ecdb825f/packages/zoe/src/typeGuards.js#L82-L85

```js
export const TimestampShape = M.or(TimestampRecordShape, TimestampValueShape);
```
https://github.com/Agoric/agoric-sdk/blob/6d9ae67c57499d7139dfce019b8c76f9ecdb825f/packages/time/src/typeGuards.js#L17C1-L17C79

main/glossary/README.md
dckc: Again, please, no novel info in the glossary. If we're going to have such details, let's have them in the [section on passables](https://docs.agoric.com/guides/js-programming/far.html#pass-styles-and-harden).

main/reference/zoe-api/zoe-data-types.md
gibson042: `id` is definitely the correct attribute (_my muscle memory here is very old and outdated_), but inner content is not necessary to establish what I know of as an "anchor link". This particular technique of [using an empty `a` element before a Markdown section heading](https://gist.github.com/rachelhyman/b1f109155c9dafffe618?permalink_comment_id=3483183#gistcomment-3483183) allows external links using an old target to still land in approximately the right place without suffering automatic slug generation issues that cause otherwise preferable Markdown like `## KeywordRecord <a id="amountkeywordrecord"></a>` to render incorrectly in many implementations, e.g.:
```diff
--- expected
+++ actual
-<h2 id="keywordrecord">KeywordRecord <a id="amountkeywordrecord"></a></h2>
+<h2 id="keywordrecord-a-id-amountkeywordrecord">KeywordRecord <a id="amountkeywordrecord"></a></h2>
```

main/reference/zoe-api/zoe.md
gibson042: > ... what is being given, what must be received to satisfy offer safety, and an exit rule ...
> 
> I'd say "is being given" because zoe requires that the proposal match the payments, and the contract won't see the offer if they don't match. (and the invitation will be burned.
> 
> "what must be received to satisfy offer safety" because the user can want any arbitrary thing (subject to `want patterns` provided by the contract), and it's a valid offer. If the contract doesn't match the want, then the offer gets its give back.

Done.

> Is there a reasonable place to describe `want patterns`?

I'll defer on this question, although https://docs.agoric.com/reference/zoe-api/zoe-data-types.html and https://docs.agoric.com/guides/zoe/proposal.html seem like reasonable locations for followup.

> Do we talk about `makeInvitation` from the contract author's perspective?

It's mentioned at https://docs.agoric.com/reference/zoe-api/zoe-contract-facet.html#zcf-makeinvitation-offerhandler-description-customproperties-proposalshape and https://docs.agoric.com/guides/zoe/#writing-and-installing-a-contract , but they could go further.

main/reference/zoe-api/zoe.md
gibson042: Done.

main/reference/zoe-api/zcfmint.md
gibson042: It's being added by this PR: https://github.com/Agoric/documentation/pull/876/files#diff-d7fc746f4ec4970ac29b349e9549ae68d4152a1c2e982580b3e1a1cef1ed6e9aR66

> ## KeywordRecord
> 
> A **KeywordRecord** is a [CopyRecord](/glossary/#copyrecord) in which every property name
> is a **[Keyword](#keyword)**, such as `harden({ Currency: quatloosBrand, CurrencyUnit: 100n })`.
> Subtypes further constrain property values (for example, an
> **AmountKeywordRecord** is a **KeywordRecord** in which every value is an
> **[Amount](/reference/ertp-api/ertp-data-types.md#amount)** and a
> **PaymentKeywordRecord** is a **KeywordRecord** in which every value is a
> **[Payment](/reference/ertp-api/payment.md)**).
> 
> Users submit their **payments** and receive their **payouts** as **KeywordRecords**:
> ```js
> quatloosPurse.deposit(payouts.Asset);
> const quatloos5 = AmountMath.make(quatloosBrand, 5n);
> const payments = {
>   Asset: quatloosPurse.withdraw(quatloos5),
> };
> ```

main/reference/zoe-api/zoe-data-types.md
gibson042: > That example seems weird. I don't believe anything in our `Record<Keyword, Brand | bigint>`. Did you mean to exhibit a `Record<Keyword, Amount<'nat'>>`? aka a `Record<Keyword, { brand: Brand<'nat'>, value: NatValue }>`

No, for that first example I specifically wanted a KeywordRecord that had heterogeneous value types and so was not an AmountKeywordRecord/PaymentKeywordRecord/etc. But if you think this one is confusing, perhaps it could drift even further away from our flavor of DeFi?
```suggestion
is a **[Keyword](#keyword)**, such as `harden({ Remotable: myMap, SupportedMethods: ['delete', 'get', 'has', 'set'] })`.
```

> I wonder if it's useful to generalize to KeywordRecord at all. If so, we might as well lean into TypeScript notation: `Record<Keyword, T>`

I thought about that and actually started with something similar, but ultimately decided not to because a) the site already uses AmountKeywordRecord and PaymentKeywordRecord, and b) we don't explain use of TypeScript syntax and AFAICT don't expect developers to know it. But a thought did occur to me in typing the [last response](https://github.com/Agoric/documentation/pull/876#discussion_r1409666981)—what would you think about using it to help explain the concept for people who are familiar with it?

> ## KeywordRecord
> 
> A **KeywordRecord** is a [CopyRecord](/glossary/#copyrecord) in which every property name
> is a **[Keyword](#keyword)**, such as `harden({ Currency: quatloosBrand, CurrencyUnit: 100n })`.
> Subtypes further constrain property values (for example, an
> **AmountKeywordRecord** is a **KeywordRecord\<Amount>** in which every value is an
> **[Amount](/reference/ertp-api/ertp-data-types.md#amount)** and a
> **PaymentKeywordRecord** is a **KeywordRecord\<Payment>** in which every value is a
> **[Payment](/reference/ertp-api/payment.md)**).

main/reference/zoe-api/zoe-data-types.md
gibson042: That should be fixed, but I don't have the necessary context. Would either of you be willing to submit a followup or commit in this PR?

main/reference/zoe-api/zoe-data-types.md
dckc: > ... for that first example I specifically wanted a KeywordRecord that had heterogeneous value types ...

Such an example is a distraction, no? There is no such data type in the Zoe API.

This is the core of my request for changes.

Chris-Hibbert: I agree. There are no KeywordRecords with heterogeneous value types, and I don't think we're likely to add them.

main/guides/js-programming/far.md
gibson042, line 1: Tagging @erights for review now that the scope has expanded to include this file.

main/guides/zoe/proposal.md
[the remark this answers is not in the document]
gibson042: Updated.

> To pass additional arguments to the **offerHandler** contract code associated with the
> invitation, send them in an **offerArgs** [CopyRecord](/glossary/#copyrecord).

main/reference/zoe-api/zoe-data-types.md
gibson042: Okay, updated.

> A **KeywordRecord** is a [CopyRecord](/glossary/#copyrecord) in which every property name
> is a **[Keyword](#keyword)**, such as `harden({ Asset: moolaIssuer, Bid: simoleanIssuer })`.

main/reference/zoe-api/zoe.md
gibson042: Fixed.

main/reference/zoe-api/zoe-data-types.md
dckc: A single letter would be a big improvement: change `PaymentKeywordRecord` to `PaymentPKeywordRecord`.

It would be more precise to say "in which every value is a Payment or a Promise for a Payment" but I think it's OK to leave the Promise stuff implicit in this discussion.

gibson042: Done.

dckc (approved):
keyword stuff is addressed to my satisfaction

main/glossary/README.md
```diff
@@ -337,10 +350,9 @@ Keys can be used as elements of CopySets and CopyBags and as keys of CopyMaps (s
```
erights, line 350: Above

> ```md
> [Values](#value) must be Keys.
> ```

Not in general. What this may originally have tried to say is that
- CopySets and CopyBags are Keys.
- CopyMaps are Keys iff all their values are Keys.

main/glossary/README.md
```diff
@@ -424,9 +436,15 @@ can immediately cause the [seat](#seat) to exit, getting back the amount it offe
...
 
-A *passable* is something that can be marshalled (see the
-[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence))
-and sent to and from remote objects.
+A *passable* is something that can be sent to and from remote objects.
+Passables include pass-by-copy primitive values such as numbers and strings and
+pass-by-reference values such as Remotables and Promises.
+Passables also include [hardened](#harden) acyclic pass-by-copy containers that
+recursively terminate in non-container passables, such as
+[CopyArrays](#copyarray) and [CopyRecords](#copyrecord).
```
erights, line 444: > ```md
>  that recursively terminate in non-container passables, such as
> [CopyArrays](#copyarray) and [CopyRecords](#copyrecord).
> ```

That makes it sound like CopyArrays and CopyRecords are examples of "non-container passables".

main/glossary/README.md
```diff
@@ -424,9 +436,15 @@ can immediately cause the [seat](#seat) to exit, getting back the amount it offe
...
+
+For more information, see the
+[Marshaling section in the JavaScript Distributed Programming Guide](/guides/js-programming/far.md#marshaling-by-copy-or-by-presence).
 
 ## Payment
```
erights, line 450: Below

> If there was a previous reallocation, the payout is different than what the user escrowed.

This is fine as it. No change suggested. But by itself it sounds too dangerous. How about

> If there was a previous reallocation, the payout is different than what the user escrowed, but still constrained by Offer Safety (appropriate link to earlier glossary entry)

main/glossary/README.md
```diff
@@ -459,18 +477,21 @@ For more information, see the [JavaScript Distributed Programming Guide](/guides
 
 ## Proposal
 
-Proposals are records with `give`, `want`, and `exit` properties. [Offers](#offer) must include a proposal, which states
-what asset you want, what asset you will give for it, and how/when the offer maker can cancel the offer
-(see [Exit Rule](#exit-rule) for details on the last). For example:
+Proposals are records with `give`, `want`, and/or `exit` properties respectively
+expressing [offer](#offer) conditions regarding what assets will be given,
+what must be received in exchange to satisfy offer safety, and
```
erights, line 482: Possible misunderstanding that offer safety means you'll get what you said you want. At least make "offer safety" and explicit link to that glossary entry. Avoiding the possible misunderstanding would be even better.

erights (commented):
Just comments so far, only on the README

erights (changes requested):
A few more. But I clearly won't finish tonight.

Btw, I am not trying to check or follow any links, as I cannot do so simply in the markdown preview mode.

main/reference/zoe-api/zoe-contract-facet.md
```diff
@@ -235,7 +236,7 @@ zcf.assertUniqueKeyword(keyword);
 - Returns: None.
 
 Prohibit invocation of invitatations whose description include any of the strings.
-Any of the strings that end with a colon (:) will be treated as a prefix,
+Any of the strings that end with a colon (`:`) will be treated as a prefix,
 and invitations whose description string begins with the string (including the colon)
 will be burned and not processed if passed to **E(Zoe).offer()**.
```
erights, line 241: @Chris-Hibbert , are they actually burned? That surprises me.

main/reference/zoe-api/zoe.md
erights: We do not yet support want patterns. Looking forward to it though!

main/reference/zoe-api/zoe-data-types.md
erights: > KeywordRecord that had heterogeneous value types

Please don't. None of our keyword records are heterogeneous, and likely none ever will be, since the point of keywords as names is to support lookup by `[]` user-defined names, and therefore computed property names.

main/reference/zoe-api/zoe.md
erights: > comment: not having these kept in sync with agoric-sdk by machine is just maddening

Indeed!

main/glossary/README.md
```diff
@@ -288,7 +301,7 @@ is available [here](https://www.computerweekly.com/blog/Open-Source-Insider/What
 A [payment](#payment) whose amount represents (and is required for) participation in a contract instance.
```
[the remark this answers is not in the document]
gibson042: Already fixed by #877, which was opened in parallel.

main/glossary/README.md
```diff
@@ -288,7 +301,7 @@ is available [here](https://www.computerweekly.com/blog/Open-Source-Insider/What
...
 Contracts often return a creator invitation on their instantiation, in case the contract instantiator wants
 to immediately participate. Otherwise, the contract instance must create any additional invitations.
-Every [offer](#offer) to participate in a contract instance must include an invitation to that instance in the first argument to [`E(Zoe).offer()`](/reference/zoe-api/zoe.md#e-zoe-offer-invitation-proposal-paymentkeywordrecord-offerargs), and any wallet receiving one will validate it via the [InvitationIssuer](#invitationissuer).
+Every [offer](#offer) to participate in a contract instance must include an invitation to that instance in the first argument to [`E(zoe).offer(...)`](/reference/zoe-api/zoe.md#e-zoe-offer-invitation-proposal-paymentkeywordrecord-offerargs), and any wallet receiving one will validate it via the [InvitationIssuer](#invitationissuer).
 
 An invitation's [amount](#amount) includes the following properties:
 - The contract's installation in Zoe, including access to its source code.
```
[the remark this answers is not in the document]
gibson042: I'm not sure how to rephrase this into a meaningful glossary entry for the InvitationIssuer, so can you open an issue or PR indicating direction?

main/glossary/README.md
gibson042: Hmm, do you have an example of a non-Key Value?

main/guides/js-programming/far.md
```diff
@@ -28,17 +28,28 @@ Another vat can then make and use the exported counters:
...
 Every object exported from a smart contract, such as `publicFacet` or
 `creatorFacet`, must be passable. All objects used in your contract's external API must
-be passable. All passables must be hardened.
+be passable.
 
-Consider what might happen if we had a remote `item` and we did not harden
+All Passables must be hardened. Consider what might happen if we had a remote `item` and we did not harden
 some pass-by-copy data that we passed to it:
 
 ```js
```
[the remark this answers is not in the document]
gibson042: I don't think so, but it's worth reconsidering in a followup.

main/guides/js-programming/far.md
```diff
@@ -79,24 +91,24 @@ this prevents malicious clients from playing tricks with cyclic data etc.
 
 <<< @/snippets/test-distributed-programming.js#importFar
 
-`Far(farName, object-with-methods)`
+`Far(farName, objectWithMethods)` marks an object as Remotable.
 
 - `farName` `{ String }`
```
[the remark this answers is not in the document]
gibson042: #883

main/glossary/README.md
```diff
@@ -230,10 +245,8 @@ See [`E()`](#e) above.
...
 ## Exit Rule
 
-Part of an [offer](#offer) specifying how the offer can be cancelled/exited. There are three values:
-- `onDemand: null`: (Default) The offering party can cancel on demand.
-- `waived: null`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
-- `afterDeadline: {…}`: The offer is automatically cancelled after a deadline, as determined by its `timer` and `deadline` properties. See [Proposals and payments](/reference/zoe-api/zoe.md#proposals-and-payments).
+An object specifying how an [offer](#offer) can be cancelled, such as on demand or by a deadline.
+For details, see [`E(zoe).offer(...)`](/reference/zoe-api/zoe.md#proposals).
 
 ## Facet
```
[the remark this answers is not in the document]
gibson042: Done, but saving for a followup.

main/glossary/README.md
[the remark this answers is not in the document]
dckc: I looked for a relevant issue and found:
 - https://github.com/Agoric/agoric-sdk/issues/5775

But it's one of those "do X" issues that doesn't say _why_ do X... other than
> That functionality can be supplied from the creatorFacet.

Chris-Hibbert (commented):
> Btw, I am not trying to check or follow any links, as I cannot do so simply in the markdown preview mode.

I am planning to pull down the whole PR and do the local build and review for link consistency, but I'm waiting for the changes to settle down.

main/reference/zoe-api/zcfmint.md
```diff
@@ -13,7 +13,7 @@ Returns an **IssuerRecord** containing the **[Issuer](/reference/ertp-api/issuer
...
   - **zcfSeat**: **[ZCFSeat](./zcfseat.md)** - Optional.
   - Returns: **ZCFSeat**
```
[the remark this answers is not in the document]
Chris-Hibbert: That seems like a good interpretation, but we don't refer to keywords as a namespace anywhere else, so better to recast this in more familiar terminology.

main/reference/zoe-api/zoe-contract-facet.md
Chris-Hibbert: Your instincts are correct. If the invitation is blocked, we `Fail` out of the `offer()` call and don't get to the immediately following `burnInvitation()`.
```
    !instanceAdmin.isBlocked(description) ||
      Fail`not accepting offer with description ${q(description)}`;
    const { invitationHandle } = await burnInvitation(
      invitationIssuer,
      invitation,
    );
```

main/glossary/README.md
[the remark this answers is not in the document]
erights: Thanks for finding that. In light of @samsiegart 's two comments at https://github.com/Agoric/agoric-sdk/issues/5775#issuecomment-1561870319 , I withdraw the request to emphasize the deprecation of creator invitation until Sam's question is resolved.

main/glossary/README.md
```diff
@@ -337,10 +350,9 @@ Keys can be used as elements of CopySets and CopyBags and as keys of CopyMaps (s
 
 ## Keyword
 
-A keyword is a property name string that is a valid
-[identifier](https://developer.mozilla.org/en-US/docs/Glossary/Identifier),
-starts with an upper case letter, and contains no non-ASCII characters.
-A KeywordRecord is a CopyRecord in which every property name is a keyword.
+A *Keyword* is a string that is an ASCII-only [identifier](https://developer.mozilla.org/en-US/docs/Glossary/Identifier)
+and starts with an upper case letter.
+See **[Zoe Data Types](/reference/zoe-api/zoe-data-types.md#keyword)**.
```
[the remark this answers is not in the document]
erights: yes. I like the idea of delegating details to a common entry. But the non-detail text should still be correct by itself.

main/guides/js-programming/far.md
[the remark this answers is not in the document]
erights: Thx. With that filed, resolving the conversation here.

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -83,11 +54,34 @@ is exclusively yours and no one else has access to it.
...
+is a **[Keyword](#keyword)**, such as `harden({ Asset: moolaIssuer, Bid: simoleanIssuer })`.
+Subtypes further constrain property values (for example, an
+**AmountKeywordRecord** is a **KeywordRecord** in which every value is an
+**[Amount](/reference/ertp-api/ertp-data-types.md#amount)** and a
+**PaymentPKeywordRecord** is a **KeywordRecord** in which every value is either a
+**[Payment](/reference/ertp-api/payment.md)** or a Promise for a Payment).
+
+Users submit their **payments** and receive their **payouts** as **KeywordRecords**:
+```js
+quatloosPurse.deposit(payouts.Asset);
```
erights, line 78: Since `payouts.Asset` may be a promise for a payment, and since `purse.deposit` demands a non-promise for its payment argument, this example is incorrect or at best misleading.

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -63,14 +35,13 @@ You can get information about the contract instance via these methods:
 
 ## Invitation
 
-An **Invitation** is a kind of **[Payment](/reference/ertp-api/payment.md)**. It's a non-fungible eright
-that can be held in **Payments** or **[Purses](/reference/ertp-api/purse.md)**, just like any other
-eright. An **Invitation** **Payment** is a **Payment** holding an **Invitation**.
+An **Invitation** is a **[Payment](/reference/ertp-api/payment.md)** holding a non-fungible eright created by the **[InvitationIssuer](#invitationissuer)**.
+An **Invitation Payment** is a **Payment** holding an **Invitation**.
```
erights, line 39: These cannot both be true. In fact, the second sentence should be deleted. There is no distinct concept of "Invitation Payment" as distinct from "Invitation", since, as the first sentence says, an "Invitation" is already a kind of Payment.

erights: There is however a genuine terminology problem here. A purse or payment of the invitation issuers can have an amount with any number of invitation details, including none. Perhaps we need to switch to using "invitation" to describe the abstract non-fungible assets, which can move between purses and payments. In the same way that we wouldn't say that a dollar "is" a payment with a 1 dollar amount, we should probably stop saying that an invitation "is" a payment with 1 invitations detail amount.

What a mess. Would be a lot of work to clean up, but probably needed. Just do some adequate stopgap in this PR and file an issue? In any case, the above pair of sentences still need to be fixed because they contradict each other.

erights: Exclusively transferable assets cannot be reified as objects. Payments and Purses reify only containers for such assets, and amounts reify only descriptions of these assets. But the assets themselves only exist at a higher layer of abstraction. Ideally, "invitation" refers to the asset, not the payment holding the asset. But this contradicts a lot of current usage.

Hmm. Perhaps then the second sentence is more correct than the first, or should be going forward?

main/reference/zoe-api/zoe-data-types.md
```diff
@@ -63,14 +35,13 @@ You can get information about the contract instance via these methods:
...
 
 ## InvitationIssuer
 
-The **InvitationIssuer** is a special type of **[Issuer](/reference/ertp-api/issuer.md)**. The single Zoe
-instance has an **InvitationIssuer** for the entirety of its lifetime. All **Invitations** come from the
+The **InvitationIssuer** is a special type of **[Issuer](/reference/ertp-api/issuer.md)**.
+Zoe has a single **InvitationIssuer** for its entire lifetime. All **Invitations** come from the
 **[Mint](/reference/ertp-api/mint.md)** associated with the Zoe instance's **InvitationIssuer**.
 
 **InvitationIssuer** has all the methods of regular **Issuers**, but the two methods that are most
```
erights, line 47: > ```md
> two methods that are most
> often used are **[anIssuer.claim()](/reference/ertp-api/issuer.md#anissuer-claim-payment-optamount)**
> and **[anIssuer.getAmountOf()](/reference/ertp-api/issuer.md#anissuer-getamountof-payment)**.
> ```

There is no longer an issuer.claim method.

main/guides/zoe/proposal.md
```diff
@@ -4,69 +4,85 @@
...
 - A **proposal** stating your offer conditions.
-- The **payments** escrowed for the offer, each in association with a `proposal`-specified keyword.
+- The **payments** escrowed for the offer, each corresponding with a **give** [Keyword](/reference/zoe-api/zoe-data-types.md#keyword) in **proposal**.
+- **offerArgs** expressing additional arguments for the **offerHandler** associated with the invitation by [`zcf.makeInvitation(...)`](/reference/zoe-api/zoe-contract-facet.md#zcf-makeinvitation-offerhandler-description-customproperties-proposalshape).
 
 ## Invitations
 
-`Invitations` are a special case of ERTP `payments`. They are linked to a specific contract `instance`, and
-having one gives you the right to participate in that contract `instance`, for example, by making offers in it.
+An [Invitation](/reference/zoe-api/zoe-data-types.md#invitation) is a special case of ERTP [Payment](/reference/ertp-api/payment.md). Each is linked to a specific contract [Instance](/reference/zoe-api/zoe-data-types.md#instance), and
```
erights, line 15: Just noting that this is another example where we say that the payment is the invitation, rather than saying that the invitation is the abstract asset that is conveyed by a payment. No change suggested for this PR.

main/guides/zoe/proposal.md
```diff
@@ -4,69 +4,85 @@
...
+having one gives you the right to participate in that contract instance by making offers in it.
 
-There are two main ways for contract users to get an `invitation`:
-- If you create the contract `instance`, you get a special creator `invitation`.
-- Someone (possibly you) who holds the right objects has created an `invitation` for a contract `instance` and gives it to
+There are two main ways for contract users to get an invitation:
```
erights, line 18: Does the generality of the second bullet cover everything not covered by the first bullet? I think so. If so, then these are not just the "two main ways", since that phrase implies that some case isn't covered.

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,67 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
- 
-```js
-const paymentKeywordRecord = {
-  Asset : quatloosPayment,
-  Price : moolaPayment
-};
-```
+**exit** specifies how the offer can be cancelled. It must conform to one of three shapes:
+- `{ onDemand: null }`: (Default) The offering party can cancel on demand.
+- `{ waived: null }`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
```
erights, line 323:
```suggestion
- `{ waived: null }`: The offering party can't cancel and relies entirely on the smart contract to complete (finish or fail) their offer.
```

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,67 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
+- `{ afterDeadline: deadlineDetails }`: The offer is automatically cancelled after a deadline,
+  as determined by its **timer** and **deadline** properties.
+  **timer** must be a timer, and **deadline** must be timestamp understood by it.
+  (Some timers use Unix epoch time, while others count block height.)
+  For more details, see [Timer Services](/reference/repl/timerServices.md).
+
+### Payments
 
-### OfferArgs
+**paymentKeywordRecord** must be either `undefined` or a **[PaymentPKeywordRecord](./zoe-data-types.md#keywordrecord)**
```
erights, line 332: Can we rename this to... ?

```suggestion
**paymentPKeywordRecord** must be either `undefined` or a **[PaymentPKeywordRecord](./zoe-data-types.md#keywordrecord)**
```

main/reference/zoe-api/zoe.md
```diff
@@ -286,73 +286,67 @@ const { creatorFacet, publicFacet, creatorInvitation } = await E(Zoe).startInsta
...
+});
+```
 
+<a href="offerargs"></a>
+### Offer Arguments
 
-Contract code should be careful interacting with *offerArgs*. These values need input validation
-before being used by the contract code since they are coming directly from the user and may
-have malicious behavior.
+**offerArgs** is an object that can be used to pass additional arguments to the
```
erights, line 346: How about... ?

```suggestion
**offerArgs** is an optional CopyRecord that can be used to pass additional arguments to the
```

erights (commented):
This review pass complete. Over to you ;)

main/guides/js-programming/far.md
```diff
@@ -28,17 +28,28 @@ Another vat can then make and use the exported counters:
...
+
+- Pass-by-copy **primitive** values: `undefined`, `null`, booleans `true` and `false`, numbers,
+  [BigInts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt),
+  strings, and symbols that are either
+  [well-known](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#well-known_symbols) or
+  [registered](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#shared_symbols_in_the_global_symbol_registry).
+- Acyclic pass-by-copy **containers** that recursively terminate in non-container passables.
+  Such containers include _CopyArrays_ like `harden(['foo', 'bar'])`,
+  _CopyRecords_ like `harden({ keys: [0, 1], values: ['foo', 'bar'] })`, and
+  _CopyTaggeds_ representing higher-order types like CopySet, CopyBag, and CopyMap.
```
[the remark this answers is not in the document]
gibson042: https://github.com/endojs/endo/pull/1868

main/guides/zoe/proposal.md
```diff
@@ -4,69 +4,85 @@
...
+rather than a [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number))
+
+::: warning Note
+Amounts are just _descriptions_ of assets, and have no intrinsic value of their own.
+In contrast, [Payments](/reference/ertp-api/payment.md) hold actual digital assets.
+:::
+
+**exit** specifies how the offer can be cancelled. It must conform to one of three shapes:
+- `{ onDemand: null }`: (Default) The offering party can cancel on demand.
+- `{ waived: null }`: The offering party can't cancel and relies entirely on the smart contract to promptly finish their offer.
```
gibson042, line 55:
```suggestion
- `{ waived: null }`: The offering party can't cancel and relies entirely on the smart contract to complete (finish or fail) their offer.
```

main/reference/zoe-api/zoe-data-types.md
gibson042: I'm so glad my confusion here was warranted! Changed to this:
> An **Invitation** is a non-fungible asset created by the **[InvitationIssuer](#invitationissuer)**.
> An **Invitation Payment** is a **[Payment](/reference/ertp-api/payment.md)** holding an **Invitation**.

main/reference/zoe-api/zoe-data-types.md
gibson042: There are still references to it all over this site, so I'm going to keep this in place for whatever work cleans up all of them: #885

main/glossary/README.md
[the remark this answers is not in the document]
gibson042: #886

main/glossary/README.md
gibson042: Sorry, let me elaborate... the full section currently reads like this:
> ## Value
> 
> Values are the part of an [amount](#amount) that describe the value of something
> that can be owned or shared: How much, how many, or a description of a unique asset, such as
> Pixel(3,2), $3, or “Seat J12 for the show September 27th at 9:00pm”.
> [Fungible](#fungible) values are usually represented by natural numbers.
> Other values may be represented as a CopySet of strings naming particular rights or
> arbitrary objects representing the rights directly (usually [non-fungible](#non-fungible) assets).
> Values must be [Keys](#key).
> 
> For more information, see the [ERTP documentation's Value section](/guides/ertp/amounts.md#values).

Are matchers/promises/errors/etc. allowed as the `value` part of an [Amount](https://docs.agoric.com/glossary/#amount)?

main/reference/zoe-api/zcfmint.md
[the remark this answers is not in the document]
gibson042: My best-effort rewording:

> ## aZCFMint.mintGains(gains, zcfSeat?)
>   - **gains**: **[AmountKeywordRecord](./zoe-data-types.md#keywordrecord)**
>   - **zcfSeat**: **[ZCFSeat](./zcfseat.md)** - Optional.
>   - Returns: **ZCFSeat**
> 
> All **amounts** in *gains* must be of this **ZCFMint**'s **[Brand](/reference/ertp-api/brand.md)**
> and the *gains*' **[Keywords](./zoe-data-types.md#keyword)** should be defined by the contract instance in which *zcfSeat* is participating.
> If *zcfSeat* is not provided, a new **seat** is used.
> Mints the *gains* **Amount** of assets and adds them to *zcfSeat*'s **[Allocation](./zoe-data-types.md#allocation)**, then returns *zcfSeat*.
> 
> ## aZCFMint.burnLosses(losses, zcfSeat)
>   - **losses**: **[AmountKeywordRecord](./zoe-data-types.md#keywordrecord)**
>   - **zcfSeat**: **[ZCFSeat](./zcfseat.md)**
>   - Returns: None
> 
> All **amounts** in *losses* must be of this **ZCFMint**'s **[Brand](/reference/ertp-api/brand.md)**
> and the *losses*' **[Keywords](./zoe-data-types.md#keyword)** must be defined by the contract instance in which *zcfSeat* is participating.
> Subtracts *losses* from *zcfSeat*'s **[Allocation](./zoe-data-types.md#allocation)**, then
> burns that **amount** of assets from the pooled **[Purse](/reference/ertp-api/purse.md)**.

main/glossary/README.md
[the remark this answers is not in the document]
gibson042: Done.

gibson042:
Thanks for the astonishingly thorough review, @erights! I believe I've addressed all feedback.

main/reference/zoe-api/zoe-data-types.md
dckc: Hm. I suggest escalating that to its own issue. There's a significant training burden around changing this sort of terminology. Getting everybody to stop saying "an invitation is a payment" is going to take some time/effort. We should perhaps have more discussion with devrel/product about how we want to talk about these things.

I suppose I don't object to changing this occurrence here/now. But let's not suppose that's all the work necessary to effect this change.

gibson042: #887

gibson042:
@Chris-Hibbert updated with your suggestions.

main/glossary/README.md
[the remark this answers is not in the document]
erights: Ah. Resolving.

main/glossary/README.md
erights: Clarified elsewhere that the `Value` def above is `AmountValue`, i.e., the `value` part of an `Amount`. But when I said above

> CopyMaps are Keys iff all their values are Keys.

I meant "values" as in the values of a key-to-value mappings in a CopyMap.

main/glossary/README.md
```diff
@@ -459,18 +477,21 @@ For more information, see the [JavaScript Distributed Programming Guide](/guides
...
-  exit: { onDemand: null }
-})
+  give: { Asset: AmountMath.make(quatloosBrand, 4n) },
+  want: { Price: AmountMath.make(moolaBrand, 15n) },
+  exit: { onDemand: null },
+});
 ```
-`give` and `want` use [keywords](#keyword) defined by the contract. Each specifies via an [amount](#amount) a description of what
-they are willing to give or want to get.
+`give` and `want` each associate [Keywords](#keyword) defined by the contract with corresponding [Amounts](#amount) describing what will be respectively given or received.
```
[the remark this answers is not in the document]
erights: (I see it's done, so resolving

main/guides/js-programming/far.md
```diff
@@ -28,17 +28,28 @@ Another vat can then make and use the exported counters:
...
-- Promises for passables.
+A Passable is a [hardened](/glossary/#harden) value that can be marshalled.
+There are four broad kinds of Passable:
+
+- Pass-by-copy **primitive** values: `undefined`, `null`, booleans `true` and `false`, numbers,
+  [BigInts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt),
+  strings, and symbols that are either
+  [well-known](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#well-known_symbols) or
+  [registered](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#shared_symbols_in_the_global_symbol_registry).
+- Acyclic pass-by-copy **containers** that recursively terminate in non-container passables, such as
```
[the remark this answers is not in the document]
erights: (I see it's done, so resolving

main/reference/zoe-api/zoe-contract-facet.md
```diff
@@ -71,27 +71,30 @@ await zcf.saveIssuer(secondaryIssuer, keyword);
...
+and returning arbitrary offer results.
+
+**description** is a required string describing the **Invitation**,
 and should include whatever information is needed for a potential recipient of the **Invitation**
-to know what they are getting in the optional *customProperties* argument, which is
-put in the **Invitation**'s **value**.
+to know what they are getting in the optional **customProperties** argument, which is
+put in the **Invitation**'s **amount**.
+Each one should be a string literal that is unique within its contract and
+used only as the argument to make invitations of a particular kind.
```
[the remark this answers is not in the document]
Chris-Hibbert: Oh, man, that is not a small topic, and I don't see it touched on anywhere in these docs. The only reference material for it that I can find is in [endo](https://github.com/endojs/endo/blob/f7979e41a7e83db8bce855c92a34cabcc3bcfa3f/packages/patterns/src/types.js), and it is far from adequate as developer documentation.

I would say "a Pattern[ref] describing the required and allowed components of the proposal. Proposals that don't match the pattern will be rejected by Zoe without interacting with the contract."  I won't claim that's adequate, but to say any more would take a week.

main/reference/zoe-api/zoe-data-types.md
erights: With that issue filed, resolving here.

main/reference/zoe-api/zoe-data-types.md
erights: With that issue filed, resolving here.

main/guides/zoe/proposal.md
erights: With issue about this filed, resolving here.

main/guides/zoe/proposal.md
```diff
@@ -4,69 +4,85 @@
...
-- The **payments** escrowed for the offer, each in association with a `proposal`-specified keyword.
+- The **payments** escrowed for the offer, each corresponding with a **give** [Keyword](/reference/zoe-api/zoe-data-types.md#keyword) in **proposal**.
+- **offerArgs** expressing additional arguments for the **offerHandler** associated with the invitation by [`zcf.makeInvitation(...)`](/reference/zoe-api/zoe-contract-facet.md#zcf-makeinvitation-offerhandler-description-customproperties-proposalshape).
 
 ## Invitations
 
-`Invitations` are a special case of ERTP `payments`. They are linked to a specific contract `instance`, and
-having one gives you the right to participate in that contract `instance`, for example, by making offers in it.
+An [Invitation](/reference/zoe-api/zoe-data-types.md#invitation) is a special case of ERTP [Payment](/reference/ertp-api/payment.md). Each is linked to a specific contract [Instance](/reference/zoe-api/zoe-data-types.md#instance), and
+having one gives you the right to participate in that contract instance by making offers with it.
```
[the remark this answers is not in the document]
erights: Those pronouns. Dereference them!

Chris-Hibbert (approved):

main/reference/zoe-api/zoe-contract-facet.md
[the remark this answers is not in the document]
erights: I still think you need to explicitly say that they are unique to a static call site within this contract. Or better, within this contract's source text. Otherwise, someone might misread to think they need to be unique per dynamic call to `makeInstance`.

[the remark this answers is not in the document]
Chris-Hibbert: @erights's comment applies to the description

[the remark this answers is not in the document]
erights: Yes. Only the description string. Thx.

gibson042:
Thanks all!

erights (approved):
LGTM

Please do look at all unresolved conversations to see if there's anything unaddressed before merging. Otherwise good to go. Awesome improvements, thanks!

PR #882, opened 2023-11-29
feat: Improve the jump-to-fragment fixup for non-heading targets

gibson042:
Router logic was swallowing history entries and not scrolling far enough for such targets[^1] to clear the fixed header.

[^1]: which do not currently exist but are being added, cf. https://github.com/Agoric/documentation/pull/876#discussion_r1409657230

### DIFF
--- a/main/.vuepress/config.js
+++ b/main/.vuepress/config.js
@@ -53,12 +53,57 @@ module.exports = {
       }
     });
 
-    if (location.hash) {
-      // Re-navigate to the page target once content has loaded.
+    if (location.hash && location.hash !== '#') {
+      // Once content has loaded, re-navigate to the page target
+      // without triggering interfering router/history/scroll logic.
+      const hash = location.hash;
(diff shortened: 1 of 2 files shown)


main/.vuepress/config.js
```diff
@@ -53,12 +53,56 @@ module.exports = {
...
+          }
+          console.log('suppress', evt.type, { __proto__: evt, ...props });
+        };
+        const stopEvents = types => {
+          const restorers = types.map(type => {
+            window.addEventListener(type, stopPropagation, true);
+            return () => window.removeEventListener(type, stopPropagation, true);
+          });
+          const passEvents = () => {
+            while (restorers.length > 0) restorers.pop()();
```
michaelfig, line 78: Any reasons for `restorers.pop()` instead of `restorers.shift()`?  Maybe add a comment as to why, or why it doesn't matter.

michaelfig (approved):
Really nice!  I'm impressed. 🥇

PR #886, opened 2023-12-01
chore: Improve descriptions per code review

gibson042:
Stacked on #876

### DIFF
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -39,7 +39,7 @@ by [issuers](#issuer) and [mints](#mint), and represent the goods and currency c
 [purses](#purse) and [payments](#payment). They represent things like currency, stock, and the
 abstract right to participate in a particular exchange.
 
-An amount is comprised of a [brand](#brand) with a [value](#value). For example, "4 Quatloos"
+An amount is comprised of a [brand](#brand) with a [value](#amountvalue). For example, "4 Quatloos"
 is an amount with a value of "4" and a brand of the imaginary currency "Quatloos".
 
 **Important**: Amounts are *descriptions* of digital assets, not the actual assets. They have no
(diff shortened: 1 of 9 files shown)


erights:
@gibson042 , since it says "Stacked on #876" I changed the github comparison base to reflect that. Makes a lot more sense to me now.

main/glossary/README.md
```diff
@@ -554,13 +554,11 @@ it set at $10. They can specify the instance's minimum bid amount in its terms.
 
 ## Value
```
erights, line 555:
```suggestion
## AmountValue
```

main/reference/ertp-api/ertp-data-types.md
```diff
@@ -36,12 +36,12 @@ someAmountShape: {
 
 ## Value
```
erights, line 37:
```suggestion
## AmountValue
```

gibson042:
> I also saw your related question on the base PR. The problem is that we use the term "value" in two contradictory ways: As the value part of an amount, or as the value of a map's key-value mapping. The unqualified term "value" should only refer to the latter. For the value part of an amount, the actual name of the type in the code is already `AmountValue`.
> 
> With this straightened out, any AmountValue is a Key. A value of a CopyMap's key-value mapping may not be a Key.

Done, but it does introduce some minor reading friction: https://github.com/Agoric/documentation/pull/886/commits/fa576aece9f13254e4a4f0625b10d7fba0fc404d

main/glossary/README.md
erights: Seems done. Resolving.

main/reference/ertp-api/ertp-data-types.md
erights: Seems done. Resolving.

Issue #582, opened 2021-11-30
interactive jsfiddle style ERTP, Zoe examples


dckc:
@erights I got one call to `makeIssuerKit` sort of working: https://jsfiddle.net/q06aco3s/3/

 - [ ] I had to preface import specifiers with https://esm.sh/ :

```
import 'https://esm.sh/@agoric/install-ses';
import { makeIssuerKit } from 'https://esm.sh/@agoric/ertp';

const t = makeIssuerKit('blort');
console.log(t)
```

 - [ ] console output is censored
   - https://esm.sh/@agoric/install-ses/debug.js might work , once @michaelfig finishes publishing it to npm

cc @kennyrowe @samsiegart @kriskowal

kriskowal:
That’s neat.

I added import ***@***.***/eventual-send/shim.js';

That gets you to the next error: “passStyleOf” is not defined. I’m not sure
what to make of that.

dckc:
oh... I just neglected to save the edit where I fixed the `passStyleOf` problem: https://jsfiddle.net/p8L7jmgw/

kriskowal:
Huzzah!

dckc:
The `esm.sh` technique doesn't seem to work today. `unpkg.com` seems to work, though:

```
import 'https://unpkg.com/@agoric/install-ses@0.5.29/install-ses.js?module';
import { E, Far, passStyleOf } from 'https://unpkg.com/@agoric/far@0.1.1/src/index.js?module';
```
 -- https://jsfiddle.net/dx8jc05f/

Tyrosine22:
Am I correct in understanding that this isn't really a doc bug at all? It's more of a testing of software capabilities & how to use software? If so... could we move this discussion, or close it out?

dckc:
It's an enhancement request. A request that many / all of the examples in our docs would have a little button next to them where you can run them and tinker them a la jsfiddle.

erights:
I am very excited by this! But I followed the links and hit "Run" and didn't see anything helpful. So I brought up the developer tools to see if the console output was there, but still didn't see anything helpful. Let's interactively debug in my browser sometime.

Thanks!

dckc:
## Hardened JavaScript works on Glitch

https://glitch.com/edit/#!/rainy-voracious-fiber

0xpatrickdev:
\+ **Codesandbox Dev Containers**  https://codesandbox.io/p/devbox/vigilant-bhabha-326wn4


[...]

Issue #436, opened 2020-11-11
Diagram Invitation/Offer/Seat interaction


rowgraus:

dckc:
done with several diagrams in https://docs.agoric.com/guides/zoe/#trading-with-offer-safety

for example:

![zoe API sequence diagram](https://docs.agoric.com/assets/trade-offer-safety-3.u_IxArtY.svg)

PR #889, opened 2023-12-06
update zoe overview to match dapp-offer-up; add contract basics

### LINKED ISSUE #891: atomic swap example in zoe docs does not fit with smart wallet architecture
This interaction isn't feasible with the messages clients can currently send to the chain:

https://github.com/Agoric/documentation/blob/a0bba25401375a0702ae6eff0ae42f4831b27967/main/guides/zoe/README.md?plain=1#L262-L265

As part https://github.com/Agoric/documentation/issues/726 , let's use the game places contract in place of atomic swap.

See also:

 - https://github.com/Agoric/agoric-sdk/issues/8550

dckc:
Our getting started material has, to date, been largely disconnected from our Zoe guide. The getting started dapp was a fungible faucet, which had no conditional exchange of assets at all.

The card store app addresses that, but
 - it's not compatible with the smart wallet architecture
 - the Zoe docs ignore it and start with atomic swap

The thinking here is to use the simple game piece contract, which is somewhat like atomic swap but
  - shows how to mint tokens, avoiding "where do the issuers come from?"
  - demonstrably fits with smart wallet
  - is 66 lines

Aspects that this contract does not exhibit:

 - inspecting an invitation: the invitations are requested by the walletFactory and immediately used in a zoe.offer() call. They're never in your purse long enough for you to inspect them.

tasks:
- [x] [smart contract basics](https://dc-contract-basic.documentation-7tp.pages.dev/guides/zoe/contract-basics.html) - hello, testing, state, access control
- [x]  [zoe overview](https://dc-contract-basic.documentation-7tp.pages.dev/guides/zoe/) covers mint / sell token using game places contract
- [ ] file/link organization - where to put contract basics vs. the README

fixes #891, #436

refs:
 - https://github.com/Agoric/documentation/issues/726
 - [Sep 9 glitch experiment](https://github.com/Agoric/documentation/issues/582#issuecomment-1712707650)


main/guides/getting-started/README.md
```diff
@@ -6,13 +6,11 @@ Welcome! We'll start with a basic _dapp_; that is: a JavaScript contract and ass
...
-...
-}
-```
+::: details gameAssetContract.js listing
+
+<<< @/snippets/zoe/src/gameAssetContract.js#file
+
+:::
 
 As we begin, note that help is not far away:
```
LuqiPan, line 16: _note_

As mentioned before

> We support any long-term support (LTS) versions of Node.js.

on line 27 is no longer valid given incompatibility issues with node 20(https://github.com/Agoric/agoric-sdk/issues/8636)

PS: commenting on this link because GitHub wouldn't allow me to comment on line 27 directly

dckc: Well, the policy is still to support Node LTS. It's just that we've got a bug. The bug should probably be noted, along with a work-around.

Note that this part of this PR is sort of obsolete in favor of #894 .

dckc:
weird. I messed up something with git. hang on...

dckc:
ah. I had the wrong target branch. it should be right now.

snippets/zoe/contracts/test-zoe-hello.js
```diff
@@ -0,0 +1,38 @@
...
+/* global harden */
+
+// #region test-imports
+import '@endo/init';
+import { E } from '@endo/far';
+import test from 'ava';
+// #endregion test-imports
+import { start as startState } from '../src/02-state.js';
+import { start as startAccess } from '../src/03-access.js';
+// #region test1
```
Chris-Hibbert, line 11: I can see the source code, so I was able to understand the distinction between `startAccess` and `startState`, but it's not obvious from looking at the generated docs.  Does it make sense to add a comment in the tests where they're called?

Chris-Hibbert (approved):
This is good.

snippets/zoe/src/gameAssetContract.js
```diff
@@ -0,0 +1,80 @@
...
+    const tmp = mint.mintGains(want);
+    atomicRearrange(
+      zcf,
+      harden([
+        [playerSeat, gameSeat, give],
+        [tmp, playerSeat, want],
+      ]),
+    );
+
+    playerSeat.exit(true);
```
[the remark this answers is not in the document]
dckc: I don't know. It seems to work ok without doing that. Doesn't letting it go out of scope and get garbage collected suffice?

Checking [ZCFSeat docs](https://docs.agoric.com/reference/zoe-api/zcfseat.html)... all I see is "Note: You should not use aZCFSeat.exit() when exiting with an error."

I thought I had seen this pattern of using a temp seat with `mintGains()`, but [psm.js](https://github.com/Agoric/agoric-sdk/blob/master/packages/inter-protocol/src/psm/psm.js#L161) seems to use a long-lived seat.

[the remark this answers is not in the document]
Chris-Hibbert: we recently noticed (#8672) that Seats that are not exited can retain some uncollectible (GC-resistant) cycles. Calling `.exit()` when you know the seat is no longer in use is now a best practice.

snippets/zoe/contracts/test-zoe-hello.js
dckc: I re-factored it to use
```js
import * as state from '../src/02-state.js';
```

and then `state.start()`. I hope that's sufficiently clear.

PR #905, opened 2024-01-03
docs(getting-started): refine docker:logs section

dckc:
- note benign error messages
 - 10x2 log messages seems excessive; 3 pairs establishes the pattern
 - use block numbers closer to what they'll see


ivanlei (approved):
🚢

PR #906, opened 2024-01-03
docs(getting-started): note major steps in `start:contract`

dckc:
@kbennett2000 does this seem more useful or distracting?

 - [rendered](https://dc-start-contract-steps.documentation-7tp.pages.dev/guides/getting-started/#starting-the-dapp-smart-contract)

Not urgent / important. But it's a blurb I wrote at one point and I'd like you to take a look before I discard it.


kbennett2000 (approved):
I like this and I think it resolves some of the confusion from the proposed summary at the end of Getting Started (#926)

PR #911, opened 2024-01-08
unblurry version of `yarn start:ui` output

ivanlei:
# Description

The image of output from `yarn start:ui` was blurry. Updated with a less blurry version.


dckc (commented):
I abstain. Blurriness is one of the many reasons we shouldn't be using images for text at all.

https://www.w3.org/WAI/WCAG21/Understanding/images-of-text

kbennett2000 (approved):

PR #912, opened 2024-01-08
upgrade to Yarn 4

turadg:
https://github.com/Agoric/agoric-sdk/pull/8461 is [failing on integration with this repo](https://github.com/Agoric/agoric-sdk/actions/runs/7340663133/job/19987035933?pr=8461)

This PR upgrades Yarn from v1 to v4 and fixes issues that emerged with the stricter package resolution.

It also formats with Prettier to reduce code whitespace churn.

Review by commit is recommended


package.json
```diff
@@ -36,25 +36,26 @@
   },
   "homepage": "https://github.com/Agoric/documentation#readme",
   "dependencies": {
-    "@agoric/zoe": "beta",
-    "@endo/far": "^0.2.19",
-    "@endo/marshal": "^0.8.6",
-    "@endo/patterns": "^0.2.3",
-    "@endo/pass-style": "^0.1.6",
+    "@agoric/zoe": "^0.26.3-u13.0",
```
dckc, line 39: I double-checked that this is the version on chain.

package.json
```diff
@@ -36,25 +36,26 @@
...
+    "@endo/far": "^1.0.1",
+    "@endo/marshal": "^1.0.1",
```
dckc, line 41: Are these 1.0.x endo versions on purpose? The marshal version corresponding to zoe 0.26.3-u13.0 is 0.8.5.

https://github.com/Agoric/agoric-sdk/blob/agoric-upgrade-13/packages/zoe/package.json#L59

dckc (commented):
I'd like to hear more about endo versions before approving

package.json
turadg: I just used `yarn add` to grab the latest. I don't think it makes a difference here but if it does let's get that documented and tested.

dckc: Does https://github.com/Agoric/agoric-sdk/issues/8621 count as documentation? That's my concern.
It's not directly relevant here, but I hope the way we do stuff in this documentation repo will converge with the way we recommend doing stuff in dapps.

@mhofman @kriskowal can I get your thoughts here?

dckc (approved):
I'm content to deal with endo version stuff separately

https://github.com/Agoric/agoric-sdk/issues/8621 remains open

turadg:
This is failing `agoric install` because that command uses the `--link-folder` option and thus assumes Yarn version 1 for downstream projects.

```
$ yarn workspaces focus [--json] [--production] [-A,--all] ...
$ yarn workspaces foreach [--from #0] [-A,--all] [-R,--recursive] [-W,--worktree] [-v,--verbose] [-p,--parallel] [-i,--interlaced] [-j,--jobs #0] [-t,--topological] [--topological-dev] [--include #0] [--exclude #0] [--no-private] [--since] [-n,--dry-run] <commandName> ...
agoric: (Error#1)
Error#1: yarn install updates failed in .
  at yarnInstallEachWorktree (file:///home/runner/work/documentation/documentation/agoric-sdk/packages/agoric-cli/src/install.js:73:15)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async installMain (file:///home/runner/work/documentation/documentation/agoric-sdk/packages/agoric-cli/src/install.js:279:3)
  at async Command.parseAsync (/home/runner/work/documentation/documentation/agoric-sdk/node_modules/commander/lib/command.js:935:5)
  at async main (file:///home/runner/work/documentation/documentation/agoric-sdk/packages/agoric-cli/src/main.js:400:5)
```

This seems like an unnecessary coupling to me. I could make it check for the Yarn version but I wonder if @michaelfig has any better ideas.

dckc (commented):
Turadg notes it's failing in some way...

michaelfig:
> seems to be failing

It would be great to port `agoric install` to whatever `corepack` would use, but I don't have much experience outside of Yarn 1, so it would take some time to gather the requisite knowledge to do it.  I'd expect to see that put in priority on a project board somewhere.

snippets/zoe/contracts/test-callSpread.js
```diff
@@ -9,8 +9,8 @@ import '@agoric/zoe/exported.js';
 import buildManualTimer from '@agoric/zoe/tools/manualTimer.js';
 import { AmountMath } from '@agoric/ertp';
 
-import { setup } from '@agoric/zoe/test/unitTests/setupBasicMints.js';
```
turadg, line 12: newer versions of `@agoric/zoe` don't bundle `tests` path (thankfully)

turadg:
[Cloudflare build failure](https://github.com/Agoric/documentation/pull/912/checks?check_run_id=29087748253) looks due to [9a487fc](https://github.com/Agoric/documentation/pull/912/commits/9a487fc8f40234d29a30347be0bd775ed0c91dfd), not yarn. If other jobs pass then I'll rerun omitting that commit. cc @Jovonni @amessbee the agoric-sdk submodule is at https://github.com/Agoric/agoric-sdk/commit/03b6420649be77d010bc783efc5aaee5c7ed45e7 which is 4.5 months old.

kriskowal (approved):
What’s the plan for the copypasta from Zoe test infra? Do you have a separate change to export those in a future version?

package.json
kriskowal: Just seeing this now (after my approval and turadg’s merge) but I don’t have much to offer. My intuition is that users will expect installing the latest of everything to be in a consistent state and if it works, it works. If it doesn’t work, then there are surprising remediations, but that’s par for the _gestures broadly_ the emergent chaos of intractable priors.

kriskowal: I chatted yesterday with @mhofman @turadg and @michaelfig and I think the conclusion is that we’re going to start having the Agoric SDK release process add `agoric-` prefix tags to the corresponding `@endo/*` releases so that one can `yarn install @endo/x@tag @agoric/y@tag` and expect a coæval snapshot.

turadg:
> Do you have a separate change to export those in a future version?

No plan. Test support code seems very copyable to me. If it provides what the tests need, it works. If the tests' needs change they can go back upstream to see if there's something better to copy.

If we want to abstract that into a supported set of exports, I think it needs a new package: https://github.com/Agoric/agoric-sdk/issues/7392

PR #916, opened 2024-01-10
chore: Miscellaneous improvements to ERTP API pages

gibson042:
Replacement for #741

* Alphabetize ERTP Data Types sections
* Use consistent phrasing for "Both **Amount** arguments must have the same **Brand**" and "an error is thrown"
* Use consistent `// <Capitalized word> …` code comment style
* Format code with `…`
* Remove implications that AmountMath methods ever return their actual arguments (rather than equivalent but distinct values)
* Remove empty lines at end of file

### DIFF
--- a/main/reference/ertp-api/amount-math.md
+++ b/main/reference/ertp-api/amount-math.md
@@ -25,8 +25,7 @@ not equal, an error is thrown and no changes are made.
 Creates an **Amount** from a given **Brand** and **AmountValue**.
 
 ```js
-// amount837 = { brand: quatloosBrand, value: 837n }
-const amount837 = AmountMath.make(quatloosBrand, 837n);
+const bid = AmountMath.make(quatloosBrand, 300n);
 ```
 
 ## AmountMath.coerce(brand, allegedAmount)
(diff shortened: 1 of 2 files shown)


ivanlei (approved):
🚢

main/reference/ertp-api/amount-math.md
```diff
@@ -25,22 +25,19 @@ not equal, an error is thrown and no changes are made.
...
 ```
 
 ## AmountMath.coerce(brand, allegedAmount)
 - **brand**: **[Brand](./brand.md)**
 - **allegedAmount**: **[Amount](./ertp-data-types.md#amount)**
 - Returns: **Amount**
 
-Makes sure an **Amount** is for the specified *brand* and if so, returns it.
-If the **Amount** is for an invalid **Brand**, this method throws an exception.
+Verifies that an **Amount** is for the specified *brand* and returns a new equivalent **Amount**.
```
erights, line 36:
```suggestion
Verifies that an **Amount** is for the specified *brand* and returns an equivalent **Amount**.
```

The PR overcorrects the original. The original mistakenly implied it is the same amount object. The PR mistakenly states it is guaranteed to be a fresh amount object, even if the original would have been fine. I think the API spec should only guarantee equivalence, without stating whether or not it is a new JS allocation.

main/reference/ertp-api/amount-math.md
```diff
@@ -183,21 +176,21 @@ AmountMath.isEqual(empty, quatloos10);
...
 If the optional *brand* argument doesn't match the **Amount**s' **Brand**, an error is thrown.
 
 For fungible **Amounts** this means adding their **[AmountValues](./ertp-data-types.md#amountvalue)**. For non-fungible
 **Amounts**, it usually means including all of the elements from *leftAmount*
 and *rightAmount*.
 
-If either *leftAmount* or *rightAmount* is empty, this method returns the non-empty 
-**Amount** argument. If both are empty, this method returns an empty **Amount**.
+If one of *leftAmount* or *rightAmount* is empty, this method returns an **Amount** equivalent to the other.
+If both are empty, this method returns a new empty **Amount**.
```
erights, line 189:
```suggestion
If both are empty, this method returns an empty **Amount**.
```

Likewise. The API spec should not say whether or not the empty amount is newly allocated or not.

dckc:
hoping the ci bots are happier with #959

PR #920, opened 2024-01-18
Updated Getting Started docs

kbennett2000:
- Changed title to "Getting Started"
- Added section to formally introduce the Dapp
- Embedded Video Walkthrough

### DIFF
--- a/main/guides/getting-started/README.md
+++ b/main/guides/getting-started/README.md
@@ -1,7 +1,10 @@
-# Your First Agoric Dapp
+# Getting Started
 In these steps you will be getting your first Agoric dapp up and running!
 
 ![Your first Agoric dapp](./assets/new_002_small2.png)
+
+## Your First Agoric Dapp - Offer Up
+The Offer Up Dapp allows users to view all listed items up for sale in a marketplace. The Dapp shows 3 items by default (maps, potions, and scrolls). As a user you are able to choose up to 3 total items (in any combination). You can then create an "offer" for an amount of at least 0.25 IST. Once you sign the transaction, you will see the 3 items in your wallet and your wallet amount deducted by the Offer amount.
 
 ## How to Get Help
 Before getting started, there are some resources you might want to keep handy in case you get stuck, have questions, or are curious about any of the components. Getting in contact with us is easy! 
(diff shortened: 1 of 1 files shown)


main/guides/getting-started/README.md
```diff
@@ -1,8 +1,11 @@
-# Your First Agoric Dapp
+# Getting Started
 In these steps you will be getting your first Agoric dapp up and running!
 
 ![Your first Agoric dapp](./assets/new_002_small2.png)
 
+## Your First Agoric Dapp - Offer Up
+The Offer Up Dapp allows users to view all listed items up for sale in a marketplace. The Dapp shows 3 items by default (maps, potions, and scrolls). As a user you are able to choose up to 3 total items (in any combination). You can then create an "offer" for an amount of at least 0.25 IST. Once you sign the transaction, you will see the 3 items in your wallet and your wallet amount deducted by the Offer amount.
```
dckc, line 7: hm. quotes around offer... I suggest dropping them.

is that to call attention to the introduction of the term? I tend to use italics. I'd use `<dfn>` if the tools supported it better. Since we're not defining it here, we should have a forward reference to where we do defined it.

But I'd rather just leave it not marked up in this case. The ordinary dictionary meaning is good enough here.

kbennett2000: Quotes are gone! :)

main/guides/getting-started/README.md
```diff
@@ -11,6 +14,12 @@ Before getting started, there are some resources you might want to keep handy in
 - Send us a message on [X](https://twitter.com/agoric)
 - Send an Email to [Developer Relations](mailto://devrel@agoric.com)
 
+::: tip Video Walkthrough
+
+As you're going through this tutorial it may be helpful to watch this video walkthrough, showing you all the steps and the desired outcome of each.
+
+<iframe width="560" height="315" src="https://youtu.be/huTI9_Gq2ow" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
```
dckc, line 21: this isn't working for me in the deploy preview: "www.youtube.com refused to connect"

image

kbennett2000: Updated the YouTube embed URL, should be loading now...

dckc (approved):
good stuff!

PR #965, opened 2024-02-06
build: update dependencies to match dapp-offer-up (agoric-upgrade-13)

dckc:
fixes #843

includes the resolutions work-around for
 - https://github.com/Agoric/agoric-sdk/issues/8621

Issue #952, opened 2024-02-01
4daydapp - Document - How does Governance work in Agoric?


sufyaankhan:
**User Story:** As a Developer I should be able to read the docs and understand how Governance works in Agoric systems. (use Dapp-Agoric-Basics as context)

dckc:
related work: #749 esp. https://github.com/Agoric/documentation/issues/749#issuecomment-1502148745

dckc:
I asked around for an example / test to work from and got

https://github.com/Agoric/agoric-sdk/blob/master/packages/inter-protocol/test/auction/test-auctionContract.js#L112

Not all the testing tools it uses are available outside the SDK, so some adaptation is needed.

https://github.com/Agoric/agoric-sdk/blob/a3f0fc5d2cd48fa398491237b683d091b1003309/packages/inter-protocol/test/supports.js#L96-L111

dckc:
closed in #999

Issue #749, opened 2023-01-09
Document Governance


Tyrosine22:

Chris-Hibbert:
We need to write guidelines for contract authors on how to use the governance framework, and for contract reviewers on how to verify that a contract is following the process correctly, and where to look for violations. There's some of the latter in the [Attacker's Guide](https://github.com/Agoric/agoric-sdk/blob/master/packages/governance/docs/AttackGuide.md), but it should be made more systematic.

Some links:
 * [contractGovernor](https://github.com/Agoric/agoric-sdk/blob/5fcfe1a89c3db7af00143385d10d60910694e5c0/packages/governance/src/contractGovernor.js#L68)
 * [contractHelper](https://github.com/Agoric/agoric-sdk/blob/5fcfe1a89c3db7af00143385d10d60910694e5c0/packages/governance/src/contractHelper.js#L205)
 * search for `CRUCIAL` in .../governance/.
 *

dckc:
closed in #999

PR #968, opened 2024-02-08
contract upgrade

### LINKED ISSUE #750: Document Durability, Upgrade
**User Story:** As a Developer I should be able to read the docs and understand how to make contract durable and upgradable (Using the dapp-agoric-basics as context)

dckc:
goal:
fixes #750

rendered:
 - [Contract Upgrade](https://dc-durable-upgrade.documentation-7tp.pages.dev/guides/zoe/contract-upgrade.html)

Audience is dapp/contract devs, so the emphasis is on how to achieve it, not the theory of how the infrastructure works. @LuqiPan , @kbennett2000  I'm particularly interested to know if this brief account covers the subject well enough for you to get a working knowledge of it.

It lacks detail on interface guards. I propose to merge this as is and add those later. Opinions?

I started trying to make sure all the code snippets are really tested, but `makeZoeKitForTest` doesn't support upgrade, and it's not clear to me how to do bootstrap tests with `makeSwingsetTestKit` from outside agoric-sdk; I have already spent probably too much time on tooling in #965.

## Testing considerations

Examples are not well tested due to the difficulty of landing...

 - #965

### DIFF
--- a/main/guides/zoe/contract-upgrade.md
+++ b/main/guides/zoe/contract-upgrade.md
@@ -1,0 +1,176 @@
+# Contract Upgrade
+
+The result of starting a contract includes the right to upgrade the contract. A call to [E(zoe).install(...)](/reference/zoe-api/zoe.md#e-zoe-startinstance-installation-issuerkeywordrecord-terms-privateargs) returns a record of several objects that represent different levels of access.
+The `publicFacet` and `creatorFacet` are defined by the contract.
+The `adminFacet` is defined by Zoe and includes methods to upgrade the contract.
+
+::: tip Upgrade Governance
+
+Governance of the right to upgrade is a complex topic that we cover only briefly here.
+
+- When [BLD staker governance](https://community.agoric.com/t/about-the-governance-category/15) makes a decision to start a contract using [swingset.CoreEval](../coreeval/),
+  to date, the `adminFacet` is stored in the bootstrap vat, allowing
+  the BLD stakers to upgrade such a contract in a later `swingset.CoreEval`.
+- The `adminFacet` reference can be discarded, so that noone can upgrade
+  the contract from within the JavaScript VM. (BLD staker governace
+  could, in theory, change the VM itself.)
+- The `adminFacet` can be managed using the [@agoric/governance](https://github.com/Agoric/agoric-sdk/tree/master/packages/governance#readme) framework; for example, using the `committee.js` contract.
+
+:::
+
+## Upgrading a Contract
+
+Upgrading a contract instance means re-starting the contract using a different [code bundle](./#bundling-a-contract). Suppose we start a contract as usual, using
+the bundle ID of a bundle we already sent to the chain:
+
+```js
+const bundleID = 'b1-1234abcd...';
+const installation = await E(zoe).installBundleID(bundleID);
+const { instance, ... facets } = await E(zoe).startInstance(installation, ...);
+
+// ... use facets.publicFacet, instance etc. as usual
+```
+
+If we have the `adminFacet` and the bundle ID of a new version,
+we can use the `upgradeContract` method to upgrade the contract instance:
+
+```js
+const v2BundleId = 'b1-feed1234...`; // hash of bundle with new feature
+const { incarnationNumber } = await E(facets.adminFacet).upgradeContract(v2BundleId);
+```
+
+The `incarnationNumber` is 1 after the 1st upgrade, 2 after the 2nd, and so on.
+
+::: details re-using the same bundle
+
+Note that a "null upgrade" that re-uses the original bundle is valid, and a legitimate approach to deleting accumulated heap state.
+
+See also `E(adminFacet).restartContract()`.
+
+:::
+
+## Upgradable Contracts
+
+There are a few requirements for the contract that differ from non-upgradable contracts:
+
+1. [Upgradable Declaration](#upgradable-declaration)
+2. [Durability](#durability)
+3. [Kinds](#kinds)
+4. [Crank](#crank)
+
+### Upgradable Declaration
+
+The new code bundle declares that it supports upgrade by exporting a `prepare` function in place of `start`.
+
+<<< @/snippets/zoe/src/02b-state-durable.js#export-prepare
+
+### Durability
+
+The 3rd argument, `baggage`, of the `prepare` function is a `MapStore`
+that provides a way to preserve state and behavior of objects
+between incarnations in a way that preserves identity of objects
+as seen from other vats:
+
+```js
+let rooms;
+if (!baggage.has('rooms')) {
+  // initial incarnation: create the object
+  rooms = makeScalarBigMapStore('rooms', { durable: true });
+  baggage.init('rooms', rooms);
+} else {
+  // subsequent incarnation: use the object from the initial incarnation
+  rooms = baggage.get('rooms');
+}
+```
+
+The `provide` function supports a concise idiom for this find-or-create pattern:
+
+```js
+import { provide } from '@agoric/vat-data';
+
+const rooms = provide(baggage, 'rooms', () =>
+  makeScalarBigMapStore('rooms', { durable: true }),
+);
+```
+
+The `zone` API is a convenient way to manage durability. Its store methods integrate the `provide` pattern:
+
+::: details import { makeDurableZone } ...
+
+<<< @/snippets/zoe/src/02b-state-durable.js#import-zone
+
+:::
+
+<<< @/snippets/zoe/src/02b-state-durable.js#zone1
+
+::: details What happens if we don't use baggage?
+
+When the contract instance is restarted, its [vat](../js-programming/#vats-the-unit-of-synchrony) gets a fresh heap, so [ordinary heap state](./contract-basics.md#state) does not survive upgrade. This implementation does not persist the rooms nor their counts between incarnations:
+
+<<< @/snippets/zoe/src/02-state.js#heap-state{2}
+
+:::
+
+### Kinds
+
+Use `zone.exoClass()` to define state and methods of kinds of durable objects such as `Room`:
+
+<<< @/snippets/zoe/src/02b-state-durable.js#exoclass
+
+Defining `publicFacet` as a singleton `exo` allows clients to
+continue to use it after an upgrade:
+
+<<< @/snippets/zoe/src/02b-state-durable.js#exo
+
+Now we have all the parts of an upgradable contract.
+
+::: details full contract listing
+
+<<< @/snippets/zoe/src/02b-state-durable.js#contract
+
+:::
+
+We can then upgrade it to have another method:
+
+```js
+  const makeRoom = zone.exoClass('Room', RoomI, (id) => ({ id, value: 0 }), {
+    ...
+    clear(delta) {
+      this.state.value = 0;
+    },
+  });
+```
+
+The interface guard also needs updating.
+<small>_See [@endo/patterns](https://endojs.github.io/endo/modules/_endo_patterns.html) for more on interface guards._</small>
+
+```js
+const RoomI = M.interface('Room', {
+  ...
+  clear: M.call().returns(),
+});
+```
+
+::: tip Notes
+
+- Once the state is defined by the `init` function (3rd arg), properties cannot be added or removed.
+- Values of state properties must be serializable.
+- Values of state properties are hardened on assignment.
+- You can replace the value of a state property (e.g. `state.zot = [...state.zot, 'last']`), and you can update stores (`state.players.set(1, player1)`), but you cannot do things like `state.zot.push('last')`, and if jot is part of state (`state.jot = { x: 1 };`), then you can't do `state.jot.x = 2;`
+- The tag (1st arg) is used to form a key in `baggage`, so take care to avoid collisions. `zone.subZone()` may be used to partition namespaces.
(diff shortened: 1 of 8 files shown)


main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+    },
+  });
+```
+
+::: tip Notes
+
+- Once the state is defined by the `init` function (3rd arg), properties cannot be added or removed.
+- Values of state properties must be serializable.
+- Values of state properties are hardened on assignment.
+- You can replace the value of a state property (e.g. `state.zot = [...state.zot, 'last']`), but you cannot mutate it (`state.zot.push('last')`).
```
Chris-Hibbert, line 151: This didn't sound right to me. I asked in #dev-kernel, and the deal is that things in state are hardened, but that doesn't make them all immutable. One. example is mapStores You can add keys or modify values in a mapStore without reassigning the store.

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+
+:::
+
+## Crank
+
+Define all exo classes/kits before any incoming method calls from other vats -- in the first "crank".
+
+::: tip Note
+
+- For more on crank constraints, see [Virtual and Durable Objects](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/docs/virtual-objects.md#virtual-and-durable-objects) in [SwingSet docs](https://github.com/Agoric/agoric-sdk/tree/master/packages/SwingSet/docs)
```
Chris-Hibbert, line 166: Great link, and it's fine to leave this detail out of the body.

Chris-Hibbert (changes requested):

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+### Durability
+
+The 3rd argument, `baggage`, of the `prepare` function is a `MapStore`
+that provides a way to preserve state and behavior of objects
+between incarnations in a way that preserves identity of objects
+as seen from other vats:
+
+```js
+let publicFacet;
+if (!baggage.has('publicFacet')) {
```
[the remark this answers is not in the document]
dckc: > How would you feel about changing this to something like openBids or treasureChest?

I wonder how to do that without substantially complicating the complete example contract here.

> publicFacet and creatorFacet retain their object identity across incarnations

only if the contract arranges for that to be the case, yes? or did I miss something?
I know the swingset assigns a fixed identity for the root object of each vat, but I'm not aware of anything like that for publicFacet.

> A durable contract is required to rebind the behavior of any Kinds that were defined in previous incarnations.

That's (supposed to be) covered under **Kinds** below.

main/guides/zoe/contract-upgrade.md
dckc: good point about stores.

(the zot stuff was taken directly from swingset docs, so it's wrong there too, I guess)

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+
+Governance of the right to upgrade is a complex topic that we cover only briefly here.
+
+- When BLD staker governance starts a contract using `swingset.CoreEval`,
+  to date, the `adminFacet` is stored in the bootstrap vat, allowing
+  the BLD stakers to upgrade such a contract in a later `swingset.CoreEval`.
+- The `adminFacet` reference can be discarded, so that noone can upgrade
+  the contract from within the JavaScript VM. (BLD staker gonvernace
+  could, in theory, change the VM itself.)
+- The `adminFacet` can be shared with a governance contract such as `committee.js`.
```
0xpatrickdev, line 17: > `committee.js`

a link to the @agoric/governance readme might be nice, or to one of these in the future:
- #749
- #952

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+
+::: tip Upgrade Governance
+
+Governance of the right to upgrade is a complex topic that we cover only briefly here.
+
+- When BLD staker governance starts a contract using `swingset.CoreEval`,
+  to date, the `adminFacet` is stored in the bootstrap vat, allowing
+  the BLD stakers to upgrade such a contract in a later `swingset.CoreEval`.
+- The `adminFacet` reference can be discarded, so that noone can upgrade
+  the contract from within the JavaScript VM. (BLD staker gonvernace
```
0xpatrickdev, line 15:
```suggestion
  the contract from within the JavaScript VM. (BLD staker governance
```

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Crank
+
+Define all exo classes/kits before any incoming method calls from other vats -- in the first "crank".
+
+::: tip Note
+
+- For more on crank constraints, see [Virtual and Durable Objects](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/docs/virtual-objects.md#virtual-and-durable-objects) in [SwingSet docs](https://github.com/Agoric/agoric-sdk/tree/master/packages/SwingSet/docs)
+
+:::
```
0xpatrickdev, line 168: This is great, I learned a few things! 

Two suggestions maybe worth implementing - 

1. Do we want to discuss `durability` vs `upgradability` (if not slated for a different section)?. Towards the beginning it may be worthwhile to mention why you may want to make parts of your contract durable (but not necessarily upgradable)
2. May be worth adding upgradability tips/best practices - e.g. backwards compatibility, optional extra arguments, making sure the prepare script doesn't reset counts / state (unless intended).

dckc: > why you may want to make parts of your contract durable (but not necessarily upgradable)

Is there any reason besides upgrade for durability? I can't think of any.

> backwards compatibility, optional extra arguments

Got any specific suggestions? I'm struggling to find a concise statement of best practices. The semver project punted on a definition, for example:

 - https://github.com/semver/semver/issues/256

https://www.linkedin.com/advice/0/what-best-practices-ensuring-backward-compatibility-i44we has the most likely looking title, but the content doesn't look that good

> making sure the prepare script doesn't reset counts / state

that seems to fall out naturally from the zone/exo API. It doesn't seem useful to say "don't go out of your way to shoot yourself in the foot."

main/guides/zoe/contract-upgrade.md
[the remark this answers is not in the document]
Chris-Hibbert: > only if the contract arranges for that to be the case, yes? or did I miss something?
> I know the swingset assigns a fixed identity for the root object of each vat, but I'm not aware of anything like that for publicFacet.

Zoe saves the original `publicFacet` on startup, and will continue handing that out in response to `zoe.getPublicFacet()`, even if the contract rebinds it internally. The creator of the contract gets `publicFacet` and `creatorFacet` in the response to `startInstance` and doesn't get new values on upgrade. 

The guarantees about upgraded objects are that clients who have references to pre-existing objects will retain them, and the behavior will be updated if they were durable. (If non-durable, they go away on upgrade.) If the `publicFacet` was durable, then it was a facet (or singleton) with durable identity. New behavior has to be assigned to that Kind on upgrade. 

I can't think of anything useful that would come from storing `publicFacet` in baggage. Is there something I'm missing?

main/guides/zoe/contract-upgrade.md
Chris-Hibbert: @warner agreed with that and volunteered to extend the example as above.

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,168 @@
...
+```js
+const v2BundleId = 'b1-feed1234...`; // hash of bundle with new feature
+const { incarnationNumber } = await E(facets.adminFacet).upgradeContract(v2BundleId);
+```
+
+The `incarnationNumber` is 1 after the 1st upgrade, 2 after the 2nd, and so on.
+
+::: details re-using the same bundle
+
+Note that a "null upgrade" that re-uses the original bundle is valid, and a legitimate approach to deleting accumulated state.
```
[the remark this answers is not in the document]
dckc: we call it heap state later in the page; I think I'll go with that

main/guides/zoe/contract-upgrade.md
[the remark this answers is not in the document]
dckc: ah. I get it now. thanks.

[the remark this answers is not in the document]
dckc: Does the new rooms example work for you?

main/guides/zoe/contract-upgrade.md
0xpatrickdev: > Is there any reason besides upgrade for durability? I can't think of any.

I've been under the impression "making my contract survive a chain upgrade" and "make my contract upgradable (new version, functionality changes)" are two separate things, but seem to be mistaken

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,186 @@
...
+The 3rd argument, `baggage`, of the `prepare` function is a `MapStore`
+that provides a way to preserve state and behavior of objects
+between incarnations in a way that preserves identity of objects
+as seen from other vats:
+
+```js
+let rooms;
+if (!baggage.has('rooms')) {
+  // initial incarnation: create the object
+  rooms = makeScalarBigMapStore('rooms', { durable: true });
```
erights, line 76: I like that you use zones to make durable exos below. But then you should also use zones to make durable stores. Both for consistency, and because it's a much better API --- without "Big" or `{ durable: true }`

main/guides/zoe/contract-upgrade.md
dckc: Contract survive chain upgrades without being durable, but it _is_ useful to be able to restart a contract to clear out the heap. That's noted in the **re-using the same bundle** details box.

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,186 @@
...
+
+<<< @/snippets/zoe/src/02b-state-durable.js#import-zone
+
+:::
+
+<<< @/snippets/zoe/src/02b-state-durable.js#zone1
+
+::: details What happens if we don't use baggage?
+
+When the contract instance is restarted, it gets a fresh [heap](../js-programming/#vats-the-unit-of-synchrony), so [ordinary heap state](./contract-basics.html#state) does not survive upgrade. This implementation does not persist the effect of `E(publicFacet).set(2)` from the first incarnation:
```
Chris-Hibbert: I think this should be converted to a `rooms` example.  Similarly in the following paragraph.

main/guides/zoe/contract-upgrade.md
dckc: That's in there, right? This section is showing the fully elaborated code that's easier to understand the first time you see it. But then it's replaced by more and more concise idioms.

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,186 @@
...
+  clear: M.call().returns(),
+});
+```
+
+::: tip Notes
+
+- Once the state is defined by the `init` function (3rd arg), properties cannot be added or removed.
+- Values of state properties must be serializable.
+- Values of state properties are hardened on assignment.
+- You can replace the value of a state property (e.g. `state.zot = [...state.zot, 'last']`), and you can update stores (`state.players.set(1, player1)`), but you cannot do things like `state.zot.push('last')` nor `state.jot = { x: 1 }; state.jot.x = 2`.
```
Chris-Hibbert, line 163: It's unfortunate that `state.jot = { x: 1 }; state.jot.x = 2` broke across a line for me. What do you think of making it a multiline expression to make it more apparent?

Chris-Hibbert (commented):
> Does the new rooms example work for you?

Yep.  I like.

main/guides/zoe/contract-upgrade.md
erights: Sorry, I was not reading carefully. Nevermind. Resolving.

main/guides/zoe/contract-upgrade.md
dckc: added a line-break

main/guides/zoe/contract-upgrade.md
```diff
@@ -0,0 +1,177 @@
...
+});
+```
+
+::: tip Notes
+
+- Once the state is defined by the `init` function (3rd arg), properties cannot be added or removed.
+- Values of state properties must be serializable.
+- Values of state properties are hardened on assignment.
+- You can replace the value of a state property (e.g. `state.zot = [...state.zot, 'last']`), and you can update stores (`state.players.set(1, player1)`), but you cannot do things like `state.zot.push('last')` nor `state.jot = { x: 1 };`  
+  `state.jot.x = 2;`.
```
Chris-Hibbert, line 160: Sorry, that made it worse, making my mis-reading even more likely. My problem was that with an unfortunate line break, I saw "you cannot do things like ... nor state.jot = { x: 1 };" and it took me a while to see that the statement following the line break was the thing that wouldn't work.

```suggestion
- You can replace the value of a state property (e.g. `state.zot = [...state.zot, 'last']`), and you can update stores (`state.players.set(1, player1)`), but you cannot do things like `state.zot.push('last')`, and if jot is part of state (`state.jot = { x: 1 };`), then you can't do `state.jot.x = 2;`
```

PR #975, opened 2024-02-14
Added definition for `bag`

kbennett2000:
- Added definition for `bag` and described differences between bags and sets

### DIFF
--- a/main/glossary/README.md
+++ b/main/glossary/README.md
@@ -97,6 +97,10 @@ For more information, see the [ERTP documentation's AmountValue section](/guides
 [Purses](#purse) and [payments](#payment) are AssetHolders. These are objects that contain
 digital assets in the quantity specified by an [amount](#amount).
 
+## Bag
+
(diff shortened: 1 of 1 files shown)


main/glossary/README.md
```diff
@@ -97,6 +97,10 @@ For more information, see the [ERTP documentation's AmountValue section](/guides
 [Purses](#purse) and [payments](#payment) are AssetHolders. These are objects that contain
 digital assets in the quantity specified by an [amount](#amount).
 
+## Bag
```
dckc, line 100: What I had in mind was not defining the general concept of Bag, but making a link between discussion of **semi-fungible** (e.g. in the glossary entry) and the discussion of `AssetKind.COPY_BAG`. Looks like the latter is in /reference/ertp-api/ertp-data-types.html#assetkind

Ugh... It looks like "AmountValues are arrays of objects." got copied from the **AssetKind.COPY_SET** bullet to the **AssetKind.COPY_BAG** bullet, where it doesn't apply.

There's some good docs in the source code:

https://github.com/Agoric/agoric-sdk/blob/23a970d49d44d3ec3159a5f402efb81bb3aff269/packages/ERTP/src/typeGuards.js#L48-L65

Here's hoping for time to talk about how to meld that with the docs site.

dckc (commented):
This is different from what I had in mind; when looking for suggestions, I found even more existing problems. I'm not sure how many of them to bite off just now.

PR #996, opened 2024-03-14
feat: home page redesign

0xpatrickdev:
refs: #998

## Summary 

- use default vitepress theme (`layout: home`), styled for agoric

## To Do

- [ ] copy/content and links
- [ ] unfurl
    - We might want to include `og:image` and `twitter:card` meta tags out as part of this PR so we get a nice unfurl: https://github.com/Agoric/documentation/issues/997
- [ ] footer links

## Screenshot

image

PR #1008, opened 2024-03-19
homepage suggestions

dckc:
suggestions for #996

 1. refine 1st row of feature boxes
 2. replace repeated row with support contacts, based on an internal sketch


0xpatrickdev (approved):
👏

PR #1012, opened 2024-03-20
docs: consolidate proposal shapes; shorten zoe overview

dckc:
fixes #951
closes #739
closes #880

Turns out [Proposal Shape](https://docs.agoric.com/reference/zoe-api/zoe.html#proposal-shape) was documented already, under `E(zoe).offer(...)`. And the endo reference docs cover Pattern Matchers pretty well.

It was also mentioned under [zcf.makeInvitation](https://docs.agoric.com/reference/zoe-api/zoe-contract-facet.html#zcf-makeinvitation-offerhandler-description-customdetails-proposalshape). Since this is the more directly relevant API, I moved the material here and cross-referenced it.

And I copied the `Matchers` interface that gives all the methods of `M`.

My editor does markdown lint on save; I hope it's not too distracting that I mixed in work on #945 here. I did put it in separate commits.

Issue #1011, opened 2024-03-19
checking links fails to account for section anchors / hashes


dckc:
another regression from #971?

 - https://github.com/vuejs/vitepress/issues/646

cc @0xpatrickdev

PR #1013, opened 2024-03-20
split offer-up walkthru out of zoe overview

dckc:
- split the offer-up walkthru out of the zoe overview
 - fix the "Contract Basics" navigation.

I started this refactor in preparation for work on proposal shapes, but then discovered we already have reasonable reference material on proposal shapes, so I just tidied it up a bit: #1012

This seems worth keeping, though.


dckc:
@kriskowal @michaelfig another one like https://github.com/Agoric/documentation/pull/1012#issuecomment-2008614853

ref https://github.com/Agoric/agoric-sdk/issues/8536 / https://github.com/Agoric/agoric-sdk/pull/9113

Chris-Hibbert (approved):

dckc:
p.s. I suspect this breaks some links to heading anchors. Our tools to check have regressed

 - https://github.com/Agoric/documentation/issues/1011

PR #1019, opened 2024-03-21
feat: add unfurl image + metadata

### LINKED ISSUE #997: unfurl url should have a preview image
We do not get the best unfurl preview for the https://docs.agoric.com url:

image

Fixing this, is straight forward. Seems we need to 

 - [ ] Determine what image we want to use - vitepress uses https://vitepress.dev/vitepress-og.jpg
 - [ ] Add relevant meta tags. It seems to be at least -
    - `<meta property="og:image" content="/images/agoric-og.png">`
    - `<meta property="twitter:card" content="/images/agoric-og.png">`

0xpatrickdev:
- closes #997


main/.vitepress/config.mjs
```diff
@@ -2,18 +2,31 @@ import { defineConfig } from 'vitepress';
 import { nav } from './themeConfig/nav.js';
 import { rewrites } from './themeConfig/rewrites.js';
 
+const SITE_TITLE = 'Agoric Documentation';
+const SITE_DESCRIPTION =
+  'The blockchain framework tailored for JavaScript developers.';
```
dckc, line 7: I don't have any objections to this copy, but I don't think I'm a good person to approve it.
@sufyaankhan ?

dckc (commented):
defer to somebody else... @sufyaankhan ?

PR #1018, opened 2024-03-21
doc: learn / build navigation

dckc:
@sufyaankhan ,

@kbennett2000 and I played around with [the nav sketch in miro](https://miro.com/app/board/uXjVNpwnZ-I=/) . This is as far as we got.

goal:
fixes #973

stacked on a rebase of
 - #996

PR #1034, opened 2024-03-26
fix: restore Hardened JavaScript etc. to nav

dckc:
I don't think we meant to take this out of the left nav bar (and "next page" links) in #1018 .


LuqiPan (approved):
LGTM

PR #1055, opened 2024-04-03
ci: added cf-log-fetcher github action to fetch deployment logs

jkhokhar1:


.github/workflows/cf-logs-fetcher.yml
```diff
@@ -0,0 +1,23 @@
...
+  fetch_comment_log:
+    if: ${{ (github.event.issue.pull_request) && (github.event.comment.user.login == 'cloudflare-pages[bot]' ) }}
+    runs-on: [ubuntu-latest]
+    steps:
+    
+      - name: Fetch & Print The Deployment Logs
+        uses: agoric-labs/cf-logs-fetcher@v1
+        with: 
+          cf_account_id: "0c4635effffcd7f36d1b9f0425a4367a"
+          cf_project: "documentation"
```
rabi-siddique, line 15: What is `cf_account_id`? Is it some sensitive data? Should we add it in secrets?

.github/workflows/cf-logs-fetcher.yml
```diff
@@ -0,0 +1,23 @@
...
+          cf_token: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            Cloudflare deployment logs are available [here](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
+          comment_tag: cflogs
```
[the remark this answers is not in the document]
jkhokhar1: This is not linking to actual logs, it point the action run history, where we can view the logs

.github/workflows/cf-logs-fetcher.yml
jkhokhar1: account id is safe to keep - since it is secret protected.

rabi-siddique (approved):

PR #1057, opened 2024-04-04
Update index.md

kbennett2000:
- Changed node version from v18.16.0 to LTS
- Removed installing nvm via curl, it was redundant


dckc (approved):
looks good.

In the future, please replace the default "Update index.md" commit message with a summary of the change.

https://github.com/Agoric/agoric-sdk/wiki/Conventional-Commits

Issue #922, opened 2024-01-18
How to build a Client UI - document ui-kit


dckc:
There is a gap in our documentation on docs.agoric.com in that we don't sufficiently explain/walk-through how a developer is supposed to a) build a client UI and b) connect it to a smart contract they've build on the Agoric chain.

https://docs.agoric.com/guides/getting-started/contract-rpc.html exists but it's more conceptual and background context - which while important to the reader/developer isn't a step-by-step guide on how to actually build the client UI.

Dapp-offer-up is a functional example contract and client UI that our docs reference in 'Getting Started' section (https://docs.agoric.com/guides/getting-started/) of our docs and in the 'Smart Contract Basics' section (https://docs.agoric.com/guides/zoe/contract-walkthru.html). It stands to reason that it the 'how to build a client UI' guide uses same smart contract for continuity and consistency. It will be easier for the developer/reader to follow along.

This ticket tracks the effort to document the step-by-step process through which a developer builds client UI (using dapp offer up as an example) and connect it to the offer up smart contract for full end-to-end functionality.

Requirements:

- Must reference dapp-offer-up contract and tie in with 'Getting Started' and 'Smart Contract Basics' such that it reads in a cohesive and easy-to-follow manner
- Must include instructions around how to query an RPC node to get data from contract
- Must reference utilization of UI components to simplify the UI building experience for developers
- Must use UI Kit (if UI kit explainer is needed, this needs to be incorporated)
- Must be written in step-by-step sequential format that is easy to follow

Acceptance Criteria:

- To ensure the how-to guide achieves our goal of being easy-to-follow and digest for developers - we will request review of @kbennett2000 (and maybe others) to follow along to elicit feedback. Iterations may be needed in order for us learn and improve on this.
- This will need to succeed before pushing to production docs.agoric.com

**User Story:** As a Developer I should be able to read the docs and get 2 things:
1. Understanding of how to build a Client UI and make it interact with my deployed smart contract
2. Get a repo/bundle of a sample Client UI that I could use to create my own Client

**ui-kit** and the basics of writing a dapp must be fully documented.

----
earlier idea:

 - reorg "Smart Wallet Dapp Architecture"?
```[tasklist]
### Tasks
- [ ] Sam to propose how-to page structure
```

dckc:
the smart wallet architecture page has a bunch of details that are kinda abstracted by ui-kit

dckc:
#1035 is not merged.

otoole-brendan:
[...]

PR #1072, opened 2024-05-01
UI tutorial

samsiegart:
fixes https://github.com/Agoric/documentation/issues/922

Link to tutorial page: https://ui-tutorial.documentation-7tp.pages.dev/guides/ui-tutorial/#ui-tutorial

Reference repo https://github.com/agoric-labs/ui-tutorial

In general I need to add some reference images, better linking to other docs and repos, and better layout/indexing of pages (in addition to prose and other content from code review/feedback)


otoole-brendan:
This is awesome @samsiegart! I might have missed it but I didn't see it reference the current 'Building Client Dapps' (https://docs.agoric.com/guides/getting-started/contract-rpc.html) page. Should it?

samsiegart:
> This is awesome @samsiegart! I might have missed it but I didn't see it reference the current 'Building Client Dapps' (https://docs.agoric.com/guides/getting-started/contract-rpc.html) page. Should it?

Yea thanks for calling out, just added links in a few places where the context was appropriate

main/guides/ui-tutorial/starting/index.md
```diff
@@ -0,0 +1,68 @@
...
+yarn add -D ses @endo/eventual-send
+```
+
+Also, we'll need to add `buffer` to use [cosmos-kit](https://github.com/cosmology-tech/cosmos-kit) later on, so let's do that now:
+
+```
+yarn add -D buffer
+```
+
+Now, create a new file `src/installSesLockdown.ts`:
```
Jovonni, line 35: we could probably link to more documentation on what this file is doing. Even if we just link to docs on lockdown. As now its just "make a new file, and copy/paste this". There an opportunity to explain lockdown briefly here. Although it isnt a blocker for the dev. More informational

main/guides/ui-tutorial/conclusion/index.md
```diff
@@ -0,0 +1,20 @@
+## Conclusion
+
+Throughout this tutorial, you've accomplished all the basics of building an Agoric Dapp UI:
+
+- Scaffolding a new React App.
+- Setting up `AgoricProvider` and connecting to a wallet.
+- Reading and rendering purse balances.
+- Reading contract data with `chainStorageWatcher`.
```
Jovonni, line 8: My thought here, and this also was a gotcha for me, but this tutorial is showing how you read data from vstorage in general, as the code is using

```
[Kind.Data, 'published.agoricNames.instance']
...
[Kind.Data, 'published.agoricNames.brand']
```

But these aren't storage paths set by the contract. I think the wording can improve here so a brand new dev doesn't think they are reading contract data.

I see an opportunity to even combine/link other resources on how the writing of vstorage works, so they can understand the relationship between how writes work and these storage paths they see in this tutorial.

For this specifically, we could even change this to be 

```
- Reading vstorage data with `chainStorageWatcher`
```

would love thoughts on this

Jovonni: i also do understand the brand is from the contract, since its minting. Therefore the contract is still reading brand into from the asset being minted by the contract, but in a round-about way, this seems like we can say this more clear. Maybe its just we don't call it contract data here, but more vstorage data... idk 

So maybe, its just some clarity in the tutorial on how those storage paths can become where you store data if you want it to be assessible to the outside world.

Jovonni (approved):
LGTM, and i was able to use it to get through building the UI. Aside from my minor comments on clarity of vstorage writing, and reading here, this looks good team 🚀

main/guides/ui-tutorial/starting/index.md
samsiegart: Tbh I'm just having a hard time succinctly explaining it. Thinking out loud, it prevents the modification of certain intrinsics in the JS environment, and adds support for `harden` and `assert` etc. But that might lead to more questions than answers, as those things aren't really concerns for most UI devs (unless you're using https://github.com/LavaMoat/LavaMoat). It's really just needed so the agoric libraries work, because they're also used in smart contracts where those things actually matter.

We've talking about having a way to use the agoric libraries without hardening, but there hasn't been any movement on that yet. Maybe just linking to endo here is enough for now.

main/guides/ui-tutorial/conclusion/index.md
samsiegart: Yea that's understandable, I wasn't sure if "vstorage" would sound too unfamiliar but I don't think it's a big concern, just changed it from "contract data" to "vstorage" everywhere.

Also added a link to https://docs.agoric.com/guides/zoe/pub-to-storage.html#publishing-to-chainstorage as well

samsiegart:
@Jovonni I just made some changes as you suggested and added the tutorial pages to the nav bar, mind taking one more look before I merge?

Also, I realized the `AgoricProvider` API has changed since https://github.com/Agoric/ui-kit/pull/109. I'll go back and update it here and in the example repo since it's kinda a pain to update all the branches in the example, and this tutorial specifies an older version of  `@agoric/react-components` so it should still work.

PR #1073, opened 2024-05-01
uploading required screenshots for UI component docs

otoole-brendan:
Screenshots that are needed for UI component documentation


kbennett2000 (approved):
Looks good to me!

dckc:
do these belong in the `/getting-started/` directory?

dckc:
Also, any particular reason to land this PR separately from #1074? This PR won't result in any visible changes to the docs.

otoole-brendan:
@dckc #1074 points to these .png files that need to be in the assets file before displaying correctly. I wasn't aware I could upload the files as part of #1074 - that would have been the better thing to do. But per your point (and my other comment on #1074) - it sounds like these need to be in their own assets file and not 'Getting Started' if I want them on the same hierarchy/level as 'Getting Started', not within

kbennett2000:
@otoole-brendan - Maybe a folder under `main/guides` for ui related docs? Then it would live at the same level as getting started...

PR #1074, opened 2024-05-01
Create uicomponentlibrary.md

otoole-brendan:
Draft documentation for UI component library

### DIFF
--- a/main/guides/getting-started/uicomponentlibrary.md
+++ b/main/guides/getting-started/uicomponentlibrary.md
@@ -1,0 +1,112 @@
+# Welcome to the Agoric UI Component Library
+
+You'll find a collection of ready-to-use or pre-coded user interface elements that help accelerate your Agoric Dapp-building. These pre-designed elements include wallet connect buttons, chain selector, node selector, amount input and wallet provisioning.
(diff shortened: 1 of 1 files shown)


main/guides/getting-started/uicomponentlibrary.md
```diff
@@ -0,0 +1,112 @@
+# Welcome to the Agoric UI Component Library
```
dckc, line 1: any particular reason to put this under `/getting-started/`?

dckc:
I suppose this is w.r.t. #922

main/guides/getting-started/uicomponentlibrary.md
otoole-brendan: @dckc good catch - I want it to be a separate page on the same hierarchy as 'Getting Started' - not within it. @kbennett2000 might need your help to do that

kbennett2000: Sure, we can create a subfolder underneath `main/guides` for all of this

otoole-brendan:
Redundant due to https://github.com/Agoric/documentation/blob/main/main/guides/UIComponentLibrary/UIComponentsstart.md

PR #1096, opened 2024-05-16
feat: WIP added a figure that gives an overview of getting started setup

mudassir-agoric:
This is a WIP and I need feedback. 

I am thinking of adding an overview figure in the getting start docs. Have a look and let me know what you think.


main/guides/getting-started/index.md
```diff
@@ -28,6 +28,10 @@ Before getting started, there are some resources you might want to keep handy in
 - Send us a message on [X](https://twitter.com/agoric)
 - Send an Email to [Developer Relations](mailto://devrel@agoric.com)
 
+## Offer Up Setup - Overview
+
+![An Overview](./assets/getting_started_overview.png)
```
dckc, line 33: Interesting diagram.

How did you edit it?
I prefer that we check in the source of diagrams.

mudassir-agoric: I created it in miro. I don't see how to download source from there. But I can download PDF which is vector based. Does that work? Or Should I create it somewhere else or find a way to generate SVG or something?

dckc: Miro might be OK, as long as the rest of us have permission to edit it.

I was a little puzzled by the significance of the colors and such; I wondered about an ordinary numbered list.

Perhaps something in between?
 - [Building a Horizontal Timeline With CSS and JavaScript](https://webdesign.tutsplus.com/building-a-horizontal-timeline-with-css-and-javascript--cms-28378t)

mudassir-agoric: Ok. I will look into and get back.

Jovonni: I like the spirit of this. Its almost like having a table of contents for the getting started tutorial. We may want to consider a common design patter for all of our dev diagrams. I also agree with @dckc a numbered list good suffice for this as well. Good stuff @mudassir-agoric  🚀

mudassir-agoric: Thank you for valuable feedback - will incorporate it. Cheers.

dckc: > ... Its almost like having a table of contents for the getting started tutorial.

Well, yes... a big question is why the existing table of contents doesn't suffice:


image


Should there be fewer top-level sections? Should the headings change?

dckc: - "Installing Prerequisites" could go under "Platform Requirements" (or somehow be combined with it)
 - "Installing Dapp Dependencies" could go under "Creating Your Dapp From a Template"
 - "Starting a Local Agoric Blockchain" could be shortened to "Blockchain Setup"
 - "Starting the Dapp Smart Contract" and "Starting the Dapp" could be combined into 1 top-level heading
 - "Setting up a Keplr Wallet Demo Account" could be shortened to "Keplr Wallet Demo Account" or "Demo Wallet Setup"

dckc: > ... We may want to consider a common design patter for all of our dev diagrams.

I think @ivanlei suggested that a while back. I guess it should have its own issue:

 - #1099

Issue #947, opened 2024-02-01
Vstorage journey - 4daydapp


sufyaankhan:
**User Story:** As a Developer I should be able to read the docs and understand what vstorage is, how I can leverage vstorage, how it should be used in the context of the dapp-agorics-basics

E.g. vstorage is a journey of requesting powers within a core-eval, creating a storage node, passing the storage node in privateArgs to a contract, using the storage node within a contract, reading from vstorage via raw HTTPS or ui-kit.

dckc:
[Querying VStorage](https://docs.agoric.com/guides/getting-started/contract-rpc.html#querying-vstorage) has a relevant diagram. It only shows querying with `agd` and `agoric follow`, not HTTP nor ui-kit.

dckc:
In order for a UI to show the remaining inventory, the contract should probably publish the allocation of `inventorySeat` to vstorage. @LuqiPan here's hoping I can have docs sufficient for you to do that after we land PR 8.

https://github.com/Agoric/dapp-agoric-basics/pull/8#discussion_r1486630828

dckc:
This seems to include #758 .

dckc:
re vstorage reference docs, I'm leaving this somewhat obscure stuff to a discussion...

 - https://github.com/Agoric/agoric-sdk/discussions/9115

dckc:
more on reference docs...

 - [test-vstorage-ref.js.md](https://github.com/Agoric/dapp-agoric-basics/blob/dc-vstorage-ref/contract/test/snapshots/test-vstorage-ref.js.md) WIP

PR #1110, opened 2024-05-31
[Draft] docs: add VStorage documentation

ansabgillani:
Draft: Adding [VStorage documentation](https://github.com/orgs/Agoric/projects/19/views/103)
File is temp. Will change name + add reference to url once approved.

### DIFF
--- a/main/reference/vstorage-temp.md
+++ b/main/reference/vstorage-temp.md
@@ -1,0 +1,134 @@
+# VStorage in Agoric
+
+In the Agoric platform, VStorage, or a Virtual Storage, is a key-value store that:
+
+- provides a **read-only interface for clients of the consensus layer**. Clients can query the stored data using specific paths.
+
+- is organized in a **hierarchical, path-based structure**. Each fragment of data is stored at a specific path, and clients can query these paths to retrieve the data.
+
+- can be written through a specialized API called **[chainStorage]**(https://docs.agoric.com/guides/zoe/pub-to-storage.html#publishing-to-chainstorage). This API allows the VM to manage data storage in a controlled and secure manner.
+
+<!-- Extracted from the data examples -->
+<!-- https://github.com/Agoric/agoric-sdk/tree/003f0c2232815a8d64a3f9a5b05521a10160ce34/golang/cosmos/x/vstorage#readme -->
+<!-- https://github.com/Agoric/agoric-sdk/tree/agoric-upgrade-13/packages/inter-protocol#reading-data-off-chain -->
+
+<details>
+<summary> Need for store in dApp? </summary>
+In a dApp, VStorage can be used for:
+
+- to store crucial state and configuration data, making it accessible for clients inside the consensus layer.
+- to read the current state of smart contracts.
+
+</details>
+
+![vstorage query diagram](./assets/vstorage-brand-q.svg)
+
+## VStorage Hierarchy
+
+VStorage is structured as a tree with paths and nodes that store the actual data. This design facilitates easy querying while ensuring data consistency and security.
+
+For example, here is an example of [Inter Protocol key structure](https://github.com/Agoric/agoric-sdk/tree/agoric-upgrade-13/packages/inter-protocol#reading-data-off-chain) inside Agoric:
+
+```
+- published
+    - vaultFactory
+        - governance
+        - metrics
+        - managers
+            - manager0
+                - metrics
+                - governance
+                - vaults
+                    - vault0
+```
+
+You can then use this hierarchy to access the data fragments.
+```sh
+# lists vaults
+$ agd query vstorage keys 'published.vaultFactory.managers.manager0.vaults'
+```
+
+<!-- Should be pushed to the end of document to ensure that this block is not too much data -->
+<details>
+<summary>Here are some of the most commonly used keys in dapp development</summary>
+
+- **Top level keys**: The `published` and `bundles` keys are the most relevant to dapp development.
+
+```js
+{
+    activityhash: 'historical',
+    beansOwing: 'swingset execution fee accounting',
+    bundles: 'MsgInstallBundle outcome',
+    egress: 'reserved for future use',
+    highPrioritySenders: 'a priority mechanism',
+    mailbox: 'reserved for future use',
+    published: 'for the chainStorage API; see below',
+}
+```
+
+- **published.\* keys**: The following keys appear under `published`.
+
+```js
+{
+    agoricNames: 'name service controlled by chain governance',
+    auction: 'see Inter Protocol',
+    boardAux: 'auxiliary data for brands etc. keyed by boardId (since #49 2023-09-21)',
+    committees: 'see Inter Protocol',
+    crabble: 'reserved by chain governance proposal #64 decided 2023-12-18',
+    kread: 'reserved by chain governance proposal #53 decided 2023-10-01',
+    priceFeed: 'see Inter Protocol',
+    provisionPool: 'provideds initial IST during smart wallet provisioning',
+    psm: 'see Inter Protocol',
+    reserve: 'see Inter Protocol',
+    vaultFactory: 'see Inter Protocol',
+    wallet: 'smart wallet status',
+}
+```
+
+- **agoricNames hubs**: agoricNames contains several other NameHubs.
+See also [agoricNames](https://docs.agoric.com/guides/integration/name-services.html#agoricnames-agoricnamesadmin-well-known-names).
+
+```js
+[
+    'brand',
+    'installation',
+    'instance', 
+    'issuer', 
+    'oracleBrand', 
+    'vbankAsset'
+]
+```
+</details>
+
+## VStorage with CLI
+VStorage can be accessed via CLI using:
+
+```sh
+$ agd /[--node {url}] query vstorage {path}
+```
+
+For example:
+```sh
+$ agd query vstorage path published.agoricNames
+children:
+- brand
+- installation
+- instance
+...
+```
+
+With Agoric CLI, you can also use `follow` command to support vstorage query plus some of the marshalling conventions discussed below:
+```sh
+$ agoric follow -lF :published.agoricNames.brand
+[
+  [
+    "BLD",
+    slotToVal("board0566","Alleged: BLD brand"),
+  ],
+  [
+    "IST",
+    slotToVal("board0257","Alleged: IST brand"),
+  ],
+...
+]
+```


main/reference/vstorage-temp.md
```diff
@@ -0,0 +1,144 @@
+# VStorage in Agoric
+
+In the Agoric platform, VStorage, or a Virtual Storage, is a key-value store that:
```
toliaqat, line 3: We don't refer to it as "Virtual Storage" so no need to mention it.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
-# VStorage Reference
+# VStorage in Agoric
 
-See also:
+In the Agoric platform, VStorage, or a Virtual Storage, is a key-value store that:
 
-- [Querying VStorage](/guides/getting-started/contract-rpc#querying-vstorage)
-- [Publishing to chainStorage](../guides/zoe/pub-to-storage)
-- [x/vstorage module](https://github.com/Agoric/agoric-sdk/tree/003f0c2232815a8d64a3f9a5b05521a10160ce34/golang/cosmos/x/vstorage#readme)
+- provides a **read-only interface for clients of the consensus layer**. Clients can query the stored data using specific paths.
```
mudassir-agoric, line 5: change to something like: "...interface where clients can query..."

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
...
+- is organized in a **hierarchical, path-based structure**. Each fragment of data is stored at a specific path, and clients can query these paths to retrieve the data.
```
mudassir-agoric, line 6: remove this: Each fragment of data is stored at a specific path, and

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
...
+- can be written through a specialized API called **[chainStorage](https://docs.agoric.com/guides/zoe/pub-to-storage.html#publishing-to-chainstorage)**. This API allows the VM to manage data storage in a controlled and secure manner.
```
mudassir-agoric, line 7: remove: This API allows the VM to manage data storage

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
...
+                    - vault0
+```
+
+You can then use this hierarchy to access the data fragments.
+```sh
+# lists vaults
+$ agd query vstorage keys 'published.vaultFactory.managers.manager0.vaults'
+children:
+- vault0
+```
```
mudassir-agoric, line 36: This is good.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
...
+    slotToVal("board0566","Alleged: BLD brand"),
+  ],
+  [
+    "IST",
+    slotToVal("board0257","Alleged: IST brand"),
+  ],
+...
+]
+```
+
```
mudassir-agoric, line 71: This does not whole lot of sense - let us get rid of this.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
-# VStorage Reference
+# VStorage in Agoric
```
mudassir-agoric, line 1: remove " in Agoric"

main/reference/vstorage-ref.md
dckc: It assumes knowledge of the `agd` CLI. Perhaps add a cross-reference to https://docs.agoric.com/guides/agoric-cli/agd-query-tx.html#query-commands ?

Also, add a section about `agd query vstorage` in https://docs.agoric.com/guides/agoric-cli/agd-query-tx.html ?

dckc: Do we expect our readers to be more familiar with `agd` or with JS APIs?

https://docs.agoric.com/guides/getting-started/ui-tutorial/querying-vstorage.html has an example using `chainStorageWatcher.watchLatest` to access `published.agoricNames.instance`. Did you consider using that?

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
...
+```
+
+For example:
+```sh
+$ agd query vstorage path published.agoricNames
+children:
+- brand
+- installation
+- instance
+...
```
dckc, line 52: Ah. yes. this is the sort of material that I'd like to see in https://docs.agoric.com/guides/agoric-cli/agd-query-tx.html

The `--node` flag is discussed more thoroughly there. I don't see any reference to it here, so you might as well drop it here.

What's the `/` in `/[--node {url}]` by the way?

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,83 @@
...
+]
+```
+
+## VStorage Query API
+
+## VStorage chainStorage API
 
-## vstorage: top level keys
+## VStorage Viewer
```
dckc, line 76: These headings are taken from an semi-automated test. see the commit message for https://github.com/Agoric/documentation/commit/5607e3bf139bbe90aefd93b227e48a53a5a096e7 for details

I didn't manage to put the test into ci in such a way that changing the heading here without changing the test would be flagged as an inconsistency. I wonder what it would take to do that.

main/reference/vstorage-ref.md
```diff
@@ -44,185 +115,193 @@ see also [Inter Protocol data](https://github.com/Agoric/agoric-sdk/tree/agoric-
...
-      priceAggregator: Object @Alleged: BundleIDInstallation#board02021 {},
-      provisionPool: Object @Alleged: BundleIDInstallation#board05311 {},
-      psm: Object @Alleged: BundleIDInstallation#board05432 {},
-      reserve: Object @Alleged: BundleIDInstallation#board00218 {},
-      scaledPriceAuthority: Object @Alleged: BundleIDInstallation#board04719 {},
-      walletFactory: Object @Alleged: BundleIDInstallation#board04312 {},
-    }
+{
+  VaultFactory: Object @Alleged: BundleIDInstallation#board05815 {},
+  auctioneer: Object @Alleged: BundleIDInstallation#board04016 {},
```
dckc, line 144: likewise, these data were produced by a semi-automated test.

Is it important to change the indentation?

main/reference/vstorage-ref.md
```diff
@@ -368,6 +420,39 @@ The `published.wallet.${address}` key has wallet's last update. For example:
...
+      },
+    ],
+  },
+  updated: 'balance',
+}
 ```
+
+See also:
+
+- [Querying VStorage](/guides/getting-started/contract-rpc#querying-vstorage)
```
dckc, line 456: https://docs.agoric.com/guides/getting-started/ui-tutorial/querying-vstorage.html should probably go before `/getting-started/contract-rpc` in this see also list

main/reference/vstorage-ref.md
ansabgillani: `/` was a typo. Fixing it in a commit. 

```
The query goes to a local node at tcp://localhost:26657 by default. To use another node:
$ agd status --node https://devnet.rpc.agoric.net:443
```

Can you confirm this is same for `query` as well? I am not sure about it. There is no mention about `--node` parameter there. Maybe I can add information there and then refer it to that.

main/reference/vstorage-ref.md
ansabgillani: 1. I will revert these changes. I understand. 
2. Maybe run a script that opens the MD file and imports the JSON object with headings from the test and compares the two. If the headings are dependent on the result of a test, we can add it as a subsequent step in the CI.

main/reference/vstorage-ref.md
ansabgillani: I did it via linter. Reverting these changes.

main/reference/vstorage-ref.md
ansabgillani: I have removed it for now. This is important. Already referred here and we can get extra info on this from 
https://fa4fcd61.documentation-7tp.pages.dev/guides/agoric-cli/#agoric-help

main/reference/vstorage-temp.md
dckc: https://github.com/Agoric/agoric-sdk/tree/003f0c2232815a8d64a3f9a5b05521a10160ce34/golang/cosmos/x/vstorage#readme starts with "Virtual Storage"

main/reference/vstorage-ref.md
ansabgillani: I understand. I have added both, JS APIs and `agd` here.

main/reference/vstorage-ref.md
dckc: yes, `--node` applies to all `agd query` and `agd tx` sub-commands, IME.

The help text is oddly inconsistent, though: `agd query --help | grep node` is empty, as is `agd query vstorage --help | grep node` but:

```console
$ agd query tx --help | grep node
      --node string        <host>:<port> to Tendermint RPC interface for this chain (default "tcp://localhost:26657")
```

@gibson042 Do you know of some reason why the help text is inconsistent? If not, I suppose we owe an issue?

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
-# VStorage Reference
+# VStorage
```
dckc, line 1: As the "See also" list at the bottom of the page shows, we have more than one page on VStorage. The heading should distinguish this one somehow. If it's not reference docs for vstorage, what is it?

Do we move reference docs for vstorage elsewhere? Or do we just not have it any more?

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
...
+    "IST",
+    slotToVal("board0257","Alleged: IST brand"),
+  ],
+...
+]
+```
 
-## vstorage: top level keys
+## VStorage Query API
+VStorage querier service can directly be incorporated in dApps directly using [agoric.vstorage.Query](https://github.com/Agoric/agoric-sdk/blob/mainnet1B/golang/cosmos/proto/agoric/vstorage/query.proto#L11) package.
```
dckc, line 79: directly incorporated how?

I don't think we have docs showing how to make / use gRPC clients.

Citing the protobuf definition is useful, but I suggest not going beyond that:

```suggestion
## VStorage Protobuf Service Definition
The protobuf definition is of vstorage is [agoric.vstorage.Query](https://github.com/Agoric/agoric-sdk/blob/mainnet1B/golang/cosmos/proto/agoric/vstorage/query.proto#L11).
```

p.s. "can directly be incorporated" is [passive voice](https://en.wikipedia.org/wiki/Passive_voice). Avoid passive voice unless it's important to emphasize the receiver of the action.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
...
+- provides a **read-only interface where clients of the consensus layer** can query the stored data using specific paths.
+- is organized in a **hierarchical, path-based structure**. Clients can query these paths to retrieve the data. For example:
+```sh
+$ agd query vstorage keys 'published.vaultFactory.managers.manager0.vaults'
+children:
+- vault0
+```
+- can be written through a specialized API called **[chainStorage](https://docs.agoric.com/guides/zoe/pub-to-storage.html#publishing-to-chainstorage)** from within the JavaScript VM.
+
+![vstorage query diagram](../guides/getting-started/assets/vstorage-brand-q.svg)
```
dckc, line 14: This page now has a lot of overlap with `/guides/getting-started/contract-rpc.html`, so there should be links in both directions to clarify how they relate. Perhaps reduce the vstorage material in `contract-rpc`?

main/reference/vstorage-ref.md
ansabgillani: Added here #1122

main/reference/vstorage-ref.md
ansabgillani: The idea is to create a centralized reference for VStorage. I feel centralized reference is okay like this without reducing context at some place else.

main/reference/vstorage-ref.md
ansabgillani: fixed it.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
...
-- [Publishing to chainStorage](../guides/zoe/pub-to-storage)
-- [x/vstorage module](https://github.com/Agoric/agoric-sdk/tree/003f0c2232815a8d64a3f9a5b05521a10160ce34/golang/cosmos/x/vstorage#readme)
+- provides a **read-only interface where clients of the consensus layer** can query the stored data using specific paths.
+- is organized in a **hierarchical, path-based structure**. Clients can query these paths to retrieve the data. For example:
+```sh
+$ agd query vstorage keys 'published.vaultFactory.managers.manager0.vaults'
+children:
+- vault0
+```
+- can be written through a specialized API called **[chainStorage](https://docs.agoric.com/guides/zoe/pub-to-storage.html#publishing-to-chainstorage)** from within the JavaScript VM.
```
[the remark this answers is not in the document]
ansabgillani: fixed it.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
...
+
+## VStorage Viewer
+To visualize the current state of VStorage, The [vstorage-viewer](https://github.com/p2p-org/p2p-agoric-vstorage-viewer) contributed by p2p is often _very_ handy:
+
+[![vstorage viewer screenshot](https://user-images.githubusercontent.com/150986/259798595-40cd22f0-fa01-43a9-b92a-4f0f4813a4f6.png)](https://p2p-org.github.io/p2p-agoric-vstorage-viewer/#https://devnet.rpc.agoric.net/|published,published.agoricNames|)
+
+## DApp UI and VStorage
+As mentioned above, VStorage offers clients an access to data from contracts, including dApp UI. For detail tutorial on tools and libraries available for this connection is available [here](https://docs.agoric.com/guides/getting-started/ui-tutorial/querying-vstorage.html#querying-vstorage).
+
+## Common VStorage Examples
```
dckc, line 123: The docs on this page were intended as not just common examples but an exhaustive reference of what's in vstorage, as part of #947.

If this PR is changing that, we would need to renegotiate with the folks involved in #947.

main/reference/vstorage-ref.md
```diff
@@ -21,8 +136,7 @@ The `published` and `bundles` keys are the most relevant to dapp development.
       published: 'for the chainStorage API; see below',
     }
 ```
-
-## vstorage: published.\* keys
+### vstorage: published.\* keys
```
dckc, line 139: Demoting these from h2 to h3 means they don't show up in the page TOC. Is that on purpose?

main/reference/vstorage-ref.md
```diff
@@ -85,6 +198,9 @@ regarding un-marshalling the data using board IDs.
 `published.agoricNames.instance` contains _instances_ of important contracts.
 The data at this key are the entries of the NameHub. Here we show the object comprised of those entries.
 
+`published.agoricNames.instance` contains _instances_ of important contracts.
+The data at this key are the entries of the NameHub. Here we show the object comprised of those entries.
+
```
dckc, line 203: I don't think you meant to repeat this.

main/reference/vstorage-ref.md
gibson042: Yes, an issue would be good. I'm pretty sure this is because the [Cobra](https://cobra.dev/) library responsible for CLI option definition shows help for applicable options defined higher in the command hierarchy (so-called ["persistent flags"](https://cobra.dev/#persistent-flags)) only at an invocable command (which `agd query tx` is but e.g. `agd query vstorage` is not because it must be followed by "children" or "data" or "path" and `agd query params` is not because it must be followed by "subspace").

But there should be a way to do it, because "--chain-id" seems to work that way (e.g., it shows in both e.g. `agd tx --help` and [as a Global Flag] in `agd tx gov --help`).

main/reference/vstorage-ref.md
dckc: The JS API that has been added is how to _write_ to vstorage / chainstorage from a contract. I was suggesting adding the API for _querying_ vstorage from a client dapp, as suggested in #947. I suppose both are relevant. I'm not sure how big the scope should grow here, though.

main/reference/vstorage-ref.md
dckc: Expanding the scope of this page from [reference](https://docs.divio.com/documentation-system/reference/) to also cover [explanation](https://docs.divio.com/documentation-system/explanation/)/[how-to](https://docs.divio.com/documentation-system/how-to-guides/) seems awkward, especially since it overlaps with several other discussions of vstorage.

dckc (commented):
re technical correctness: this change is mostly neutral. I don't think it introduces any new material. The "Common VStorage Examples" heading suggests that it's no longer exhaustive, but the content is just as exhaustive as it was, and the top-level heading still says "Reference".

I don't see any errors.

So the impact is mainly organizational / editorial. I defer to others on that.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
 # VStorage Reference
 
-See also:
+In the Agoric platform, VStorage is a key-value store that:
+
+- provides a **read-only interface where clients of the consensus layer** can query the stored data using specific paths.
```
Jovonni, line 5: maybe want to clarify here, "read-only" to external, but contracts can write to it. something like this, so they don't think vstorage is read-only in general.

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
...
+    slotToVal("board0257","Alleged: IST brand"),
+  ],
+...
+]
+```
+
+## VStorage Protobuf Service Definition
+The protobuf definition of vstorage is [agoric.vstorage.Query](https://github.com/Agoric/agoric-sdk/blob/mainnet1B/golang/cosmos/proto/agoric/vstorage/query.proto#L11) package.
+```
+service Query {
```
Jovonni, line 81: good idea to put the proto here

main/reference/vstorage-ref.md
```diff
@@ -1,13 +1,128 @@
...
+...
+const installation = await E(zoe).startInstance(
+  ...,
+  privateArgs: harden({
+      storageNode,
+      marshaller,
+  })
+)
+```
+The `chainStorage` node corresponds to the `published` key in the [vstorage hierarchy](https://docs.agoric.com/reference/vstorage-ref.html#vstorage-hierarchy). Using `E(chainStorage).makeChildNode(contractName)` gives the contract access to write to the `published.{contractName}` key and all keys under it.
```
Jovonni, line 112: maybe put a security warning here to ensure contracts aren't passed the top most storage node, because then they can write to any key under it. I made this mistake in the past. 

could be something like 
```
Note: Always ensure to only pass the storage node to your contract that you explicitly intend for it to use. This will avoid giving a contract too much write powers to write to unintended storage nodes.
```

What do you think @ansabgillani ?

PR #1112, opened 2024-05-31
docs(vstorage): start headings at level 1

dckc:
this page is missing a title because there is no h1


samsiegart:
We should probably make sure all the ui-tutorial pages do this as well

samsiegart (approved):
Just gonna approve this but still think we should fix the other pages as well

PR #1113, opened 2024-05-31
docs: include UI tutorial in nav menu

dckc:
I couldn't find the UI Tutorial the 1st time I looked for it.


samsiegart (approved):
Didn't think of this at first, looks good though

image

PR #1116, opened 2024-06-03
docs: Add follow command to Agoric CLI Reference

ansabgillani:
Add `follow` command reference to doc

### DIFF
--- a/main/guides/agoric-cli/index.md
+++ b/main/guides/agoric-cli/index.md
@@ -46,7 +46,7 @@ See also:
 
 ## agoric follow
 
-Follow an Agoric Casting leader tosupport vstorage query along with some of the marshalling conventions.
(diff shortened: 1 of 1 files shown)


toliaqat (changes requested):
This change has output of `--help`, which might not be ideal content for documentation. The reason is that these options can get change over time. In the documentation, we usually add stuff that is long-lasting and might not have explained in the --help option.

ansabgillani:
https://github.com/Agoric/documentation/pull/1116#pullrequestreview-2097618721
Just followed the conventions here from the rest of the page. Based on our conversation, all good.

mudassir-agoric (commented):
Looks good to me!

Jovonni (approved):
lgtm

toliaqat (approved):

PR #1121, opened 2024-06-06
Update index.md with brief on yarn installation.

Jovonni:
Adding brief clarity on the following commands:

```
corepack enable
yarn --version # for verification
```

How's this @ivanlei ?

### DIFF
--- a/main/guides/getting-started/index.md
+++ b/main/guides/getting-started/index.md
@@ -94,6 +94,7 @@ node --version
 </details>
 
 ### Installing Yarn
+This will install yarn, and verify it is installed correctly. Install yarn by either running ```npm install --global yarn```, or visit https://classic.yarnpkg.com/en/docs/install for more information. See below, ```corepack enable``` will add the Corepack shims to the install directories, and ```yarn --version``` will verify yarn is installed correctly.
 
 ```sh
 corepack enable


ivanlei (approved):
ty sir! 🫡

main/guides/getting-started/index.md
```diff
@@ -94,6 +94,7 @@ node --version
 </details>
 
 ### Installing Yarn
+This will install yarn, and verify it is installed correctly. Install yarn by either running ```npm install --global yarn```, or visit https://classic.yarnpkg.com/en/docs/install for more information. See below, ```corepack enable``` will add the Corepack shims to the install directories, and ```yarn --version``` will verify yarn is installed correctly.
```
LuqiPan, line 97: _question_

Would the user still need `corepack enable` if they already have `yarn` installed globally?

Also, note to ourselves, the direction will need to be changed once we get on modern yarn

turadg: They should always use corepack. We should never advise to install yarn globally because that will prevent when we want to get past Yarn v1. (As @LuqiPan points out)

`yarn --version` just verifies that _a_ Yarn version is installed correctly. But I think it's redundant with `corepack enable` because if that command succeeds then yarn will be available.

amessbee:
@Jovonni this one too...

main/guides/getting-started/index.md
turadg: @Jovonni did you see the comment above before merging?

PR #1122, opened 2024-06-07
docs: Add agd query vstorage keys command reference

ansabgillani:
Add agd query vstorage keys reference mentioned in #1110

### DIFF
--- a/main/guides/agoric-cli/agd-query-tx.md
+++ b/main/guides/agoric-cli/agd-query-tx.md
@@ -118,7 +118,7 @@ $ agd query bank balances $addr -o json
 
 ## agd query gov proposals
 
-Query for a all paginated proposals that match optional filters:
+Query for a all paginated proposals that match optional filters.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


dckc (approved):

main/guides/agoric-cli/agd-query-tx.md
```diff
@@ -130,6 +130,17 @@ $ agd query gov proposals --output json | \
...
+Example:
+```
+$ agd query vstorage keys 'published.vaultFactory.managers.manager0.vaults'
+children:
+- vault0
+```
+
 ## Transaction Commands
 
 Making transactions requires setting up an **account** with a private key for signing. The [basic dapp local chain](../getting-started/#starting-a-local-agoric-blockchain) container has a number of keys set up for use with `--keyring-backend=test`. Use `agd keys list --keyring-backend=test` to see them.
```
Jovonni, line 146: makes me wonder, do we have any docs on helping users understand creating new keys, keyrings, adding accounts by recovery etc?

might be useful if we don't have anything. will look to see if we have anything here shortly.

Jovonni (approved):
lgtm 🚀

main/guides/agoric-cli/agd-query-tx.md
dckc: > ... do we have any docs on helping users understand creating new keys, keyrings, adding accounts by recovery etc?

Not much. Maybe you want to open an issue?

We have sort of gotten by assuming folks get that from stuff like the [cosmos delegator guide](https://hub.cosmos.network/delegators/delegator-guide-cli) and [Interacting with Gaiad (CLI) | Cosmos Hub](https://hub.cosmos.network/hub-tutorials/gaiad) ... but pages like that keep going 404.

We used to do more validator training; for example: https://github.com/Agoric/agoric-sdk/wiki/Validator-Guide-for-Devnet#if-you-dont-have-an-operator-key

PR #1125, opened 2024-06-14
working chatbot poc, with new Chatbot.vue

Jovonni:
Proof of concept of dedicated AgoricGPT docs chatbot. Screenshots below show the button
agoricGPT

Just added one chatbot component, the magic is happening in the iframe

Largely modeled this POC after:
https://docs.celestia.org/

We would need to host the actual chatbot somewhere, and can work on throttling the usage, etc


Jovonni:
We can also sign up with something like kapa.ai, or one of its competitors. This POC is just doing it in-house.

PR #1155, opened 2024-07-10
WIP: docs: added a document for prepareVowTools

frazarshad:


amessbee:
@frazarshad this looks good, but I doubt people will want to have it merged as is in docs due to relevance. Here are a couple of things we should do, before we close this PR.

1. Create a short version of this thing - probably half of it - and then add it as a glossary item.
2. Copy this whole thing and create a wiki page out of it. Maybe on agoric-sdk repo. So that all this information is not lost, and we can add it to docs when we are documenting orca API.

@toliaqat suggestions?

PR #1157, opened 2024-07-11
chore: update vitepress

0xpatrickdev:
## Description

Update vitepress to the latest version and fix a style regression by moving local css variable overrides to `styles.css`

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "prettier": "^1.19.1",
     "ses": "^0.18.8",
     "stylus": "^0.62.0",
-    "vitepress": "1.0.1"
+    "vitepress": "1.3.0"
   },
   "globals": {
     "harden": "readonly"
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "lint": "eslint 'snippets/**/*.js'",
     "build": "exit 0"
   },
+  "packageManager": "yarn@1.22.22",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -70,7 +71,7 @@
     "prettier": "^1.19.1",
     "ses": "^0.18.8",
     "stylus": "^0.62.0",
-    "vitepress": "1.0.1"
+    "vitepress": "^1.3.0"
```
LuqiPan, line 74: _question_

Worth pinning vitepress to 1.3.0 exactly?

0xpatrickdev: I don't have a preference, but happy to do this

LuqiPan: The reason I was asking is because I noticed that previously with `^1.0.1`, our cloudflare build process would run `yarn install` which would install vitepress 1.3.0 instead. So I think pinning would give us better control and visibility on which version of vitepress we're running.

0xpatrickdev: Agree with your reasoning, amended to pin

package.json
```diff
@@ -18,6 +18,7 @@
     "lint": "eslint 'snippets/**/*.js'",
     "build": "exit 0"
   },
+  "packageManager": "yarn@1.22.22",
```
LuqiPan, line 21: Thanks for specifying yarn version too!

LuqiPan (approved):
LGTM

PR #1158, opened 2024-07-15
doc(orch): use `try-catch`, not `.then` and `.catch`

michaelfig:
Mixing `await` style with `.then` and `.catch` is sufficiently confusing to novice JS programmers.  Please take advantage of `try-catch` instead.


erights (approved):
Good! LGTM.

I had this same thought when looking at the recent vow `when` retry logic. Would this same guidance still apply there? (If you like to turn that cleanup into a PR as well, I'd be happy to review.)

michaelfig:
> I had this same thought when looking at the recent vow `when` retry logic. Would this same guidance still apply there? 

Absolutely!  Here's the PR that I just now tagged you in: Agoric/agoric-sdk#9731

PR #1171, opened 2024-08-13
fix: bad link in core-eval proposal section

amessbee:
Contract Installation link in https://docs.agoric.com/guides/coreeval/proposal.html
points to: 
https://docs.agoric.com/guides/zoe/#contract-installation

There is no discussion about contract installation on that page. It seems that intended page with the 
relevant section is: https://docs.agoric.com/guides/zoe/contract-walkthru.html

I am fixing the link.


mujahidkay (commented):
Can we also fix it on core eval page `main/guides/coreeval/index.md` ?

amessbee:
@mujahidkay good catch. Fixed that too.

mujahidkay (approved):

PR #1192, opened 2024-09-06
feature example apps code on homepage

dckc:
Based on today's hackathon debrief, the worked examples in dapp-offer-up and dapp-agoric-basics were key learning materials.

 - [rendered](https://8b203704.documentation-7tp.pages.dev/)

screenshot:
image

### DIFF
--- a/main/index.md
+++ b/main/index.md
@@ -13,8 +13,14 @@ hero:
       text: Get Started
       link: /guides/getting-started/
     - theme: alt
-      text: View on GitHub
-      link: https://github.com/agoric/agoric-sdk
+      text: 'App #1: Offer Safety'
+      link: https://github.com/Agoric/dapp-offer-up#readme
+    - theme: alt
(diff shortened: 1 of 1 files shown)


0xpatrickdev:
`Platform:sdk` doesn't seem like a good use of real estate since we have `github` in the main navbar, nor the existing "View on Github". I think the same can be said for the discord and twitter tiles.

App 1-3 seem like great candidates for tiles with a description. With Twitter and Discord tiles removed, we can rework `Tutorials` and `Build` to capture the paths you'd like to highlight. `Build` seems like the CTA for App #3 Orchestration, maybe we should reword the title.

I also suggest we link `🎻 Orchestration Tutorial` for the second button, which would link it a total of two times.

dckc:
@0xpatrickdev thanks for the feedback. I don't understand your suggestions well enough to make them happen. Here's hoping for a moment to discuss.

Issue #1183, opened 2024-08-31
issues rendering "ratio math" documentation page


tgrecojs:
I have observed an issue with the loading of the "ratio math" page within the documentation. Navigating to the ratio math page using embedded hyperlinks (shown in the video) or the documentation search feature leads to a page with no content displayed. However, manually reloading the page results in the content rendering correctly. It is unclear if this issue exists in other sections of the Zoe API documentation.

https://github.com/user-attachments/assets/a133c7de-7a9f-4b78-a922-50a779dbda96

PR #1204, opened 2024-09-11
fix: Ratio Math page partially rendering

mujahidkay:
Closes: https://github.com/Agoric/documentation/issues/1183

### Description
Fixes Ratio Math page's rendering issues due to hydration. The root cause turned out to be how BigInts were being translated to html code when written in non-codeblock form. 

Also fixes some cases where there were no matching closes braces/parentheses.

### Changes
**Before**

https://github.com/user-attachments/assets/41a41ccb-819b-4383-9df9-bb2199a5aa3f
 
**After**
 
https://github.com/user-attachments/assets/64e05ed3-17fa-49fb-8f15-20ac8a18d48f

### DIFF
--- a/main/reference/zoe-api/ratio-math.md
+++ b/main/reference/zoe-api/ratio-math.md
@@ -13,7 +13,7 @@ Throws an error if the argument is not a valid **Ratio**.
 
 Throws messages for errors:
 - **Ratio ${ratio} must be a record with 2 fields.**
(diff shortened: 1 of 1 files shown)


dckc (approved):
It's good to get rid of these rendering bugs.

Meanwhile, manually maintaining these function signatures is something we'd like to move away from. See:
 - https://github.com/Agoric/agoric-sdk/issues/4291

Issue #945, opened 2024-02-01
normalize markdown formatting style


dckc:
vs-code keeps reformatting markdown files when I save them. I presume it's using prettier or the like. I could turn it off, I suppose, but perhaps it's better to set up tooling to do this uniformly?

kbennett2000:
I'm in favor of running all our code through something to make it look uniform, and I think prettier is what most folks are using? Anyway, the more consistent we can make things look to a new developer, the better IMO...

PR #1205, opened 2024-09-11
chore: normalize md formatting

mujahidkay:
Closes: https://github.com/Agoric/documentation/issues/945

### Description

 - Adds a prettier config for formatting md files
 - Adds a `lint:format` and `format` script for yarn
 - Adds CI for `lint:format`
 - Formats all markdown files


.husky/pre-commit
```diff
@@ -0,0 +1 @@
+npx lint-staged
```
dckc, line 1: `npx` involves runnnig code that's not verified against hashes in yarn.lock. Doing this in ci is a risk we do not usually take, I think.

package.json
```diff
@@ -11,7 +11,9 @@
     "test": "ava",
     "lint-fix": "yarn lint --fix",
     "lint": "eslint 'snippets/**/*.js'",
-    "build": "exit 0"
+    "format": "prettier --write '**/*.md' --config .prettierrc.json",
+    "build": "exit 0",
+    "prepare": "husky"
```
dckc, line 16: I think the shop has avoided pre-commit hooks, to date. Enforcement is done by preventing PRs from merging until they pass ci.

A `prettier --check` in the `lint` script is what I would expect. For precedent, see...

https://github.com/Agoric/agoric-sdk/blob/66e4cc0f36b5ed33f1b3fa235a335757ca868b9c/package.json#L60-L63

.prettierrc.json
```diff
@@ -0,0 +1,21 @@
+{
+    "printWidth": 80,
+    "tabWidth": 2,
+    "useTabs": false,
+    "semi": false,
```
dckc, line 5: Is this config used to format `.js` as well?

Our [house style](https://github.com/Agoric/agoric-sdk/wiki/Coding-Style) follows Airbnb, which _does_ include semicolons.

See also https://github.com/Agoric/agoric-sdk/tree/master/packages/eslint-config

.prettierrc.json
```diff
@@ -0,0 +1,21 @@
...
+    "singleQuote": true,
+    "proseWrap": "preserve",
+    "trailingComma": "none",
```
dckc, line 8: we do trailing commas too

dckc (commented):
This is good progress, but please use a lint check rather than precommit hook.

.prettierrc.json
mujahidkay: This is just for markdown files

.husky/pre-commit
mujahidkay: I'll replace it with a lint check. I wasn't aware of this - thanks for letting me know.

dckc: In this case, I suppose `lint-staged` _is_ in `yarn.lock`. But one typo and you're running `lint-stoged` which does who-knows-what.

mujahidkay:
> This is good progress, but please use a lint check rather than precommit hook.

I have addressed pre-commit hook related comments. The others don't apply since we are not messing with js files at all. PTAL.

package.json
```diff
@@ -11,6 +11,8 @@
     "test": "ava",
     "lint-fix": "yarn lint --fix",
     "lint": "eslint 'snippets/**/*.js'",
+    "format": "prettier --write '**/*.md' --config .prettierrc.json",
+    "lint:format": "prettier --check '**/*.md' --config .prettierrc.json",
```
dckc, line 15: To keep things from decaying, this has to be enforced in ci. So the `lint` script should call `lint:format`, I think. Or the ci workflow should be changed to call `lint:format`.

I suppose we can take care of that in a later PR, but here would be better.

dckc (approved):
I'd like ci enforcement to go into this PR. But doing it in a later PR is acceptable.

package.json
mujahidkay: Already added in this PR. https://github.com/Agoric/documentation/blob/a7bb1e108b5b5a9dc2e5a9465906a1ba5837ac08/.github/workflows/test-markdown.yml#L29

PR #1211, opened 2024-09-20
docs(upgrade): clarify installation vs. instance

dckc:


Chris-Hibbert (approved):
Interesting that it linked to the right method doc.

PR #1212, opened 2024-09-24
chore: lint inline js snippets

mujahidkay:
Recently, set up a prettier config to normalize markdown formatting. It enforced no trailing semi colons which resulted in some mismatches if a dev has semis enabled extension wide. So firstly, this PR fixes that by removing that option.

Secondly, not all inline JS snippets were being fixed as such. Neither `prettier`, not `eslint-markdown` plugins natively handled that support (there still exists cases of js snippets without trailing semis). Also explored `remark-lint` but to no success. Concocted a custom linter (s/o claude) to handle all cases that occur in the repo and it should cover almost all use-cases (don't expect it to be perfect, have tried out manual examples generic in nature that do get filtered). Also, normalizes all ''\`\`\`javascript' usages to '\`\`\`js' for consistency

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -2,7 +2,6 @@
     "printWidth": 80,
     "tabWidth": 2,
     "useTabs": false,
-    "semi": false,
     "singleQuote": true,
     "proseWrap": "preserve",
     "trailingComma": "none",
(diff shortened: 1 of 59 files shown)


toliaqat (approved):
Let's merge it, but I would recommend moving the linter code to some other repo later.

PR #1228, opened 2024-10-03
chore: downgrade glob

mujahidkay:
The issue arose because we introduced `glob` as a dependency in https://github.com/Agoric/documentation/pull/1212

The version introduced in this repo doesn't mesh well with the version(s) being used in some of the packages in agoric-sdk and hence, the linking step failed. Once I aligned the CI check breakage with the changes in docs repo, it was easier to pinpoint the root cause. 

As a fix, I have downgraded `glob` to the version that works for both cases. CI check now passes on agoric-sdk with this branch: https://github.com/Agoric/agoric-sdk/actions/runs/11157393879/job/31011590384?pr=10206

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "eslint-plugin-import": "^2.30.0",
     "eslint-plugin-jsdoc": "^35.5.1",
     "eslint-plugin-prettier": "^5.2.1",
-    "glob": "^10.4.5",
+    "glob": "7.1.7",
     "import-meta-resolve": "^2.2.2",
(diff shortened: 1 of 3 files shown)


Muneeb147 (commented):
Are the other changes required with downgrade?

mujahidkay:
> Are the other changes required with downgrade?
> Are the other changes required with downgrade? mjs -> cjs and adding a run-linter wrapper

no, good catch. updating

Muneeb147 (approved):

Issue #1225, opened 2024-10-02
Deadlink checker for Navbar/sidebar items


mujahidkay:
vitepress automatically detects deadlink in markdown files during the build process but it fails to catch deadlinks from paths specified in navbar and sidebar.

There is a need to check for deadlinks in config files (non-markdown) so issues like [these](https://agoricopco.slack.com/archives/C03UF6JMWE7/p1727883598580209) don't happen.

PR #1236, opened 2024-10-08
chore: add a deadlink checker for nav/sidebar items

mujahidkay:
Closes: https://github.com/Agoric/documentation/issues/1225

Added this check to build script to keep it consistent with vitepress's built-in deadlink checker for md files (raises error during build time)

Unrelated but also formats the scripts dir.

Testing:

You can test by checking out the branch and entering invalid links.

### DIFF
--- a/scripts/checkLinks.mjs
+++ b/scripts/checkLinks.mjs
@@ -5,7 +5,7 @@ import { fileURLToPath } from 'url';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-const extractLinks = (content) => {
+const extractLinks = content => {
   const noSingleLineComments = content.replace(/\/\/.*$/gm, '');
   const noComments = noSingleLineComments.replace(/\/\*[\s\S]*?\*\//g, '');
   const linkRegex = /link:\s*(['"])([^'"]+)\1/g;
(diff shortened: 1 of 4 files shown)


scripts/checkLinks.mjs
```diff
@@ -0,0 +1,64 @@
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const extractLinks = (content) => {
+  const linkRegex = /link:\s*'([^']+)'/g;
```
amessbee, line 9: I guess there is nothing to stop someone from using double quotes with links?

mujahidkay:
> @mujahidkay wondering whether we have any other options to fix #1225 ?

I think [Linkinator](https://www.npmjs.com/package/linkinator) seems like another option but I haven't really tested it. I will give it a go in my free time.

scripts/checkLinks.mjs
mujahidkay: good catch!

amessbee: I think this would also break the CI if there is a comment that contains a broken link?

amessbee (approved):
lgtm - nice work!

PR #1238, opened 2024-10-09
Adding Images to Orchestration Docs

amessbee:
Here are the changes:

1. Adding an orchestration contract architecture graphic.
2. Adding a graphic with potential connections between chains established through Agoric Orchestration.
3. Updating sequence diagrams to use mermaid code directly in markdown.
4. Removing unused assets.


main/guides/orchestration/index.md
```diff
@@ -10,6 +10,10 @@ The Agoric Orchestration API sits on top of Agoric’s novel VM that provides th
 
 Agoric’s Orchestration APIs simplify controlling accounts on remote chains, moving assets, and using capabilities on any chain the API reaches.
 
+<br/>
+<img src="./assets/orch_arch.svg" width="125%" />
+<br/>
```
dckc, line 15: Nice figure. I'd like to see it right after the 1st paragraph (ending "novel cross-chain-focused products.")

dckc: Also, how do we maintain it? Is this .svg file the editable source form? What tool did you use to create it?

amessbee: @dckc I created the original image using latex in ipe tool, and yes this was SVG. 

As per offline comments by @toliaqat I have updated the image with a new version, and moved it to [how-orchestration-works](https://ms-add-orch-images.documentation-7tp.pages.dev/guides/orchestration/how-orch-works.html) page. This image use the original source image from [Agoric Orchestration Docs document](https://docs.google.com/document/d/1xniiU38GcCKn66q-YurGf_DCA5w_2a8SE6dh-1O1M4c/edit) and was created use Canva online editor for which SVG is not available.

main/guides/orchestration/how-orch-works.md
```diff
@@ -15,6 +15,10 @@ To achieve these goals, several foundational technologies and protocols have
 been developed, with the [Inter-Blockchain Communication (IBC) protocol](https://ibcprotocol.org/) being
 one of the most prominent.
 
+<br/>
+<img src="./assets/orch_contract_arch.png" width="100%" />
```
dckc, line 19: any particular reason to switch from .svg to .png? .svg is more accessible - the text is searchable and selectable

dckc: oic

main/guides/orchestration/index.md
dckc: Canva? I found it in some Google slides. Here's an SVG export:

orch-arch

main/guides/orchestration/how-orch-works.md
amessbee: I think free version of Canva does not let you download SVG.

Back to SVG again for that image by converting the source slide you linked. There is another image on the landing page now (in PNG for now). I would probably tinkering a bit more, and perhaps add one more image too.

main/guides/orchestration/index.md
toliaqat: @dckc can you update the text in "User Story" box as per the image mudassir uploaded, and reupload the SVG?

main/guides/orchestration/how-orch-works.md
toliaqat: > the text is searchable and selectable

@dckc how is the text searchable and selectable in the SVG image? I cannot do that in your uploaded SVG. I could not find DEX in the source of your image as well.

amessbee: I thought I already fixed that?

https://8abcef3f.documentation-7tp.pages.dev/guides/orchestration/how-orch-works.html

main/guides/orchestration/index.md
toliaqat: I don't think svg vs png is a debate that should delay improving docs. We already have both kinds of images in the repo and they solve the problem at hand. Optimizing the format should come when we have learned enough after many iterations and have nothing else to improve.

toliaqat (commented):
@amessbee please remove images from the PR that are not needed anymore. For example: [main/guides/orchestration/assets/orch_arch.svg](https://github.com/Agoric/documentation/pull/1238/files#diff-89f50e0763730ee21e6f80b93f3327b69218e3bc4c00f99e7abdc847c00fc942)

amessbee:
@dckc @mujahidkay @toliaqat  I am adding a [vite plugin for mermaid](https://www.npmjs.com/package/vitepress-plugin-mermaid) that renders diagrams from within markdown and seems to work fine. This will save us from keeping both `.mmd` source file and the `.svg` graphic file and we can directly add the mermaid code to the markdown. I have updated the sequence diagram on [Orchestration landing page](https://ms-add-orch-images.documentation-7tp.pages.dev/guides/orchestration/) - never mind rest of the mess on that page. I will update docs wherever we are using mermaid code unless there is a reason not to.

main/guides/orchestration/index.md
```diff
@@ -10,6 +10,10 @@ The Agoric Orchestration API sits on top of Agoric’s novel VM that provides th
 
 Agoric’s Orchestration APIs simplify controlling accounts on remote chains, moving assets, and using capabilities on any chain the API reaches.
 
+<br/>
+<img src="./assets/chains4.png" width="100%" />
```
mujahidkay, line 14: nit: appropriate filename `chains.png` should be fine

package.json
```diff
@@ -67,10 +67,12 @@
     "eslint-plugin-prettier": "^5.2.1",
     "glob": "7.1.7",
     "import-meta-resolve": "^2.2.2",
+    "mermaid": "^11.3.0",
```
mujahidkay, line 70: we also want to probe agoric-sdk CI against this branch. This is to avoid any underlying package used by these new packages to break it (it's broken right now so we may have to wait for https://github.com/Agoric/documentation/pull/1240 to merge and then rebase)

toliaqat:
@amessbee I really like the mermaid code and rendering package—it’s great that it allows users to select and search text on the page. Thanks for adding that! However, I would have preferred it as a separate PR so we could review and merge the graphic and mermaid PRs independently.

mujahidkay (approved):
Good addition to the docs!

Kindly update branch and re-verify before merging.

main/guides/orchestration/index.md
dckc: > ... We already have both kinds of images in the repo and they solve the problem at hand.

Yes; we use .png for screenshots and, until now, we used .svg for diagrams.

> Optimizing the format should come when we have learned enough after many iterations and have nothing else to improve.

So only after many iterations we should consider accessibility? That's not a position I'm prepared to support.

https://www.ohchr.org/en/instruments-mechanisms/instruments/convention-rights-persons-disabilities#article-9

PR #1240, opened 2024-10-11
chore: Sync Endo versions

kriskowal:


kriskowal:
The Cloudflare errors are HTTP gateway problems with `yarn`. I assume that’s unrelated to the proposed changes, and we may just have to wait for some weather to pass.

0xpatrickdev (approved):
I see some duplicates in `yarn.lock`,  seemingly from references to `community-dev`, but assume fixing CI in `agoric-sdk` is more important

PR #1243, opened 2024-10-15
Mermaid Plugin: Render Inline Mermaid

amessbee:
Changes include:

adding mermaid and mermaid vite plugin 
update to sequence diagram on orchestration to use inline mermaid

### DIFF
--- a/main/.vitepress/config.mjs
+++ b/main/.vitepress/config.mjs
@@ -1,8 +1,8 @@
-import { defineConfig } from 'vitepress';
 import { nav } from './themeConfig/nav.js';
 import { rewrites } from './themeConfig/rewrites.js';
+import { withMermaid } from "vitepress-plugin-mermaid";
 
-export default defineConfig({
+export default withMermaid({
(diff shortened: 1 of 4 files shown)


toliaqat (commented):
The deleted image has a reference in other files. For example https://github.com/Agoric/documentation/blob/c1e76b7a6a6dd3bebf94598c83cd5afec15e2e3d/main/guides/orchestration/contract-walkthroughs/cross-chain-unbond.md#L3-L4

Please update the comment or not delete the image if the comment is important and we want to bring the image back in future.

amessbee:
> The deleted image has a reference in other files:
> 
> https://github.com/Agoric/documentation/blob/c1e76b7a6a6dd3bebf94598c83cd5afec15e2e3d/main/guides/orchestration/contract-walkthroughs/cross-chain-unbond.md#L3-L4
> 
> Please update the comment or not delete the image if the comment is important and we want to bring the image back in the future. If I were you, I would not add mermaid of the deleted image, and instead undelete the file to merge the original PR as soon as possible.

Do you think I should undelete other two unused SVG files too?

PR #1251, opened 2024-11-19
chore: Update Endo versions

kriskowal:
https://github.com/endojs/endo/releases/tag/ses%401.10.0


michaelfig (approved):
Looks like version updates to me!

PR #1260, opened 2024-12-27
Update key-concepts.md

Chris-Hibbert:
`OrchestrationAccountI` link leads to a blank page. Is `OrchestrationAccountCommon` the intended destination?


turadg (approved):
Yeah, @erights improved the naming recently. I suppose we need some integration builds of the docs with agoric-sdk. Or move this content into the monorep.

PR #1261, opened 2024-12-27
Update send-anywhere.md

Chris-Hibbert:
Various discrepancies have crept in between the docs and the code. This updates the doc to match the code in agoric-sdk.


main/guides/orchestration/contract-walkthroughs/send-anywhere.md
```diff
@@ -81,7 +81,7 @@ these functions with the necessary context (such as the contract state, logging,
 const orchFns = orchestrateAll(flows, {
   log,
   sharedLocalAccountP,
-  zoeTools,
+  zoeTools
```
Chris-Hibbert, line 84: Lint insistes on `no-trailing-comma`, which is incorrect for our JavaScript style. Is it worth finding and fixing this?

turadg: not worth fixing imo

turadg (approved):
Since this is a walk-through for a particular source file in agoric-sdk, I think we'd be better off having this doc there. It's bound to get out of sync again otherwise.

PR #1262, opened 2025-01-03
docs: getPayout() throws if there's no payment for the keyword

Chris-Hibbert:
Issue [8610](https://github.com/Agoric/agoric-sdk/issues/8610) pointed out that `E(zoeSeat).getPayout('foo');` returned undefined when the keyword was unknown, while the types claimed that it always returned a Payment. 

[10738](https://github.com/Agoric/agoric-sdk/issues/10738) changed Zoe to throw if the Payment didn't exist. This PR updates the documentation to mention that error.


turadg (approved):
Doesn't hurt but seems to me like a detail we should leave to docs generated from source code

